PR #294, opened 2013-06-04
Update to enable auto-elements for properties not inheriting from object...

csteeg:
I have another one for auto-elements

Eg, if a property was :

``` c#
public int MyInt {get; set}
```

`Expression<Func<object>>` will result in an expression with body set to UnaryExpression because of boxing the int to an object.
This caused

``` c#
new EntryAuto(bindingExpression: () => MyInt, caption: "MyInt")
```

to fail

### DIFF
--- a/CrossUI/CrossUI.Droid/Dialog/LinearDialogListView.cs
+++ b/CrossUI/CrossUI.Droid/Dialog/LinearDialogListView.cs
@@ -125,7 +125,7 @@ namespace CrossUI.Droid.Dialog
                 view.SetBackgroundDrawable(Resources.GetDrawable(Android.Resource.Drawable.ListSelectorBackground));
 
                 _list.AddView(view);
-                if (_divider != null)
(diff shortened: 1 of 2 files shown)


softlion:
@martijn00, why does mvvmcross still displays `OnViewNewIntent called - but this is not fully handled within MvvmCross currently. Check https://github.com/slodge/MvvmCross/pull/294 for more info` ?

nmilcoff:
@softlion I think it is because the ViewModel doesn't get notified of a new Intent that comes to an Activity. Suppose your app is already opened but in a background mode. Then you launch it again from another app... the app will come to foreground and the new intent will be received, but nothing else will happen

softlion:
@nmilcoff in fact you can reproduce this log message easily by putting your app to the background then tapping the icon to reopen it instead of using the task manager.

MKuckert:
Is this warning even necessary. Looks like a known limitation that could be documented and done. I don't see anything Mvx should support here.

PR #454, opened 2013-10-03
MvxRadioGroup

BenGladman:
New MvxRadioGroup control for Android, you can specify an ItemTemplate and bind the ItemsSource and SelectedItem properties

### DIFF
--- a/Cirrious/Cirrious.MvvmCross.Binding.Droid/Cirrious.MvvmCross.Binding.Droid.csproj
+++ b/Cirrious/Cirrious.MvvmCross.Binding.Droid/Cirrious.MvvmCross.Binding.Droid.csproj
@@ -72,6 +72,7 @@
     <Compile Include="Target\MvxAdapterViewSelectedItemPositionTargetBinding.cs" />
     <Compile Include="Target\MvxAutoCompleteTextViewPartialTextTargetBinding.cs" />
     <Compile Include="Target\MvxAutoCompleteTextViewSelectedObjectTargetBinding.cs" />
+    <Compile Include="Target\MvxRadioGroupSelectedItemBinding.cs" />
     <Compile Include="Target\MvxSeekBarProgressTargetBinging.cs" />
     <Compile Include="Target\MvxSpinnerSelectedItemBinding.cs" />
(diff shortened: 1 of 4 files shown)


iqbalmineraltown:
I'm sorry to necroposting, but I've stumbled upon the very same problem as @enterpredator 
and I can't find any thread discussing about this (or maybe I'm not searching hard enough)

I was trying to create border drawable and set it as RadioButton's  `background`

here's my findings:

I find that `AddView(View child, int index, ViewGroup.LayoutParams @params)` is called when adding radio buttons, but this `@params` variable is not what I've set in RadioButton template

I set `width` as `match_parent`, yet `@params` has `-2` or `wrap_content`

so my workaround is using `GetAttributeIntValue` from this constructor:
```
public MvxRadioGroup(Context context, IAttributeSet attrs, IMvxAdapterWithChangedEvent adapter)
            : base(context, attrs)
        {
            var itemTemplateId = MvxAttributeHelpers.ReadListItemTemplateId(context, attrs);
            if (adapter != null)
            {
                Adapter = adapter;
                Adapter.ItemTemplateId = itemTemplateId;
            }

            _layoutWidth = attrs.GetAttributeIntValue("http://schemas.android.com/apk/res/android", "layout_width", LayoutParams.WrapContent);
            _layoutHeight = attrs.GetAttributeIntValue("http://schemas.android.com/apk/res/android", "layout_height", LayoutParams.WrapContent);

            ChildViewAdded += OnChildViewAdded;
            ChildViewRemoved += OnChildViewRemoved;
        }
```
then I override `AddView(View child, int index, ViewGroup.LayoutParams @params)` like this
```
public override void AddView(View child, int index, ViewGroup.LayoutParams @params)
        {
            @params.Width = _layoutWidth;
            @params.Height = _layoutHeight;
            base.AddView(child, index, @params);
        }
```

kjeremy:
@iqbalmineraltown Don't hijack this ancient PR.  Please post an issue.

sumit4palz:
@softlion Can you share how to do that?

> RadioGroup from code or with childs statically added from xml layout will not work with the binding included in mvvmcross. It is not supported. you should write your own binding.

PR #582, opened 2014-02-04
Console + macviews

tofutim:
Stuart, I put the console nuget and macviews PR together, plus there is an additional change that fixes AutoLayout for Mac. Also, this changes the naming convention for nuget packages so that the platform is at the end. Can probably close out the other two PRs if you want.


martijn00:

PR #616, opened 2014-03-03
Added custom binding for UiSearchBar SelectedScopeButtonIndex property.

andbaciu:
The UISearchBar supports addding filtering criteria using scope buttons. I've saw that the binding for SearchText works (thanks to mverm) but not the one for selected scope button. With this changes now we can bind like this:

set.Bind(searchBar).To (vm => vm.SearchText);
set.Bind(searchBar).For(v => v.ScopeButtonTitles).To(vm => vm.SearchCriteriaNames);
set.Bind(searchBar).For(v => v.SelectedScopeButtonIndex).To(vm => vm.SelectedSearchCriteria);
set.Apply();


martijn00:
This can be done without adding it to MvvmCross

PR #684, opened 2014-05-07
WindowsStore file plugin, faster WriteFile

guillaume-fr:
The call to SafeDelete is slow when there is no file to delete (Exception handling).
The code can be done in one line using CreationCollisionOption and avoiding multiple access to the storage.
A similar improvement could be done on other implementations of this plugin to avoid multiple access to storage when only one is required.

Also, I think that ReadFile should catch only FileNotFoudException (DirectoryNotFoundException?) or maybe check FileExists intstead of catching all errors.


guillaume-fr:
Closing old PR, optimization included in this one https://github.com/MvvmCross/MvvmCross/pull/975 and working outside of LocalFolder.

PR #728, opened 2014-06-27
Fixed the memory used by the Touch plugin

softlion:
Fixed the memory used by the plugin. Changes are in ChoosePictureCommon of Touch


martijn00:

softlion:
Sry i won't :)

PR #730, opened 2014-06-30
Delete folder recursive for Windows Phone

pazi146:
A non-recursive implementation for recursive folder deletion on Windows Phone


slodge:
You wait **months** for Stuart to do something... and finally he does...

Thanks for the Pull :+1:

merged here - will be in 3.5 shortly

PR #735, opened 2014-07-06
Upgraded MvvmCross project to work with XamMac

brettyj:
Because mixing MonoMac and XamMac projects (if the project it's being used in is XamMac) causes problems.

Might be an idea to create a new branch with these changes, so that MonoMac can still be used by people who are using MonoMac?


martijn00:

PR #760, opened 2014-08-01
Navigation cache logic for WP8.1

promontis:
The navigation cache logic in WP8.1 seems to be changed.
Previously, the back button in WP8 worked without doing anything, preserving state and all.
In WP8.1 this has been revamped and caching is by default disabled: when pressing the back button it creates a new VM with a fresh new state.

I've researched this topic, but there isn't a lot of information available (too new I guess?). Some blog post suggest using the created NavigationHelper, but I think this is not the way to go... I would need to serialize the whole VM as state and then load it when needed. I've tried to fix this using a more generic approach, namely by setting the NavigationCacheMode to Required in the constructor of the MvxWindowsPage (base page for WP8.1). This will enable caching and the back button works as expected!

However, when I go from page A to B, back to A, back to B but with different parameters... it loads a cached version of B. This is correct IF we would press the back button, but otherwise I would expect it to load a new viewmodel with the different parameters. I've fixed this by checking if we are navigating to a new view, and if so, clearing the cached viewmodel so that MvvmCross instantiates it again.

### DIFF
--- a/Cirrious/Cirrious.MvvmCross.WindowsCommon/Views/MvxWindowsPage.cs
+++ b/Cirrious/Cirrious.MvvmCross.WindowsCommon/Views/MvxWindowsPage.cs
@@ -22,11 +22,6 @@ namespace Cirrious.MvvmCross.WindowsCommon.Views
     {
         private IMvxViewModel _viewModel;
 
-        protected MvxWindowsPage()
-        {
-            NavigationCacheMode = NavigationCacheMode.Required;
-        }
-
         public IMvxViewModel ViewModel
         {
             get { return _viewModel; }
--- a/Cirrious/Cirrious.MvvmCross.WindowsCommon/Views/MvxWindowsPage.cs
+++ b/Cirrious/Cirrious.MvvmCross.WindowsCommon/Views/MvxWindowsPage.cs
@@ -37,7 +37,7 @@ namespace Cirrious.MvvmCross.WindowsCommon.Views
 
         public void ClearBackStack()
         {
-            throw new NotImplementedException();
+            //throw new NotImplementedException();
             /*
             // note - we do *not* use CanGoBack here - as that seems to always returns true!
             while (NavigationService.BackStack.Any())
@@ -94,6 +94,11 @@ namespace Cirrious.MvvmCross.WindowsCommon.Views
                     nextPageIndex++;
                     nextPageKey = "Page-" + nextPageIndex;
                 }
+
+             
+
+               
+
             }
             else
             {
--- a/Cirrious/Cirrious.MvvmCross.WindowsCommon/Views/MvxWindowsPage.cs
+++ b/Cirrious/Cirrious.MvvmCross.WindowsCommon/Views/MvxWindowsPage.cs
@@ -22,6 +22,11 @@ namespace Cirrious.MvvmCross.WindowsCommon.Views
     {
         private IMvxViewModel _viewModel;
 
+        protected MvxWindowsPage()
+        {
+            NavigationCacheMode = NavigationCacheMode.Required;
+        }
+
         public IMvxViewModel ViewModel
         {
             get { return _viewModel; }
@@ -48,6 +53,11 @@ namespace Cirrious.MvvmCross.WindowsCommon.Views
 
         protected override void OnNavigatedTo(NavigationEventArgs e)
         {
+            if (e.NavigationMode == NavigationMode.New)
+            {
+                ViewModel = null;
+            }
+
             base.OnNavigatedTo(e);
 
             this.OnViewCreate(e.Parameter as MvxViewModelRequest, () => LoadStateBundle(e));
--- a/Cirrious/Cirrious.MvvmCross.WindowsCommon/Views/MvxWindowsPage.cs
+++ b/Cirrious/Cirrious.MvvmCross.WindowsCommon/Views/MvxWindowsPage.cs
@@ -104,11 +104,6 @@ namespace Cirrious.MvvmCross.WindowsCommon.Views
                     nextPageIndex++;
                     nextPageKey = "Page-" + nextPageIndex;
                 }
-
-             
-
-               
-
             }
             else
             {
--- a/Cirrious/Cirrious.MvvmCross.WindowsCommon/Views/MvxWindowsPage.cs
+++ b/Cirrious/Cirrious.MvvmCross.WindowsCommon/Views/MvxWindowsPage.cs
@@ -42,7 +42,7 @@ namespace Cirrious.MvvmCross.WindowsCommon.Views
 
         public void ClearBackStack()
         {
-            //throw new NotImplementedException();
+            throw new NotImplementedException();
             /*
             // note - we do *not* use CanGoBack here - as that seems to always returns true!
             while (NavigationService.BackStack.Any())


jonstoneman:
I'm a bit late on this one but just found it having hit problems using NavigationCacheMode=Required in a Universal app I'm working on.

My fix was also to add ViewModel = null as  @promontis proposed initially but I hadn't seen the binding problems then described in https://github.com/MvvmCross/MvvmCross/pull/760#issuecomment-52469800.  I've tested in both W8.1 and WP8.1.

Is the MyToolkit Frame really needed?  Am I misunderstanding the problem or your reproduction instructions @promontis?

promontis:
@jonstoneman: 

> I tried to solve this using NavigationCacheMode = NavigationCacheMode.Required;
> When I hit the backbutton the correct state is being loaded. So that works!

> However, when I go from page A to B and back to A (hitting the backbutton) and then again to B (by clicking on something that navigates to B) but with different parameters... it loads a cached version of B. > So for example page B can be a CustomerView which accepts a parameter named Id, which loads the customer with that Id. Navigating from page A to page B (with Id = 1), back to A and then again to B (with Id = 2), loads up the cached version of page B with Id = 1 instead of Id = 2.

Did you try this? So for example, you have a list of CustomerViewModel. You navigate to the customer with id A, it loads and you navigate back to the list. Then you navigate to the same viewmodel, but with a different user id, say B. It loads the view, but with the details of A instead of B.

Let me know if you can reproduce this

jonstoneman:
In my original app, it is navigating to a 2nd view when an item is selected within a list and using navigation parameters to specify the content to be shown in the second view.  The correct content is still being shown when the 2nd view is set to NavigationCacheMode = Required / Enabled.

The WP app hasn't been developed yet so I took a copy of the TipCalc Mvx sample, added two buttons to TipView which navigated to a new VM (SecondViewModel) using navigation parameters.  SecondView showed one of two TextBlocks using SecondViewModel properties with a BooleanToVisibilityConverter.  All pretty much the same as your "red and blue rectangles" suggestion.  This again is working fine even when NavigationCacheMode is set.

jonstoneman:
Just trying to think about what could be causing this to behave differently to what you have seen previously..  My app is using 3.5.1-beta1 and MvxSuspensionManager but TipCalc is still using 3.5.0.  TipCalc uses NavigationHelper but my app doesn't. 

*shrug*

promontis:
So, you are saying you can't reproduce this? Can you make a working sample on GitHub?

jonstoneman:
Correct.  Sample app now here: https://github.com/jonstoneman/NavingationCacheRepro

promontis:
Trying it...

promontis:
Seems to work! You should set caching to required on the TipView page aswell, otherwise, when you change same values (move the slider) on that page, navigate to the SecondView, and then back... you'll end up with a clean page.

I think it works because of the NavigationHelper. I'll try this weekend to see if my own app will work with how you use it. Will let you know!

jonstoneman:
Yeah, not worried about TipView - only did what was necessary for the repro.  :)

The real app is definitely not using NavigationHelper, it doesn't even exist.  It may be that it's needed for WP8.1 but W8.1 works fine without it.

promontis:
Yeah... AFAIK W8.1 has a different caching mechanism. Is your app W8.1?

jonstoneman:
It's universal but the WP part hasn't been developed yet.

promontis:
Ah I see! With W8.1 you should be fine without the NavigationHelper and the custom frame stuff.

jonstoneman:
Great, so we can set the ViewModel to null in the base views and recommend the use of NavigationHelper in WindowsPhone then?  The interface will still allow the mytoolkit frame to be used instead if people prefer.

promontis:
In my app (WP8.1) I'm not using the NavigationHelper. When I comment out the use of NavigationHelper in TipCalc, I see the back state is not being handled correctly. That is, when you navigate, press back, it exits the app; it seems the backstack is empty. 

The NavigationHelper probably sets this. However, I've not seen this behavior before: the problem I did have was that the viewmodel was not being updated. I think setting the ViewModel to null resolves this issue!!! 

I'll try and see what happens if I revert back to the normal Frame. 
I'll test two things:
1) backstack being filled when not using NavigationHelper.
2) viewmodel being set when navigating forward for the second time

This GitHub issue is only discussing 2. It seems that it is probably resolved by setting it to zero. I've never tried that, so I don't know yet. 

The use of NavigationHelper is recommended by MS though, so issue 1 should be resolved. 
I'll try to see if I can discover why everything is happening. Great job on finding out that setting the ViewModel to null helps.

promontis:
@jonstoneman Tested it in my own app and it seems to work! So the combination of NavigationHelper and setting the viewmodel to null. 

The NavigationHelper should be set on each page (View), so that the backstack is correctly set. If you don't set it on each page, it will 'skip'' that page when going backwards which isn't correct. For example, if you don't set it on the first page and you navigate from the first page to a second page, pressing back will result in exiting the app instead of going back to the first page.

Setting the viewmodel to null is needed when you set the NavigationCache to required. When you set the NavigationCache to required and when you navigate forward to a view already in the cache, it uses the old viewmodel (with its parameters) attached to the cached view instead of the viewmodel with its parameters you navigate to. By setting the viewmodel to null, it seems to force setting the correct viewmodel again.

@slodge: it seems the whole fork for this navigation cache thing is not needed anymore: shall I remove the fork and the related issues/pulls? The changes we introduced with the IMvxWindowsFrame is probably not needed anymore too, but I guess it is not a problem to keep them. It could however simplify things though.... what do you think?

ryanherman:
This has been driving me nuts, when to use NavigationCache and not...  Like having a page with some info then clicking a button to set a picture and going back only to see the viewmodel restored from when the view first launched.

promontis:
@ryanherman As a general rule, I think you want to use NavigationCache = Required when there is a possibility that you can navigate forward from that view. So, in your case, you have a button which does a forward navigation. 

Then, like I said in the previous comment, when you have set the NavigationCache to Required you'll need to set the ViewModel property to null when navigating forward to that view.

```
protected override void OnNavigatedTo(NavigationEventArgs e)
        {
            if (e.NavigationMode == NavigationMode.New)
                ViewModel = null;

            base.OnNavigatedTo(e);
            this.navigationHelper.OnNavigatedTo(e);
        }
```

I do think you can drop the check for NavigationMode.New. 
By setting the ViewModel to null, it seems the correct viewmodel is loaded. Otherwise, you'll see what your are describing; the viewmodel is restored from the cached view.

Let us know if that doesn't work!

ryanherman:
@promontis Now when I go to my view, the ViewModel isn't connecting.  Do I have to add the additions from your repo in WindowsCommon?

ryanherman:
@promontis I also set the ViewModel as such in my page:
        public new CreateCroudViewModel ViewModel
        {
            get { return (CreateCroudViewModel) base.ViewModel; }
            set { base.ViewModel = value; }
        }

ryanherman:
@promontis I changed the name of my ViewModel it something else, that seemed to load the ViewModel correctly on the New Navigation.  Testing out one more thing.

promontis:
@ryanherman you should just use the normal mvvmcross nuget libs. No additions have to be added (beside the NavigationHelper class if you don't have that... the new project wizard will add it)

In your code, you are wrapping the ViewModel to cast it to your concrete implementation. However, this is not the same as setting it to null. I hope you are doing that as well.

Do note that you don't really have to cast-wrap the viewmodel like that: https://github.com/MvvmCross/MvvmCross/blob/c306ba37afd9024f68b7a4f1fedcaf4cf7d01b8d/Cirrious/Cirrious.MvvmCross.WindowsCommon/Views/MvxWindowsPage.cs#L124

There already is a generic page class to specify a concrete viewmodel implementation.

promontis:
@jonstoneman @ryanherman @slodge I've edited the wiki to show the already created tutorial for the Universal Windows app created by @jonstoneman. I've also added the info learned here.

jonstoneman:
Good news that it worked for you too @promontis.  If you remove `NavigationCacheMode = NavigationCacheMode.Required;` from this PR then I think it should now be merged in.  

We'll have to remember to remove your note about setting the ViewModel to null in the tutorial once 3.5.2 is released...

promontis:
@jonstoneman I was thinking, should we perhaps add the NavigationHelper code to the MvxWindowsPage or are we then locking people in? I guess everyone will be adding a base class with the NavigationHelper code extending MvxWindowsPage, so why not add it directly to MvxWindowsPage? 

Oh and you are probably much more aware of new changes... could you enlighten me of the changes you are referring to in 3.5.2?

jonstoneman:
I had thought that this would be a bad idea given that NavigationHelper gets configured when a Basic Page is created or someone, not knowing it's in the base, could add it to a Blank Page manually.  I guess we could actually have the functionality and either disable it by default or auto-disable it if NavigationHelper has been configured on the page?

By "3.5.2" I just meant "the release that your PR is included in", not any other changes.  :)

promontis:
@jonstoneman Ah I see what you mean know :) lol... yeah we should change the tutorial when this PR is accepted. 

From a tutorial POV, I guess we shouldn't add the NavigationHelper code to the base class... almost all references (tutorials, samples, etc) say to make use of the NavigationHelper. I guess people will add it, and if they see it's a little cumbersome to add it each time, they'll create the extra base class themselves. Probably better then confusing them, right? 

We can just show them in the tutorial how things could be in such a base class, but that the default way of adding the NavigationHelper code to each page works too.

jonstoneman:
The tutorial should stay until 3.5.2 (or whatever) is actually released rather than when the PR is accepted as most people reading the tutorial will be using nuget rather than building from the 3.5 branch.  The good thing is that there won't be any problems for anyone following the tutorial (and adding the code in the views) when they come to upgrade Mvx.  The ViewModel will just get set to null twice.

With regards to NavigationHelper, I'm really not sure what's best but as it's not clear I'd leave it out of the base view just to keep it simpler.

promontis:
Just added the change. Users should set the NavigationCacheMode to Required if it's required themselves. Could you pull it and close this?

I'll delete the branch which was used to try out the mytoolkit frame as it seems to works now without it: https://github.com/MvvmCross/MvvmCross/pull/835. The issue was already closed, but the branch is still there! Cleaning up a bit :)

Do you guys have a todo list somewhere? Perhaps we could put the tutorial thingy on it?

jonstoneman:
I've pulled but will leave the branch for @slodge to delete.

ryanherman:
@promontis Loving Navigation Helper!  Works like a charm, sorry for the update!

promontis:
@ryanherman oh no problem... good to know it works for other people as well :)
@slodge if you find the spare time, you can delete the promontis-3.5 branch

slodge:
/me honestly I never really delete anything ... I'm a bit of a hoarder.... I also leave commented out code lying around (which is bad...)... and since you asked so nicely (and a few times...)... I'll go delete it :)

Good stuff :+1:

promontis:
Yeay... thnx for helping me clean up :) hehe... Normally, my girl cleans up around me, and if she's not there, she'll eventually yell at me when I don't... so probably just conditioned by her :P lol

alexsorokoletov:
Having same problem right now in UWP app. Looking into code I don't see similar check here.
https://github.com/MvvmCross/MvvmCross/blob/4.0/MvvmCross/Windows/WindowsUWP/Views/MvxWindowsPage.cs#L55

PR #771, opened 2014-08-22
Whoa! Fragments!

Cheesebaron:
I've spent some time fixing up Fragments and ViewModel rehydrating and saved states. This is only done in the Fragging library, but should be easy enough to pull out and put into the FullFragging library.

I've made a small sample showing it working here: https://github.com/Cheesebaron/MvxFragmentsAndHamburger there is a small readme telling what it does and what to expect.

Could someone review this, probably not the cleanest code, but it seems to work!


martijn00:
Since this is merged into 3.5, this PR can be closed.

davidschwegler:
Disclaimer: I'm sorry if this isn't the correct place to comment (if not, let me know where to post this kind of thing in the future). :)

I appreciate that support fragments were included in this release, but I just wanted to emphasize this.

As I see it, it's extremely important that support for Android.Support.V4.App.Fragment is *not* removed from MvvmCross. It's not about whether or not to support Gingerbread (we dropped support about a year ago for it), but the stability and feature set, for example http://stackoverflow.com/questions/17295497/fragment-or-support-fragment

A few of my reasons:
- There were several crashes in the Android 4.0 release of the built-in Android.App.Fragment system (in fact, there's even crashes in 4.2 https://code.google.com/p/android/issues/detail?id=42601).
- The ChildFragmentManager isn't supported in official fragments until Android 4.2, which is a very common use-case (at least for us). http://developer.android.com/reference/android/app/Fragment.html#getChildFragmentManager()
-  There were implementation fixes/tweaks in several releases since 4.x, and that have caused slightly different behavior in edge cases. As a result, knowing that you have the same implementation on all OS versions has made for far easier debugging. 
- When I talked to Roman Nurik at Google I/O this year, he said that support packages are not going away, but actually going to be using them as much as possible because of how challenging it is to get carriers/OEMs to publish OS updates in a timely manner.

PR #785, opened 2014-09-16
WindowsCommon PhoneCall Plugin improvement

jonstoneman:
Updated WindowsCommon PhoneCall Plugin to use tel: protocol on win81 and PhoneCallManager on wpa81


slodge:
I think this is merged :)

So closing....

Thanks :+1:

PR #795, opened 2014-10-08
Network Plugin, Using generic type in MvxRestClient for MvxRestResponse

voydz:
I just wrote some custom classes inheriting from MvxStreamRestResponse and I was pretty confused that I was not able to pass them as callback argument into `MvxRestClients' MakeRequest()`.

Long story short: I implemented the `MakeRequest()` method in a way that it will handle any custom implementation inheriting MvxRestResponse.

Also the line indention was pretty awkward, so I normalized and code formatted the whole `MvxRestClient' class file.


guillaume-fr:
Adding a virtual initialization method to `MvxRestResponse` ?

martijn00:
I am closing this for now. If still applicable you can reopen this pull request to the Plugins repo: https://github.com/MvvmCross/MvvmCross-Plugins

PR #817, opened 2014-10-28
Android Location to use Network provider

leobragatti:
When using getBestProvider, most of situations the provider selected will be the GPS, and I think it should work something like the iOS, which uses GPS and Network to get the location


martijn00:
I am closing this for now. The work can be continued on https://github.com/MvvmCross/MvvmCross-Plugins/pull/86

If still applicable you can reopen this pull request to the Plugins repo: https://github.com/MvvmCross/MvvmCross-Plugins

PR #827, opened 2014-11-07
Mvx exc hndl

alexeystrakh:


slodge:
Thanks. There does look to be some interesting things in here... but I'm not sure what the target of this is - so leaving it as "needs more info"

PR #830, opened 2014-11-12
Basic Silverlight support added.

milanjaros:
This seems to be stable support for Silverlight. 
There are still 10 TODOs but application runs, it can also use pages, bindings, etc.
Please check if it can be suitable for main branch or just for 3.1.


martijn00:

PR #835, opened 2014-11-20
Integrated mytoolkit custom frame within mvvmcross

promontis:


slodge:
Thanks - I'm going to close this one now without merging - as I think/hope the interface approach gives us a chance to replace the RootFrame.

If more is needed here - e.g. a refactoring of the interface then happy to do it - we should be able to do this without breaking anyone's apps - this is a new interface :)

PR #842, opened 2014-12-03
Fixed Bug where MvxCommand Parameter could not be parsed / casted, when ...

crazyfx1:
...the CommandParameter was defined in XAML on WP8 and WinRT.

I was unsure about the IMvxAutoConverter-thing, because all related classes where in the Binding-Project. That's why I chose to move the "core"-Implementation of MakeSafeValue to the Core-Project and call it in the Binding-Project. I did the same with "ConvertToBoolean", because I didn't want to create breaking changes.


Cheesebaron:
@martijn00 this needs to be merged into 3.5 as well

martijn00:
@Cheesebaron i checked, and it seems that it is already in.

PR #860, opened 2014-12-20
Added port for Xamarin.Mac Unified API.

lothrop:
I'd like to hear people's opinions on this. Since Mac OS X doesn't have the same strict deadline as iOS has I felt we should keep the Classic API version. This, however, leads to #if in some places.

The Xamarin.Mac Unified API projects are using the same code as the MonoMac projects and the Xamarin.Mac Classic API projects. Ideally, they might share code with the iOS implementations.

Also, file linking might not be the best approach now that we have shared projects. What are your thoughts?

Also, I'm not sure if I broke any old Mac project builds since I couldn't compile those.


lothrop:
Oops. I didn't intentionally merge. I actually wanted to push a different minor change...

Issue #550, opened 2014-01-13
WinStore: suspension changes to allow persistent serialisation/deserialisation


slodge:
Comment on: https://twitter.com/MikeFayer/statuses/422501923309113344

Gist: https://gist.github.com/tempywritescode/8391807

Also needs looking at in the light of win8.1 which has a different suspension manager setup

tal33:
Initial tests with W(P) 8.1 Universal Apps look good.

My code in App.xaml.cs now looks like this:
var setup = new Setup(rootFrame);
setup.Initialize();
var suspensionManager = Mvx.GetSingleton<IMvxSuspensionManager>() as MvxSuspensionManager;
Debug.Assert(suspensionManager != null, "suspensionManager != null");
suspensionManager.RegisterFrame(new MvxWrappedFrame(rootFrame), "AppFrame");

Is there a smarter way to call suspensionManager.RegisterFrame. As far as I can tell MvxWindowsSetup already has the rootFrame as the right kind of object (copied my code from there), but it is private.

I was even wondering why you didn't include RegisterFrame in MvxWindowsSetup.InitializeSuspensionManager?

tal33:
Update:
 The solution I described above did only work for one Resume/Restore cycle because there are multiple MvxWrappedFrame objects but the suspension managers need to use exactly the same frame object. I added 
public IMvxWindowsFrame RootFrame { get { return _rootFrame;} }
to MvxWindowsSetup in order to access this frame object and now it seems to work fine (but I don't know why you wanted to keep it private).

Personally I would add a new parameter to MvxWindowsSetup-constructor:
string suspensionManagerSessionStateKey=null) 
and if this isn't null I would call suspensionManager.RegisterFrame in MvxWindowsSetup.InitializeSuspensionManager
This way you have backwards compatibility and the option to initialize it manually but when specifying the stateKey everything is set up nicely automatically.

tal33:
I just created a pull request for this (hope it works this way - I'm new to GutHub ;-) also to show what I meant before
https://github.com/MvvmCross/MvvmCross/pull/867

bitdisaster:
I was searching around for Suspension/Restore strategy with Mvx but I coudl find much. What is recommend strategy?
I did a demo App with the latest version from Nuget. But the SaveAsync fails with come exception even if I just try to persist a string.

_suspensionManager.SessionState.Add("test", "that");
await _suspensionManager.SaveAsync();

tal33:
[...]

PR #867, opened 2015-01-11
Mvx550 windows suspension

tal33:
Fix for issue 550 (Windows suspension) as proposed in https://github.com/MvvmCross/MvvmCross/issues/550

Additionally added automatic suspensionManager.RegisterFrame since it needs the same _rootFrame object as MvxStoreSetup owns (and also to make it easier to use) with backwards compatibility (optional constructor parameter with default null)

Notes:
- also made the same changes to WindowsStore-classes without testing
-- maybe WindowsPhone-classes need this path, too?
- added a Windows-Only-Solution to facilitate building MvvmCross without Xamarin


slodge:
Thanks

I've been trying to work out how to test this...

I'm a bit tempted just to pull it in and then to let people scream if it breaks things... Im just holding off for now...

Wonder if @jonstoneman or any other Win-wizard has an opinion?

tal33:
I think I've tested it quite a bit for Universal Apps but I have no experience with classic Windows Store (WinRT 8) apps but since the original patch was suggested for that target I'm quite confident it'll work (I'm wondering who is still targeting WinRT Store Apps and not switching to Universal Apps anyways ;-)

Right now, if you register a SuspensionManager for Universal Apps your app just crashes. Can't get much worse :-)

BTW: I've uploaded a small Universal App for testing it (I've used Debugger-->Lifecycle Events, especially Suspend & Shutdown and then restarting it)

https://github.com/tal33/Foo/

slodge:
Thanks :+1: :+1: :+1: :+1:

Wondering what to do with the `Foo` sample - would be good to somehow publicise and reuse that sample :)

tal33:
Since my Foo-git repro is my universal playing field, I've created a branch with the latest version of that sample for this issue as a reference on how suspend/resume support in universal apps can be implemented:
https://github.com/tal33/Foo/tree/Demo-App-for-Mvx-Issue-550

I've updated the Foo sample a little bit:
- it references MvvmCross 3.5.1-alpha1 from nuget instead of included assemblies
- cleaned up app.xaml.cs and setup.cs-code to make it more readable

If I can help with the sample, please let me know. What I would suggest is that I update "_Windows Store UI.txt" with my version of app.xaml.cs as well as Setup.cs.pp with my version of Setup.cs. Maybe with defaulting to no suspend/resume support and only adding the relevant changes as comments.

PR #877, opened 2015-01-15
Do not swallow ViewModel Locator construction error if any

alexsorokoletov:
and make it available for ViewModelLoader

### DIFF
--- a/Cirrious/Cirrious.MvvmCross/ViewModels/MvxDefaultViewModelLocator.cs
+++ b/Cirrious/Cirrious.MvvmCross/ViewModels/MvxDefaultViewModelLocator.cs
@@ -16,21 +16,19 @@ namespace Cirrious.MvvmCross.ViewModels
         : IMvxViewModelLocator
     {
         public virtual bool TryLoad(Type viewModelType,
-                                    IMvxBundle parameterValues,
-                                    IMvxBundle savedState,
-                                    out IMvxViewModel viewModel)
+            IMvxBundle parameterValues,
+            IMvxBundle savedState,
+            out IMvxViewModel viewModel)
         {
             viewModel = null;
 
             try
             {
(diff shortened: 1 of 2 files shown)


alexsorokoletov:
@slodge  could you please look at the pull request? Does it make any sense to you?

Cheesebaron:
If you are always returning true anyways, does it even make sense to have a bool return type?

alexeystrakh:
Agree, and if you agree to that exception handling need to be changed than I can update returning type as well. @Cheesebaron should I proceed with that change as well?

Cheesebaron:
I guess this would be OK. However it will be a breaking change, so not sure if this should go into a new branch for the next version or if it goes into 3.5. @slodge?

slodge:
I could cope with the breaking change - and we could squeeze it into 3.5 if you're quick... but it might be better to preserve `IMvxViewModelLocator` as an interface (even if the default locator no longer really uses it)

Code notes:
- Might be useful to keep the inner exception - you can use `MvxWrap` to do that - https://github.com/MvvmCross/MvvmCross/blob/bbf9a2ac76e74d9404f4b57036c6e29dfe2cc6c3/CrossCore/Cirrious.CrossCore/Exceptions/MvxExceptionExtensionMethods.cs#L52
- If we were removing the `bool` then I guess it would no longer be a `Try` method and so it would no longer need an `out` parameter...

slodge:
I just got myself very confused trying to cherry-pick this commit into 3.5....

The reason is that this issue has already been addressed in 3.5 - see https://github.com/MvvmCross/MvvmCross/blob/3.5/Cirrious/Cirrious.MvvmCross/ViewModels/MvxDefaultViewModelLocator.cs

So... thanks for the PR - but the good news is that this is already done :)

alexeystrakh:
Thank you Stuart, looking forward to update to 3.5 @slodge

Issue #868, opened 2015-01-11
Windows Phone - MvxWindowsPage layout issues


raketenhunddev:
Hi there!

I started using MVVMCross a couple of days ago. I used it for my first W8/WP8 Universal App. I wondered about some layout issues but thought that there were changes between WP8 and WP8.1 I didn't pay attention to. Everything looked fine in Blend and the VS Designer. But when I started the app on a device or emulator some controls have moved.

Today there was another strange layout problem and I tried to find the reason for it. At least I found out that when I copy the content of my MvxWindowsPage to a view with the default Page class, the controls also moved in the designer like it looked on my device. So i placed the controls to their right positions, copied the Content of the Page based view (where everything looks fine) to the MvxWindowsPage based view and finally the layout looked "destroyed" in blend. But when I deploy it on a device it looks like it should!

EDIT: While I was creating a sample for you I found out that this error occurs when using the BottomAppBar

Here is a sample page where this layout issue occurs. Change it's base class (and also the Page.BottomAppBar) to views:MvxWindowsPage and you'll see, that the third row of the root grid is missing.

````XML
<Page
    x:Class="Test.Universal.Views.BlankPage1"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:local="using:Test.Universal.Views"
    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
    mc:Ignorable="d"
    xmlns:views="using:Cirrious.MvvmCross.WindowsCommon.Views"
    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
    
    <Page.BottomAppBar>
        <CommandBar>
            <AppBarButton Icon="Back" Label="Previous Month" Command="{Binding PreviousMonthCommand}"/>
            <AppBarButton Icon="Forward" Label="Next Month" Command="{Binding NextMonthCommand}"/>
        </CommandBar>
    </Page.BottomAppBar>

    <Grid>
        <Grid.Background>
            <SolidColorBrush Color="DarkGray"/>
        </Grid.Background>
        
        <Grid.RowDefinitions>
            <RowDefinition Height="50"/>
            <RowDefinition/>
            <RowDefinition Height="50"/>
        </Grid.RowDefinitions>
        
        
[...]

PR #879, opened 2015-01-16
Fixed ViewModel property in MvxStorePage<T>.

lothrop:
It seems the implementation of MvxStorePage<T> was wrong. @slodge, do you know why the implementation was different than MvxPhonePage<T>?


slodge:
Looking at https://github.com/MvvmCross/MvvmCross/pull/808 I think it was just sticky fingers in the cut&paste 
- the store page was in https://github.com/Stephanvs/MvvmCross/commit/27792922a5b8bfafcbf39dd1400fecb0f36dcc74
- most of the other classes (not interfaces) were in https://github.com/Stephanvs/MvvmCross/commit/f1c7d13f86da8f856263f65f3cf194c437cff72e

While we're in the neighbourhood... should we drop the `abstract` from the pages (linked to some weird Blend feature - #868)

PR #885, opened 2015-01-19
Fixed typo in MvxSeekBarProgressTargetBinding

martijn00:


Cheesebaron:
Target Binging sounds much more fun than Binding, but ok...

PR #898, opened 2015-01-29
Location plugin provides event to track permission now

SeeD-Seifer:
IMvxLocationWatcher has an event OnPermissionChanged now.

New enum introduced:
```csharp
public enum MvxLocationPermission {
	Unknown,
	Denied,
	Granted
}
```

iOS and Android plugins tested for sure. See https://github.com/SeeD-Seifer/MvvmCross-Playground
Win plugins are not tested (and even not complied). But according to documentation they should compile and work well (I hope).

Basic usage example:
```csharp
locationWatcher.OnPermissionChanged += (object sender, MvxValueEventArgs<MvxLocationPermission> args) => {
	var newPermission = args.Value;
}
```

**The reason of new feature**:
One of our iOS app has a flow on startup with few UIAlerts.
Each alert should wait for dismiss before the next one appears.
One of such alerts is Location Request.
The new OnPermissionChanged event allows to handle the case when user has pressed a button on the Alert, and so that the flow could be continued smoothly.

### DIFF
--- a/Plugins/Cirrious/Location/Cirrious.MvvmCross.Plugins.Location.Droid/MvxAndroidGeoLocationWatcher.cs
+++ b/Plugins/Cirrious/Location/Cirrious.MvvmCross.Plugins.Location.Droid/MvxAndroidGeoLocationWatcher.cs
@@ -155,7 +155,7 @@ namespace Cirrious.MvvmCross.Plugins.Location.Droid
 
         public void OnProviderDisabled(string provider)
         {
-            SendError(MvxLocationErrorCode.PositionUnavailable);
+            SendError(MvxLocationErrorCode.ServiceUnavailable);
         }
 
         public void OnProviderEnabled(string provider)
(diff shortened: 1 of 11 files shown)


slodge:
Thanks :+1:

Not sure I personally like the `Action<T>` format for the `event` delegate (I'm old fashioned me... I use EventHandler - there is an Mvx Value based one somewhere...)

On the permissions themselves, I'll tweet it out - see if anyone has any platform-level comments - better to get these now than after people are using the interface :)

SeeD-Seifer:
I've considered EventHandler as an option and seems something confused me to use Action<T>.
I'll prepare this change set.

Yeah, would be great to get any comments/recommendations. Hopefully we can keep this extension as part of core plugin (I really need this functionality :) and would be great to keep using pure Mvx plugins)

slodge:
Hmmm..... well, I didn't do so well on attracting comments - sorry...

Let's pull this... :)

Wonder who we can get to test the Network plugin still works in Windows?

PR #903, opened 2015-02-02
MvxExpandableListView, adapter and selected item binding

kjeremy:
This is based on code from kjeremy's internal project code
which in turn is based on
https://stackoverflow.com/questions/15010246/mvvmcross-expandablelistview


slodge:
Thanks :)

Note for future trouble-squashing - have confirmed on chat that @kjeremy is happy to contribute this for Mvx to license (currently MS-PL, future could be Apache 2 or other!)

Issue #884, opened 2015-01-19
Change Seekbar subscriber to use an event (not a listener)


slodge:
See discussion in http://stackoverflow.com/questions/28030458/mvvmcross-binding-android-eventhandler/28031151#28031151 (I've not checked the Xam API but I believe my answer should work...)

Sylapse:
I have created a new binding which uses the ProgressChanged event handler and solves the problem stated on Stack Overflow.

However I was unable to successfully register the binding when subclassing

    MvxPropertyInfoTargetBinding<SeekBar>

and registering it using 

    protected override void FillTargetFactories(Cirrious.MvvmCross.Binding.Bindings.Target.Construction.IMvxTargetBindingFactoryRegistry registry)
	{
		base.FillTargetFactories(registry);
		registry.RegisterPropertyInfoBindingFactory(typeof(MySeekBarProgressTargetBinding), typeof(SeekBar), "Progress");
	}


No break points in the binding were hit. But using the following base class and registration works fine.

    MvxAndroidTargetBinding
    ...
    protected override void FillTargetFactories(Cirrious.MvvmCross.Binding.Bindings.Target.Construction.IMvxTargetBindingFactoryRegistry registry)
	{
		base.FillTargetFactories(registry);
		registry.RegisterCustomBindingFactory<SeekBar>("Progress",	seekBar => new MySeekBarProgressTargetBinding(seekBar));
	}

Not sure why the MvxPropertyInfoTargetBinding seems to fail to register but the MvxAndroidTargetBinding works? Is it preferable to use one or the other?

Sylapse:
I also notice that in the MvxAndroidTargetBinding version that I have made, I must provide a binding name other than "Progress" in order for it to work. If I use "Progress" then the custom binding is never instantiated. The same is not true for subclassing MvxPropertyInfoTargetBinding<SeekBar>, I can't get that to instantiate either way. I have tried registering them before the core registers its implementations but made no difference.

PR #905, opened 2015-02-04
SeekBar binding to use event subscription instead of on change listener

Sylapse:
Issue #884

Allows user to apply additional event handlers without breaking the mvx
binding.


Sylapse:
Fixed. Must have been a Mac thing!

slodge:
Thanks :+1:

Issue #878, opened 2015-01-15
Possible need for more binding auto-conversion for new unified numeric types


slodge:
Question seen on jabbr

FabienM
Hi guys
10:26:03 AM
Question: in a unified app that has nint, nfloat, ... all this
10:26:20 AM
if you have some mvx bindings on iOS
10:26:41 AM
like :
10:26:50 AM
set.Bind(_pageControl).For(p => p.CurrentPage).To(vm => vm.ImageIndex);
10:27:21 AM
here the CurrentPage is a nint
10:27:31 AM
and my existing code ImageIndex is an int
10:27:39 AM
so at runtime it fails

Could suggest we need to add code to automatically convert between int and nint

lothrop:
I'm wondering if we should provide a value converter or do some implicit conversions.

softlion:
Should be implicit, or use the default .NET TypeConverter class and register a typeconverter from int to nint, and use TypeConverter in the binding code (like WPF does).

softlion:
public class MvxIntToNIntTargetBinding<T> : MvxTargetBinding
    {
        private PropertyInfo _targetPropertyInfo;

        public MvxIntToNIntTargetBinding(object target, PropertyInfo targetPropertyInfo) : base(target)
        {
            _targetPropertyInfo = targetPropertyInfo;
        }

        public static void Register(IMvxTargetBindingFactoryRegistry registry)
        {
            //Must add all target types and properties here ....
            registry.RegisterFactory(new MvxSimplePropertyInfoTargetBindingFactory(typeof(MvxIntToNIntTargetBinding<UILabel>), typeof(UILabel), "Lines"));
            registry.RegisterFactory(new MvxSimplePropertyInfoTargetBindingFactory(typeof(MvxIntToNIntTargetBinding<UIView>), typeof(UIView), "Tag"));
        }

        public override void SetValue(object value)
        {
            //Convert int to nint
            var target = Target;
            if (target != null && value != null)
            {
                try
                {
                    _targetPropertyInfo.SetValue(target, (nint) (int) value);

                }
                catch (Exception)
                {
                    Mvx.Error("Can not convert {0} to {1} in {2}.{3}", value.GetType().Name, _targetPropertyInfo.PropertyType.Name, typeof(T).Name, _targetPropertyInfo.Name);
                }
            }
        }

        protected override void FireValueChanged(object newValue)
        {
            //Convert nint to int
            base.FireValueChanged((int)newValue);
        }

[...]

PR #906, opened 2015-02-05
Added IMvxSafeValueCreator

nitescua:
Added IMvxSafeValueCreator which handles value conversation


nitescua:
MvvmCross/MvvmCross#878

slodge:
I'm a bit unsure what to do on this one...

It's basically "just refactoring" away from a couple of layers of statics to one layer of statics and an interface?

What is this hoping to achieve here? Can we add some commentary here about why this should happen? Does another part need adding to achieve some result for #878?

Also are there any side effects to especially worry about? e.g. I'm wondering about anyone using `CrossLight` binding without MvvmCross...

nitescua:
yes, it's probably a bad idea

PR #913, opened 2015-02-17
Unified enum-compatible segmented control with default binding

TheAlmightyBob:
Trying to bind UISegmentedControl to an enum previously resulted in failing to convert an nint to that enum (in MakeSafeValueCore) and throwing an exception.
Also I thought a default binding for SelectedSegment might be nice.

Possibly MakeSafeValueCore could be made more Unified-friendly, but this was the easy/safe fix...


slodge:
Thanks for this :+1:

Definitely agree we need some `n*` conversions going on somewhere... just don't know where yet..

Issue #910, opened 2015-02-12
Fragment regression since 3.5


kjeremy:
I have a viewpager which works below 3.5 but breaks with 3.5 and up.  If you clone https://github.com/kjeremy/FragmentStatePagerAdapterCrash and swipe the viewpager over to the end and then back it will crash when trying to set the viewmodel.  Note that if you remove the ChildFragment assignment it won't crash.

Maybe a fix for #908 would help this.

davidschwegler:
Yes, this is the same bug AFAIK. If you decide to work around it, I'd be curious on your solution.

kjeremy:
I'm not sure how I should work around it.  I've tried a number of things that don't work.

davidschwegler:
#915 has been merged in, which will fix this problem for you. I think this case can be closed.

PR #915, opened 2015-02-20
Require opt-in for Fragments' ViewModel create/save/restore lifecycle being owned by MvvmCross.

### LINKED ISSUE #908: MvxFragments auto-creating their ViewModels causes crashes in 3.5 when restored from saved state
We need a way to opt-out MvxFragments from auto-looking up/creating their ViewModels.

**Bug**
Previously, you could do something like this:

```
var fragment = new MyFragment();
fragment.ViewModel = myViewModel;

// show the fragment
...
```

However, this crashes in 3.5 when the fragment is restored from saved state.

**Cause**
In 3.5, new code in the MvvmCross system looks for/creates a ViewModel for MvxFragments when OnCreate is called. If there is not one, it will crash. This appears to happen if there are any arguments supplied or if the fragment is restoring from saved state: https://github.com/MvvmCross/MvvmCross/blob/3.5/Cirrious/Cirrious.MvvmCross.Droid.Fragging/Fragments/MvxBindingFragmentAdapter.cs#L37

You currently cannot override this behavior as the parent/host of the fragment, AFAIK, because the child fragment viewmodel lookup is executed when you call base.OnCreate(), and your own ViewModel is not yet set at that point.

**Requested Change**
Opting out from this most simply, I believe, would just involve skipping the calls to HandleCreateCalled() and HandleSaveInstanceStateCalled(). This would restore the pre-3.5 behavior and allow apps to continue using manual viewmodel assignments. 

I'd be fine with [MvxUnconventional] opting out, though that might be a stretch of it's meaning.

That said, it would also be nice to put some additional thought into how to tweak the 3.5 system to support things like fragment pagers, eg by adding an additional option to support saving/reloading but not caching &#91;or initially creating? ... not sure&#93; (since there will be many of the same type).

**Rationale**
In addition to things like #846, there are many scenarios where fragments are used to host or divide up the screen as part of the Android-specific view, where there is no applicable distinct viewmodel, as well as scenarios where the viewmodel must be programmatically set (eg fragment pagers, which are currently broken in 3.5).

davidschwegler:
This adds a MvxOwnedViewModelFragment attribute, which is required to enable 3.5 fragments create/save/restore behavior. Without this attribute, the pre-3.5 behavior exists.

See #908 for additional discussion.
Fixes #908 and #910

To use 3.5 behavior:
```
[MvxOwnedViewModelFragment]
public class MyFragment : MvxFragment
{
    ...
}
```

To use pre-3.5 behavior:
```
public class MyFragment : MvxFragment
{
    ...
}
```


Cheesebaron:
Looks good to me

slodge:
Thanks! :+1:

slodge:
Merging locally (as seems like it can't merge automatically)

PR #916, opened 2015-02-20
Added support for storyboard views

geirsagberg:
Usage: 

    // Will look for a UIViewController with identifier "MyView" inside a Storyboard named "MyView.storyboard"
    [MvxFromStoryboard]
    public class MyView : MvxViewController
    {
        public MyView(IntPtr handle) : base(handle) {}
    }
 
    // Will look for a UIVIewController with identifier "MyOtherViewInSameStoryboard" inside a Storyboard named "MyView.storyboard"
    [MvxFromStoryboard(StoryboardName = "MyView")]
    public class MyOtherViewInSameStoryboard : MvxViewController
    {
        public MyOtherViewInSameStoryboard(IntPtr handle) : base(handle) {}
    }

### DIFF
--- a/Cirrious/Cirrious.MvvmCross.Touch/Views/MvxTouchViewsContainer.cs
+++ b/Cirrious/Cirrious.MvvmCross.Touch/Views/MvxTouchViewsContainer.cs
@@ -42,11 +42,19 @@ namespace Cirrious.MvvmCross.Touch.Views
         protected virtual IMvxTouchView CreateViewOfType(Type viewType, MvxViewModelRequest request)
         {
             var storyboardAttribute = viewType.GetCustomAttribute<MvxFromStoryboardAttribute>();
-            if (storyboardAttribute != null) 
+            if (storyboardAttribute != null)
             {
-                var storyboard = UIStoryboard.FromName(storyboardAttribute.StoryboardName ?? viewType.Name, null);
-                var viewController = storyboard.InstantiateViewController(viewType.Name);
-                return (IMvxTouchView) viewController;
+                var storyboardName = storyboardAttribute.StoryboardName ?? viewType.Name;
+                try
+                {
+                    var storyboard = UIStoryboard.FromName(storyboardName, null);
+                    var viewController = storyboard.InstantiateViewController(viewType.Name);
+                    return (IMvxTouchView) viewController;
+                }
+                catch (Exception ex)
+                {
+                    throw new MvxException("Loading view of type {0} from storyboard {1} failed: {2}", viewType.Name, storyboardName, ex.Message);
+                }
             }
 
             var view = Activator.CreateInstance(viewType) as IMvxTouchView;
--- a/Cirrious/Cirrious.MvvmCross.Touch/Views/MvxTouchViewsContainer.cs
+++ b/Cirrious/Cirrious.MvvmCross.Touch/Views/MvxTouchViewsContainer.cs
@@ -6,9 +6,11 @@
 // Project Lead - Stuart Lodge, @slodge, me@slodge.com
 
 using System;
+using System.Reflection;
 using Cirrious.CrossCore.Exceptions;
 using Cirrious.MvvmCross.ViewModels;
 using Cirrious.MvvmCross.Views;
+using UIKit;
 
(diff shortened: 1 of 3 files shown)


slodge:
This looks fab :+1:

Wondering if there is any way we can enumerate the storyboards in the project - but I couldn't spot anything in the NSBundle reference...

slodge:
Another idea - we could provide a `Storyboards` virtual property or method on the Setup for Touch - or we could have a `StoryboardRegistry` of some description - this would then allow the user to specify a list of default storyboard files to try for all views?

Cirrious/Cirrious.MvvmCross.Touch/Views/MvxTouchViewsContainer.cs
```diff
@@ -39,6 +41,14 @@ public virtual IMvxTouchView CreateView(MvxViewModelRequest request)
 
         protected virtual IMvxTouchView CreateViewOfType(Type viewType, MvxViewModelRequest request)
         {
+            var storyboardAttribute = viewType.GetCustomAttribute<MvxFromStoryboardAttribute>();
+            if (storyboardAttribute != null) 
+            {
+                var storyboard = UIStoryboard.FromName(storyboardAttribute.StoryboardName ?? viewType.Name, null);
+                var viewController = storyboard.InstantiateViewController(viewType.Name);
+                return (IMvxTouchView) viewController;
```
slodge: if `storyboard` or `viewController` is null then should we throw an exception here - this won't help at runtime - but it would help any developer who is trying to debug why their view doesn't load?

cyrilcathala:
That's interesting and very welcome!
But like Stuart said, I'd rather opt for some kind of registration instead, just to avoid creating several MvxViews for lookup purpose only.

Cirrious/Cirrious.MvvmCross.Touch/Views/MvxTouchViewsContainer.cs
geirsagberg: The problem is that when a custom exception handler is registered (e.g. HockeyApp, TestFlight) line 47 crashes the app if the storyboard is missing; try/catch doesn't help. In other words, the storyboard MUST exist in order for the app not to crash. I will add a try/catch anyway though, for those who are not using custom exception handlers.

geirsagberg:
@cyrilcathala not sure what you mean by creating several MvxViews for lookup purpose only; the attribute is meant to be placed on your already existing view. Can you clarify the problem?

geirsagberg:
I have not yet found a safe way to determine if a storyboard exists, or if a viewcontroller with a given identifier is inside. Try/catch would seem to be the workaround, but if a custom exception handler is registered even this will fail; also see http://stackoverflow.com/questions/28598392/monotouch-exception-not-getting-caught-on-device.

geirsagberg:
By the way, originally I was going for `[FromStoryboard]` as it looks nicer; how strict is the `Mvx`-prefix enforcement?

cyrilcathala:
Well I though you asked to create a MvxView for each view you want to register with its associated storyboard. I guess you can tag your MvxViewController too, which is a better way I think.

geirsagberg:
Ah yeah my bad, I thought the MvvmCross views were MvxViews for a second, they should of course be MvxViewController. I have updated the usage examples.

hollywoof:
Merging this in - thanks

I'm torn on the FromStoryboard vs MvxFromStoryboard thing... I always hate libraries which have smurf-prefixes... but then I forked this codebase originally from monocross with its Mx naming :/

dazinator:
Can this new attribute be used with Xamarin IOS "Universal" projects - i.e where there are 2 different storyboards (1 for IPhone, and another for IPad) and the correct storyboard to use is determined at runtime based on the type of device?

Stephanvs:
There's also a `MvxFormFactorSpecificAttribute`, that was meant for this purpose, but I'm not sure whether this still works/works in combination with this `MvxFromStoryboardAttribute`.. :eyes:

geirsagberg:
@dazinator I am honestly not sure, haven't tried it. Should be easy to test, drop a comment with your results if you find out :)

PR #920, opened 2015-02-26
I made changes to MvxTouchSetup and MvxTouchDialogSetup to take a new in...

Bowman74:
...terface, IMvxApplicationDelegate instead of the concrete MvxApplicationDelegate class.  This allows Mvx to be used with other frameworks that have their own ApplicationDelegate that must be used such as Xamarin Forms 1.3.X+.  Users of Mvx and Xamarin Forms can now create their own application delegate that inherits from FormsApplicationDelegate and pass it to MvxTouchSetup as long as their custom application delegate also implements the IMvxApplicationDelegate interface.

### DIFF
--- a/Cirrious/Cirrious.MvvmCross.Dialog.Touch/MvxTouchDialogSetup.cs
+++ b/Cirrious/Cirrious.MvvmCross.Dialog.Touch/MvxTouchDialogSetup.cs
@@ -17,13 +17,13 @@ namespace Cirrious.MvvmCross.Dialog.Touch
     public abstract class MvxTouchDialogSetup
         : MvxTouchSetup
     {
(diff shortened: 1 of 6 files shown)


lothrop:
This introduces a minor breaking change for those accessing the `protected` properties. @slodge, what do you think?

https://twitter.com/bowman74/status/570566423556329472

slodge:
It looks might fine to me :+1:

I'm not too worried about the `protected MvxApplicationDelegate ApplicationDelegate` - as breaking changes go I agree this is pretty minor (famous last words :)) - and the change definitely looks like an improvement - definitely in line with the manifesto :)

------

PS My vote is an extension library... possibly bundled as an alternative set of starterpack nuspecs... the solution I like most is where Mvx is open to extension, change and even replacement :)

Issue #928, opened 2015-03-05
Missing constructors (JniHandleOwnership, mono)


JanZeman:
Hello, our MvvmX based app is crashing on freshly updated Android Lollipop with this exception: http://stackoverflow.com/questions/26573627/no-constructor-found-for-system-intptr-android-runtime-jnihandleownership

I've forked MvvmX and added these constructors to the corresponding classes - will create a pull request. Thank you, Jan

JanZeman:
PR created: https://github.com/MvvmCross/MvvmCross/pull/929

Cheesebaron:
This has been merged, thanks :+1:

PR #929, opened 2015-03-05
Missing mono-related jni constructors

JanZeman:
Hello this is my first PR for MvvmX, trying to fix issue https://github.com/MvvmCross/MvvmCross/issues/928

Our MvvmX based app was crashing on freshly updated Android Lollipop with this exception: http://stackoverflow.com/questions/26573627/no-constructor-found-for-system-intptr-android-runtime-jnihandleownership, so this is my attempt to fix those. I have previously done it on some other non-MvvmX controls and I can confirm that the crashes stopped appearing on those controls. To be able to add jni constructors on other our classes derived from MvvmX I need MvvmX defines those. Thank you, Jan


Cheesebaron:
Looks good to me. Thanks a bunch for this Jan :+1:

PR #931, opened 2015-03-06
RunOnBackgroundThread methods are added to ThreadDispatcher

evnik:
Set of methods which allow to run a task or method in a new background thread or continue in current if it is already background.


slodge:
Thanks :+1: 

***Thinking*** We might have to delay this to a minor point release... just worried that adding code to `IMvxMainThreadDispatcher` will break people's test code.

Alternatively, I wonder if there's some other way this could be introduced... (not sure - maybe it could be introduced via a new interface?)

Will think about it some...

@gshackles you write lots of test code... what do you think is the right time to break people's code?

gshackles:
Only glanced at this quickly but it looks like this is really just adding new code rather than changing existing behavior. What do you see being problematic here from a testing perspective? Looks like the main breaking change would be the removal of `CallbackMockMainThreadDispatcher`

martijn00:
Since this has been waiting for a long time, i'm going to close it. If this still is a valid addition please open a new pull request targeting the develop branch, and we would be happy to review it! Thanks for contributing anyway!

Issue #933, opened 2015-03-08
MvxDynamicImageHelper not reset when recycling view/cell during scroll


PaulLeman:
I use a MvxImageView in a MvxListView or in a UITableView with the bindings "ImageUrl" and "DefaultImagePath" for profile images. Sometime DefaultImagePath is used because ImageUrl has not been set . During scroll, Layouts/Cells are re-used and it may be that image that uses "ImageUrl" then uses "DefaultImagePath". I encounter an unexpected behavior because the field "_currentImageState" of the "MvxDynamicImageHelper<T>" class is not reset when recycling View/Cell. The same behavior happen with "ErrorImagePath".

martijn00:
Fixed in: https://github.com/MvvmCross/MvvmCross/pull/934

PR #934, opened 2015-03-08
MvxDynamicImageHelper not reset when recycling view/cell during scroll

PaulLeman:
#933


slodge:
Thanks :+1: Good find and debug!

PR #935, opened 2015-03-10
Fixed passing of some tests on systems with non-english culture

evnik:
Some tests can be failed on systems with non-english culture due to unexpected formatting of numbers. Setting of InvariantCulture to the test thread solves this issue.


slodge:
Thanks :+1:

PR #938, opened 2015-03-10
Support open generic types in MvxBindingNameRegistry

Bognar:
Currently, adding a binding name only works with a closed generic type. For example, this will work for any MyView\<int, string\>:

    registry.AddOrOverwrite(typeof(MyView<int, string>), "ItemsSource");

however, this will not work for any MyView\<T1, T2\>, i.e. using the open generic type:

    registry.AddOrOverwrite(typeof(MyView<,>), "ItemsSource");

This commit adds support for checking the open generic type if the closed type is not found. This allows a default for a given open generic type, and an override for a specific closed type.


slodge:
Thanks

Issue #932, opened 2015-03-06
FloatElement not properly styled


henrik-d:
Hey,
I'm using dialogs in my Touch-Project. Works great, but the FloatElement is not correctly styled (see screenshot).

In the original implementation of MonoTouch.Dialog the slider's top-value is calculated as following:
`UIDevice.CurrentDevice.CheckSystemVersion (7, 0) ? 18f : 12f;`

In the CrossUI-Version it's always `12f`! Would be great, if that could be fixed.
![ios simulator screen shot 06 03 2015 20 18 39](https://cloud.githubusercontent.com/assets/6033679/6532444/0f7f8f20-c43f-11e4-88c9-3b2bec6364fe.png)

(it's only about the missing margin, I set divergent the font myself)

slodge:
This sounds good - sounds like a change that happened after we forked the code.

Thanks

I'll make this change - if you have further requests, then the best (quickest) way is to make a Pull Request.

slodge:
Actually, I've just looked at https://github.com/MvvmCross/MvvmCross/blob/3.5/CrossUI/CrossUI.Touch/Dialog/Elements/FloatElement.cs and have no idea which line your change applies to.

It'll be much quicker if you make the change :)

Thanks

Cheesebaron:
I suspect it is this one, but not sure at all: https://github.com/MvvmCross/MvvmCross/blob/3.5/CrossUI/CrossUI.Touch/Dialog/Elements/FloatElement.cs#L108

henrik-d:
Yes that's the line!

Ok, next time I'll try a pull request, although I never have done it before :$

Cheesebaron:
This was merged. Closing.

PR #940, opened 2015-03-13
Fix for #932

Cheesebaron:
I haven't compiled this :smile:


slodge:
If it checks into source control, ship it :)

PR #941, opened 2015-03-17
DeselectRow support in base UITableViewSource

alexsorokoletov:
@slodge what about adding "DeselectRow" to base MvxTableViewSource class so no one need to write this code in the app ever. From our experience, it's only small amount of apps don't remove selection in row after it's been done.

### DIFF
--- a/Cirrious/Cirrious.MvvmCross.Binding.Touch/Views/MvxBaseTableViewSource.cs
+++ b/Cirrious/Cirrious.MvvmCross.Binding.Touch/Views/MvxBaseTableViewSource.cs
@@ -35,7 +35,7 @@ namespace Cirrious.MvvmCross.Binding.Touch.Views
             get { return _tableView; }
         }
 
-        public virtual bool DeselectAutomatically { get { return false; } }
+        public bool DeselectAutomatically { get; set; }
 
         public ICommand SelectionChangedCommand { get; set; }
 
--- a/Cirrious/Cirrious.MvvmCross.Binding.Touch/Views/MvxBaseTableViewSource.cs
+++ b/Cirrious/Cirrious.MvvmCross.Binding.Touch/Views/MvxBaseTableViewSource.cs
@@ -35,12 +35,7 @@ namespace Cirrious.MvvmCross.Binding.Touch.Views
             get { return _tableView; }
         }
 
-        /// <summary>
-        /// Gets a value indicating whether this
-        /// <see cref="Cirrious.MvvmCross.Binding.Touch.Views.MvxBaseTableViewSource"/> should deselect row automatically after selection.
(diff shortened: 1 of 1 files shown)


slodge:
I'm not sure the "only small amount of apps" is true. I've written a few split view apps with master-detail which used selection - that can't be very uncommon?

However, I'd be happy to see an option put in... and the default state can be argued about :)

Some questions/points:

1. I'd argue not to include the `/// <summary>` comment - there aren't any other comments in the project so I don't know if adding them adds value. Also adding them does add maintenance overhead - it's so very easy for comments to become lies... ;)

2. Maybe the deselect should only fire if `Command` is `null` (not sure about this though)

3. Would a `public virtual {get;set;}` property be better than a `protected virtual {get;}` property?

Not sure on any of those 3 questions... I'm just asking/talking :)

alexsorokoletov:
I see your point, well, usually people use those UITableView effects. 
Regarding 1, I was also doubtful but decided to add it. It's easy to remove, if that is a general practice for MvvmCross. 

2. Well, it's more about the UI experience - if you selected row, you need to deselect it back or it stays there until you select another. So, in master-detail example, you select an item, it's highlighted, you go to a next page, then go back and the item is still selected. 
If you don't want to have any selections, you can just use Selection = NoSelection property on TableView itself in storyboard/xib

3. Also was thinking of making it's either public or protected. Public virtual is a good balance, I think.

If the request is not really useful from your point of view, then let's not add any additional code to the codebase. I wanted to fix a small thing that bothers me around for last 5 projects.

slodge:
Thanks

On the Master-Detail, I was thinking more of the **ipad** apps I've done with split view - where "going to the next page" means changing what's in the right-hand pane. In these apps, having the left-hand selected was important.

Happy to add this - but suspect the default should be `true` - as otherwise this might burn other developers. No-one likes their app breaking when they update nuget (or xamarin or...)

Stuart

alexsorokoletov:
you mean "false"?

slodge:
> you mean `false`?

:) Thx - I'm glad one of us is awake

alexsorokoletov:
:+1:

slodge:
Final question... should it just be `get;set;` instead of (or as well as) `virtual`?

alexsorokoletov:
Glad both of us are awake. If it's public, it should be get;set; and maybe with virtual to override default value?

slodge:
OK... I can definitely live with that :+1:

Sorry for so many questions :)

Will leave open just for a few hours... or maybe a day (until I go home and do a merge)

Thanks :+1:

alexsorokoletov:
Sure, thanks for the feedback! I forgot about iPad apps, actually ;)

PR #947, opened 2015-03-30
MvxAndroidLocalFileImageLoader: implemented mem-cache

SeeD-Seifer:
Implemented ```shouldCache``` case for MvxAndroidLocalFileImageLoader.

Reason:

- whenever placeholder set to MvxImageView.DefaultImagePath on list items - image loading works with visible freezes

Caching local images in memory boosts list scrolling performance

### DIFF
--- a/Plugins/Cirrious/DownloadCache/Cirrious.MvvmCross.Plugins.DownloadCache.Droid/MvxAndroidLocalFileImageLoader.cs
+++ b/Plugins/Cirrious/DownloadCache/Cirrious.MvvmCross.Plugins.DownloadCache.Droid/MvxAndroidLocalFileImageLoader.cs
@@ -20,16 +20,16 @@ namespace Cirrious.MvvmCross.Plugins.DownloadCache.Droid
         : IMvxLocalFileImageLoader<Bitmap>          
     {
         private const string ResourcePrefix = "res:";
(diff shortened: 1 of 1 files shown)


roubachof:
Congrats ! Today I just run into OutOfMemoryError because DefaultImagePath res images are not cached.

softlion:
roubachof any way to contact you ?

dbeattie71:
This seemed to fix my oom issues with DefaultImage.

slodge:
Sorry - I've been meaning to try to test this to work out what this fixes and what impact it has...

On first glance, I was a bit worried about the cache just growing....

... but I suspect this was just me not thinking correctly about `WeakReference`

Thanks for the pull... sorry for the delay!

softlion:
This code does not sync the access to the dictionary, it may be an issue with mutiple cores which are in most of the devices.

SeeD-Seifer:
@softlion do you think ```ConcurrentDictionary``` should be used or a simple lock should be enough?

softlion:
ConcurrentDictionary should be much better as it won't lock unless necessary, and this will maximize the speed of this class.

dbeattie71:
I ran into the same issue as roubachof.  I had to force shouldCache=false to fix it.  Maybe I need to resize the images before they're put into the memCache, though they're pretty small.  Maybe it has something to do with InPurgable being deprecated, not sure yet.

PR #957, opened 2015-04-08
Added neccessary references to build in Visual Studio 2015

monesprit54:
Branch 3.5 cannot be built in Visual Studio 2015.

Added necessary references to project so that the solution can be built

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ AppPackages/
 *.nupkg
 *.sdf
 *.opensdf
+*.ide
 
 #NuGet
 packages/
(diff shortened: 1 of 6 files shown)


jamie94bc:
Do we know if this is going to be needed when VS2015 RTMs or is it a bug in the RC (or Xamarin)?

monesprit54:
I tried to use Visual Studio 2015 RC and it turns out that Xamarin.Android is not working fine in Visual Studio 2015 RC at the moment. That means, this could be broke down when Visual Studio 2015 is officially released.

Think that we have to wait until the final version's released to merge this PR.

martijn00:
This is a bug in Xamarin and will be solved by them soon. Bug report: https://bugzilla.xamarin.com/show_bug.cgi?id=32622

PR #960, opened 2015-04-10
Lambda is replaced by action call

evnik:
Creating of a new lambda expression object is replaced by a call of passed action with passed argument. This should decrease memory allocation while using of `MvxNotifyPropertyChangedExtensions` methods.


lothrop:
Looks good, thanks!

PR #963, opened 2015-04-18
Add missing OnSaveInstanceState override in MvxEventSource[Dialog|List]Fragment

nextmunich:
The MvxEventSource[Dialog|List]Fragment classes did contain the OnSaveInstanceState override which is present in MvxEventSourceFragment. Thus, the MvxViewModel's save state methods are not called for views which inherit MvxEventSource[Dialog|List]Fragment. This patch provides the missing override.


slodge:
Thanks :+1:

Issue #958, opened 2015-04-09
Explicitly use `[Register]` flags on all classes referencable from AXML


Cheesebaron:
As the title says, Xamarin.Android 5 series add a new requirement of explicitly using `[Register]` on all classes that can be used in AXML. So we need to do that.

http://developer.xamarin.com/releases/android/xamarin.android_5/xamarin.android_5.1/#Android_Callable_Wrapper_Naming

https://forums.xamarin.com/discussion/37277/alpha-release-xamarin-android-5-1-0-new-features-and-bug-fixes

andyci:
https://forums.xamarin.com/discussion/comment/118422/#Comment_118422 for more info

jamie94bc:
This can be closed now - was merged yesterday.

PR #965, opened 2015-04-20
Android ACW naming

martijn00:
Fixes for #958. Let me know if i missed something.

More information about the naming change: http://developer.xamarin.com/releases/android/xamarin.android_5/xamarin.android_5.1/#Android_Callable_Wrapper_Naming


slodge:
:+1: for the fixes :)

Looking at the links for #958 I'd guess that only things that go in external scripts, axml or in Java need to be set - so I'd agree the adapters aren't needed.

- http://developer.xamarin.com/releases/android/xamarin.android_5/xamarin.android_5.1/#Android_Callable_Wrapper_Naming seems to suggest that these are only needed when something accesses the Java equivalent name of a C# class.

Although (actually) given that MvvmCross does the axml inflation for views using the C# namespaces, then I'm not sure we'll hit issues with views either... but it's definitely good to have them safely non-MD5 named :) Thanks :)

kjeremy:
See comment 7 on http://forums.xamarin.com/discussion/37277/beta-release-xamarin-android-5-1-0-new-features-and-bug-fixes for clarification on how this affects axml.

kjeremy:
5.1 is now available as stable.

slodge:
woohoo

PR #971, opened 2015-04-22
Avoid loading resource when resourceId == 0 in MvxImageViewDrawableTargetBinding.

danielcweber:
In MvxImageViewDrawableTargetBinding.GetBitmap, do not attempt to load bitmap resource if resourceId == 0. Instead, set the returned bitmap to null and signal success.


slodge:
Thanks :)

PR #974, opened 2015-04-23
EnsureFolderExists support nested folders on Windows

guillaume-fr:
Adds method that tries to create parents folder recursively before creating sub folder.


slodge:
I get a bit lost with the implications of ConfigureAwait(false) here when we're forcing this all back to the calling thread anyway...

I'm not really in a position to check this win code right now... let's see if I can tease any Windows guys in to coming along, reading the code and shouting "ship it!"

jonstoneman:
Looks fine to me.  I've done some testing and blocking/exception handling is all working as expected.  I also haven't been able to create a deadlock.

slodge:
:+1: Thanks :)

Issue #954, opened 2015-04-07
RaiseCanExecuteChanged is not marshalled to UI thread automatically like RaisePropertyChanged 


FrikkinLazer:
When setting a property that a CanExecute depends on from another thread (Task.Factory.CreateNew) , the setter of the property calls RaiseCanExecuteChanged. If I dont put it inside a InvokeOnMainThread block, the button becomes unresponsive.

PR #976, opened 2015-04-23
CanExecuteChanged thread safe an raisable on UI thread

guillaume-fr:
Fixes https://github.com/MvvmCross/MvvmCross/issues/954
Events add/remove should be thread safe. They are with the default implementation. I added a lock around custom add/remove/copy for weak events in MvxCommand.


lothrop:
Thanks a lot for this!

Regarding the `lock (_eventHandlers)`: It is best practice to define a `new object()` and apply the `lock()` to that. https://msdn.microsoft.com/en-us/library/c5kehkcz.aspx

guillaume-fr:
This is intentional. I want to avoid keep memory footprint as low as possible and `_eventHandlers` is not accessible outside the class so that shouldn't be an issue.

Let me know if I missed something or if this is a requirement (mvx coding style, readability...) and I'll make the change.

lothrop:
The idea is that `List<T>` might use `lock (this)` now or at some point in the future. Looking at other code in MvvmCross, there doesn't seem to a consistent pattern for the lock object (sometimes a separate object, sometimes `this`, sometimes a different object). Is this something that should be fixed, @slodge?

slodge:
In general, I definitely prefer using a `new object()` for the `lock`. 

I agree that it won't do any harm in this case - it's a private variable, a well-known class, etc, etc - but the pattern is the pattern is the pattern and following it means no-one else can ever copy it wrong and ....

I'm also not that worried about the loss of the 8 bytes for the `lock` object (did I get that right?) - although it is possible for each item in a list of objects to each have a few commands (edit, delete, details, etc)...

How about we cover off this situation here by putting in a comment?

> there doesn't seem to a consistent pattern for the lock object (sometimes a separate object, sometimes this

Originally there was a consistent pattern - `lock (this)` was used everywhere - but then I got educated so `new object()` was the (correct) new king. This education might be somewhere here in issues actually...

IMO, the `lock(this)` code is way more dangerous than the `lock(List)` code in this request.... whoever wrote all that `lock(this)` code must have been a complete muppet :)

guillaume-fr:
I'm nitpicking but that's 4 bytes for the reference + 12 bytes for the created object in a regular x86 32bits process (don't know on iOS/Android).

Anyway you are right, I don't think memory will be an issue as the number of commands is relatively low. I remove the optimization to follow the pattern.

Issue #912, opened 2015-02-16
About the MvxWpfFileStore.cs


antasis9:
Hello developers,

Today I found some issue where 'MvvmCross/Plugins/Cirrious/File/Cirrious.MvvmCross.Plugins.File.Wpf/MvxWpfFileStore.cs'
This class has FullPath method which returns roaming folder. 

But roaming folder is shared with other computer that is signed same windows id.
So it isn't suitable save big data of applications.

So I think it may be changed to 'local' folder instead of 'roaming'.

How about your opinion?

wenyonk:
I agree with the proposed solution, or maybe an enum of special folders…? That corresponds to the enum folder path?

slodge:
I guess we could add this as a configuration option for the plugin. I think we'd happily accept a pull that did this.

I'd be unsure about changing the default setting as people are relying on it...

PR #978, opened 2015-04-24
Configurable root folder for WPF file store

guillaume-fr:
Fixes https://github.com/MvvmCross/MvvmCross/issues/912

### DIFF
--- a/Plugins/Cirrious/File/Cirrious.MvvmCross.Plugins.File.Wpf/Plugin.cs
+++ b/Plugins/Cirrious/File/Cirrious.MvvmCross.Plugins.File.Wpf/Plugin.cs
@@ -5,18 +5,40 @@
 // 
 // Project Lead - Stuart Lodge, @slodge, me@slodge.com
 
+using System;
+using System.IO;
 using Cirrious.CrossCore;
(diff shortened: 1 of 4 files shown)


Plugins/Cirrious/File/Cirrious.MvvmCross.Plugins.File.Wpf/Plugin.cs
```diff
@@ -5,18 +5,40 @@
...
-        : IMvxPlugin          
+        : IMvxConfigurablePlugin
     {
+        private string _rootFolder;
+
         public void Load()
         {
-            Mvx.RegisterType<IMvxFileStore, MvxWpfFileStore>();
-            Mvx.RegisterType<IMvxFileStoreAsync, MvxWpfFileStore>();
+            var rootFolder = _rootFolder ?? Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData);
```
slodge: Something needs to set _rootFolder at some point - it's not really a "bug", but I don't like having `_rootFolder` left dangling as `null`

guillaume-fr:
Is it ok now ?

Cheesebaron:
Looks good to me.

PR #979, opened 2015-04-24
Add initial support for a MvxPageViewController that dynamically creates...

scastria:
... UIViewController pages via additional child ViewModels.

This is my initial attempt at creating an MvxPageViewController that relies on a special interface of IMvxViewModel called IMvxPageViewModel.  MvxPageViewController asks the special interface for each child ViewModel that makes up each page of the PageViewController as needed as the user scrolls through the pages.  The child ViewModels are turned into UIViewControllers with the standard MvvmCross methods so that each UIViewController page supports data binding as an MvvmCross user would expect.  I also included a Sample Touch project to show it in action.  I didn't know what solution to add the Sample to so I chose the All solution.


yuewah:
It is great, how about having similar support for MvxTabBarViewController?

Stephanvs:
This is great! I've actually been fiddling around with something similar but this implementation beats mine :grin: 

Great idea :sparkles:

martijn00:
Doesn't this break anything?

Stephanvs:
I don't see how this would break anything, because it just introduces some new concepts, which weren't used before?

What are your concerns @martijn00 ?

martijn00:
I can't test this anytime soon, but it looks like it could work.

Stephanvs:
Still thinking about the naming though, if you consider also using a similar setup for TabBar scenarios then I'm not sure `IMvxPagedViewModel` is the best name. You wouldn't want to introduce yet another `IMvxTabbedViewModel` or something at that time...

Maybe a more general name such as `IMvxContainedViewModel` or something like that would be more appropriate?

What do you guys think?

scastria:
I don't care too much what the names are.  I have made a few tweaks since I created this pull request.  I am using my implementation in my production app and it is working just fine.  If this pull request makes it in, I can create a new one with my small tweaks.

For example, one tweak is to add a public method for:
`public virtual void NavigateToViewModel(IMvxPagedViewModel targetVM,UIPageViewControllerNavigationDirection direction, bool animated = true)`

martijn00:
@scastria This PR is anyways targeting 3.5 and would need manual merging. Could you add a new PR targeting 4.0 with your fixes included?

scastria:
Sure, I should be able to do that tonight.

scastria:
Created new PR #1121 against 4.0 branch and with my small tweaks.

PR #980, opened 2015-04-27
ability to clear the Download cache

guillaume-fr:
One method to clear all the cache, another to clear cache for a specific url.
Fix https://github.com/MvvmCross/MvvmCross/issues/499


slodge:
Thanks :)

slodge:
Thanks

I've merged - but I did wonder....

- why does ClearAll delete all the files immediately, while Clear adds them to the `_toDeleteFiles` queue? Should they both do the same thing?
- what happens if an exception gets thrown during file deletion... it can happen if something has the file open...  is the code OK against this?
- if `ClearAll` is delete files like this, should it try to delete all files (even ones not in the list?) - just in case some have "leaked"

Sorry - these are quite open questions - asking the questions is easier than working on the answers :)

PR #982, opened 2015-04-29
Avoid ArgumentOutOfRangeException when ItemSource is empty

mohibsheth:
(Background: UIPickerView has a bug where if its bound to an empty list and user tries to scroll up or down, the selected method is invoked even though the list is empty)


slodge:
Thanks...

One question - would this be better as:

           if (_itemsSource.Count() < (int)row)
           {
                  Mvx.Trace(/* some message */);
                 return;
           }

or do you think the only place where this will happen is the zero case?

slodge:
I've merged anyway (but question still out there)

Thanks :)

mohibsheth:
No thats not required. Zero case is the only place which requires this special handling. It's beyond me why iOS would even raise "Selected" for an empty list but I guess they like theatrics.

Issue #981, opened 2015-04-28
Installing AppCompat 22.1.1 causes inflation of Mvx classes to fail


kjeremy:
I just installed AppCompat 22.1.1 (up from latest 21.x) and now my MvvmCross views don't inflate from XML.  

It throws an exception when inflating the views at: https://github.com/MvvmCross/MvvmCross/blob/c306ba37afd9024f68b7a4f1fedcaf4cf7d01b8d/Cirrious/Cirrious.MvvmCross.Binding.Droid/BindingContext/MvxAndroidBindingContext.cs#L85

Now the java inflater thinks that ```MvxGridView``` is ```android.views.MvxGridView```.  If I use ```Mvx.MvxGridView``` then it tries to inflate ```Mvx.MvxGridView```.  If I use the full class name in the axml: ```Cirrious.MvvmCross.Binding.Droid.Views.MvxGridView``` then it seems to work.

kjeremy:
See https://bugzilla.xamarin.com/show_bug.cgi?id=29434

kjeremy:
Also this could be taking in the new view naming logic and might be the root cause: https://bugzilla.xamarin.com/show_bug.cgi?id=29471

kjeremy:
So I now have 22.1.1.1 and it still fails.

According to https://developer.android.com/reference/android/support/v7/app/AppCompatDelegate.html#installViewFactory%28%29 it looks like if you're using your own ```LayoutInflater.Factory``` then you shouldn't call ```installViewFactory```.  In fact if I do this in my custom activity then it doesn't crash in ```MvxAndroidBindingContext.CommonInflate```'s inflater anymore.

If that is the case then we can't use ```AppCompatActivity``` or ```AppCompatDialog``` or any other third party activity/fragment that will ultimately call ```AppCompatDelegate.InstallViewFactory``` unless someone can think of a workaround.  I'm pretty lost at this point.  Maybe @Cheesebaron, or @gshackles has an idea.

Cheesebaron:
I think we need to look at how others do it. Calligraphy does it here: https://github.com/chrisjenx/Calligraphy/blob/master/calligraphy/src/main/java/uk/co/chrisjenx/calligraphy/CalligraphyLayoutInflater.java

They had a similar issues:
https://github.com/chrisjenx/Calligraphy/issues/84
https://github.com/chrisjenx/Calligraphy/issues/92

[...]

PR #983, opened 2015-04-29
Use LayoutInflaterCompat and ILayoutInflaterFactory to install the custo...

kjeremy:
...m layout inflater

I was a little nervous about making this change since it essentially adds a dependency on the
V4 support library for every Android app that uses MvvmCross.  That said almost every
application or library in existence uses the support libraries anyway.

Alternatively we could probably use LayoutInflater.IFactory2 on devices that support it but I
couldn't quite get that dance working correctly.

Allows us to use material design with appcompat v7 22.1+.

### DIFF
--- a/Cirrious/Cirrious.MvvmCross.Binding.Droid/Binders/IMvxAndroidViewFactory.cs
+++ b/Cirrious/Cirrious.MvvmCross.Binding.Droid/Binders/IMvxAndroidViewFactory.cs
@@ -13,6 +13,6 @@ namespace Cirrious.MvvmCross.Binding.Droid.Binders
 {
     public interface IMvxAndroidViewFactory
     {
-        View CreateView(string name, Context context, IAttributeSet attrs);
+        View CreateView(View parent, string name, Context context, IAttributeSet attrs);
     }
 }
--- a/Cirrious/Cirrious.MvvmCross.Binding.Droid/Binders/IMvxLayoutInfactorFactory.cs
+++ b/Cirrious/Cirrious.MvvmCross.Binding.Droid/Binders/IMvxLayoutInfactorFactory.cs
@@ -6,13 +6,14 @@
 // Project Lead - Stuart Lodge, @slodge, me@slodge.com
 
 using System.Collections.Generic;
+using Android.Support.V4.View;
 using Android.Views;
 using Cirrious.MvvmCross.Binding.Bindings;
 
 namespace Cirrious.MvvmCross.Binding.Droid.Binders
 {
     public interface IMvxLayoutInfactorFactory
-        : LayoutInflater.IFactory
+        : ILayoutInflaterFactory
     {
         IList<KeyValuePair<object, IMvxUpdateableBinding>> CreatedBindings { get; }
     }
--- a/Cirrious/Cirrious.MvvmCross.Binding.Droid/Binders/MvxAndroidViewFactory.cs
+++ b/Cirrious/Cirrious.MvvmCross.Binding.Droid/Binders/MvxAndroidViewFactory.cs
@@ -32,7 +32,7 @@ namespace Cirrious.MvvmCross.Binding.Droid.Binders
             }
         }
 
-        public virtual View CreateView(string name, Context context, IAttributeSet attrs)
+        public virtual View CreateView(View parent, string name, Context context, IAttributeSet attrs)
         {
             // resolve the tag name to a type
             var viewType = ViewTypeResolver.Resolve(name);
--- a/Cirrious/Cirrious.MvvmCross.Binding.Droid/Binders/MvxBindingLayoutInflatorFactory.cs
+++ b/Cirrious/Cirrious.MvvmCross.Binding.Droid/Binders/MvxBindingLayoutInflatorFactory.cs
@@ -54,7 +54,7 @@ namespace Cirrious.MvvmCross.Binding.Droid.Binders
             get { return Binder.CreatedBindings; }
         }
 
-        public View OnCreateView(string name, Context context, IAttributeSet attrs)
+        public View OnCreateView(View parent, string name, Context context, IAttributeSet attrs)
         {
             if (name == "fragment")
             {
@@ -62,7 +62,7 @@ namespace Cirrious.MvvmCross.Binding.Droid.Binders
                 return null;
             }
 
-            View view = AndroidViewFactory.CreateView(name, context, attrs);
+            View view = AndroidViewFactory.CreateView(parent, name, context, attrs);
             if (view != null)
             {
                 Binder.BindView(view, context, attrs);
--- a/Cirrious/Cirrious.MvvmCross.Binding.Droid/BindingContext/MvxAndroidBindingContext.cs
+++ b/Cirrious/Cirrious.MvvmCross.Binding.Droid/BindingContext/MvxAndroidBindingContext.cs
@@ -7,6 +7,7 @@
 
 using System;
 using Android.Content;
+using Android.Support.V4.View;
 using Android.Views;
 using Cirrious.CrossCore;
 using Cirrious.MvvmCross.Binding.BindingContext;
@@ -80,7 +81,7 @@ namespace Cirrious.MvvmCross.Binding.Droid.BindingContext
                     {
                         if (factory != null)
                         {
-                            clone.Factory = factory;
+                            LayoutInflaterCompat.SetFactory(clone, factory);
                         }
                         var toReturn = clone.Inflate(resourceId, viewGroup, attachToRoot);
                         if (factory != null)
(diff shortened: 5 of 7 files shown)


kjeremy:
Fixes #981 

I've tested this locally and it prevents the crash and gives us AppCompat tinting on activities/fragments that support it.

softlion:
+1 !

martijn00:
I like it!

jamie94bc:
:+1: 

But do we need [this](https://docs.nuget.org/consume/package-restore/migrating-to-automatic-package-restore)? :wink: 
image

kjeremy:
Good point on that nuget thing.  I wonder how that got in there.

kjeremy:
Actually @jamie94bc I think that it's correct since it's consistent with the other csproj files.  We should probably migrate to Automatic Package Restore eventually but that's outside the scope of this pull.

jamie94bc:
@kjeremy makes sense. I hate the MSBuild based package restore - definitely should move towards Automatic Package Restore (ideally before VS 2015 is released)!

slodge:
My only question is.....


This puts a dependency on all of MvvmCross Droid that it **has** to use the Android support package.


Is this a dependency that we want to force on all Android apps? If so, why?

martijn00:
I think as @kjeremy says that a lot of users use the android support packages already, and because Google has made clear that they will focus and maintain those packages, i think it is reasonable to include a dependency on one of them.

kjeremy:
@slodge Yeah that's my only reservation but most apps require support for v4 anyway.

There may be a way to workaround it with ```LayoutInflater.IFactory2``` but then we need to guard the implementation for API < 11 and for some reason I still got the layout crash when pursuing that method.  Chris Banes said to just use LayoutInflaterCompat.

slodge:
Y - totally understand about "a lot" and "most" - but what about the few...

Thinking.... trying to keep the footprint small... but it's never been easy....

kjeremy:
Alternatively something like this might work (basically a port of AppCompat's implementation crammed into one file)

```csharp
    public static class MvxLayoutInfactorFactory
    {
        private static readonly int SdkInt = (int)Android.OS.Build.VERSION.SdkInt;

        public interface ILayoutInflaterCompatImpl
        {
            void SetFactory(LayoutInflater layoutInflater, IMvxLayoutInfactorFactory factory);
        }

        private class LayoutInflaterCompatImpl : ILayoutInflaterCompatImpl
        {
            public void SetFactory(LayoutInflater layoutInflater, IMvxLayoutInfactorFactory factory)
            {
                LayoutInflaterCompatBase.SetFactory(layoutInflater, factory);    
            }
        }

        private class LayoutInflaterCompatImplV11 : ILayoutInflaterCompatImpl
        {
            public void SetFactory(LayoutInflater layoutInflater, IMvxLayoutInfactorFactory factory)
            {
                LayoutInflaterCompatHoneyComb.SetFactory(layoutInflater, factory);
            }
        }

        private static class LayoutInflaterCompatBase
        {
            internal class FactoryWrapper : Java.Lang.Object, LayoutInflater.IFactory
            {
                protected readonly IMvxLayoutInfactorFactory DelegateFactory;

                public FactoryWrapper(IMvxLayoutInfactorFactory delegateFactory)
                {
                    this.DelegateFactory = delegateFactory;
                }

                public View OnCreateView(string name, Context context, IAttributeSet attrs)
                {
                    return this.DelegateFactory.OnCreateView(null, name, context, attrs);
                }
            }

            public static void SetFactory(LayoutInflater layoutInflater, IMvxLayoutInfactorFactory factory)
            {
                layoutInflater.Factory = (factory != null ? new FactoryWrapper(factory) : null);
            }
        }

        private static class LayoutInflaterCompatHoneyComb
        {
            private class FactoryWrapper2 : LayoutInflaterCompatBase.FactoryWrapper, LayoutInflater.IFactory2
            {
                public FactoryWrapper2(IMvxLayoutInfactorFactory delegateFactory)
                    : base(delegateFactory)
                {}

                public View OnCreateView(View parent, string name, Context context, IAttributeSet attrs)
                {
                    return DelegateFactory.OnCreateView(parent, name, context, attrs);
                }
            }

            public static void SetFactory(LayoutInflater layoutInflater, IMvxLayoutInfactorFactory factory)
            {
                layoutInflater.Factory2 = (factory != null ? new FactoryWrapper2(factory) : null);
            }
        }

        private static ILayoutInflaterCompatImpl Impl
        {
            get
            {
                if (SdkInt >= 11)
                    return new LayoutInflaterCompatImplV11();
                else
                    return new LayoutInflaterCompatImpl();
            }
        }

        public static void SetFactory(LayoutInflater layoutInflater, IMvxLayoutInfactorFactory factory)
        {
            Impl.SetFactory(layoutInflater, factory);
        }
    }
```

And we set the factory like so ```csharp MvxLayoutInfactorFactory.SetFactory(layoutInflator, factory);```

slodge:
(looks up from trying to work out why Xam's update refuses to accept some of my Android SDKs....)

ooooh.... interesting....

If that'd buy us some time to make a proper decision on support library that'd be fab :+1: 

The realist in me suspects you're right - that 99.9% of new apps need Android support to get things done... so maybe one day soon we should try a major release increment that pulls Android forwards...  with support on board, that increment could even drop old Android (3.x could do?) and could merge fragments back into the core... it would definitely be a breaking release though...

(back to those Android SDKs....)

slodge:
(hmmm.... wonder where the old mono maps stuff has gone... I guess it's not used much any more...)

kjeremy:
I think I might have a slightly simpler implementation...

With Xamarin 5.1 they've dropped compiler support for API <= 8 (I think that's the version).  They dropped the google maps assembly and now recommend google play services nuget package or component: https://developer.xamarin.com/releases/android/xamarin.android_5/xamarin.android_5.1/

kjeremy:
As far as I can tell the following SHOULD work... but it doesn't.  It's basically doing the same thing as ```LayoutInflaterCompat```. I can verify that our factory2 is being set but ```MvxListView``` still won't inflate.  Can anyone spot the error?

```csharp
    public static class MvxLayoutInfactorFactory
    {
        private static readonly int SdkInt = (int)Android.OS.Build.VERSION.SdkInt;

        private class FactoryWrapper : Java.Lang.Object, LayoutInflater.IFactory
        {
            protected readonly IMvxLayoutInfactorFactory DelegateFactory;

            public FactoryWrapper(IMvxLayoutInfactorFactory delegateFactory)
            {
                this.DelegateFactory = delegateFactory;
            }

            public View OnCreateView(string name, Context context, IAttributeSet attrs)
            {
                return this.DelegateFactory.OnCreateView(null, name, context, attrs);
            }
        }

        private class FactoryWrapper2 : FactoryWrapper, LayoutInflater.IFactory2
        {
            public FactoryWrapper2(IMvxLayoutInfactorFactory delegateFactory)
                : base(delegateFactory)
            {}

            public View OnCreateView(View parent, string name, Context context, IAttributeSet attrs)
            {
                Mvx.Warning("FactoryWrapper2.OnCreateView {0}", name);
                return DelegateFactory.OnCreateView(parent, name, context, attrs);
            }
        }

        public static void SetFactory(LayoutInflater layoutInflater, IMvxLayoutInfactorFactory factory)
        {
            if (SdkInt >= 11)
            {
                Mvx.Warning("Setting FactoryWrapper2");
                layoutInflater.Factory2 = (factory != null ? new FactoryWrapper2(factory) : null);
            }
            else
            {
                layoutInflater.Factory = (factory != null ? new FactoryWrapper(factory) : null);
            }
        }
```

jamie94bc:
@kjeremy how are you referencing `MvxListView` in axml? I noticed that view namespace abbreviations didn't seem to play nice in Xamarin.Android 5.1 yesterday but didn't have a chance to look into it properly.

If you're using `<Mvx.MvxListView />` it might be worth trying `<cirrious.mvvmcross.binding.droid.views.MvxListView />` instead.

kjeremy:
I'm saying "MvxListView" in the axml. It works in a regular activity.  The
moment I switch to an ActionbarActivity it fails to inflate.

jamie94bc:
`ActionBarActivity` is V22.0 - don't you want `AppCompatActivity` (V22.1)?

kjeremy:
It's the same thing
https://github.com/android/platform_frameworks_support/blob/7cc23e31790b02f31707b36655401a69b8e6ebc2/v7/appcompat/src/android/support/v7/app/ActionBarActivity.java

kjeremy:
To be clear this happens with both Xamarin 4 and 5.1.  It is strictly an AppCompat v22.1 problem.

kjeremy:
I'm going to close this since #988 is now merged.

PR #985, opened 2015-04-30
Migrate to Automatic Package Restore

kjeremy:
It's the wave of the future.  Packages are restored during build and do not rely on msbuild.

See: https://docs.nuget.org/consume/package-restore/migrating-to-automatic-package-restore


slodge:
Thanks :)

Gosh - the future is shiny!

jamie94bc:
I think this can be removed too? https://github.com/MvvmCross/MvvmCross/tree/3.5/.nuget New issue?

kjeremy:
Yeah I think so... can't remember if I left it on purpose or just missed
it.  I'd make a new issue or better yet a PR.

PR #988, opened 2015-05-01
Reimplement LayoutInflaterCompat in MvvmCross.

kjeremy:
Works around a platform bug with API 11 - 21 setting the layout factory.
Allows us to use material design with appcompat v7 22.1+.

Alternate fix for #981.  Does not rely on Support v4 unlike #983.


slodge:
Reviewed the code...

....

...

:peach: 

Will merge and test :)

:+1:

PR #991, opened 2015-05-03
Null exception when not creating attatchments

The-Bytemaster:
An exception is now thrown with 3.5.1 when calling the ComposeEmail method without any attachments. This simply adds a null check before it tries to call the ToList extension method on attachments.


Cheesebaron:
Really indecisive with those commits, aren't you?

The-Bytemaster:
Embarrassing, I know.  After years of using other systems, this was actually my first full experience with GIT or a pull request.  I am amazed myself it has taken me this many years to do so, but you have to start sometime.  I did not realize the was not a clean way to back out a commit... live and learn.

martijn00:
Thanks! This is fixed through pull request #1019

Issue #714, opened 2014-06-08
Async File Plugin for WP and WinRT


slodge:
Left over work from https://github.com/MvvmCross/MvvmCross/pull/677 - still need implemenations on other platfoms

Kavignon:
Has someone made some advancements on this issue ?

PR #993, opened 2015-05-04
FileStore is async with CancellationToken support on all platforms

jonstoneman:
Implement async file plugin for Windows Phone, fixes https://github.com/MvvmCross/MvvmCross/issues/714


Cheesebaron:
Looks good, should we consider adding overloads with `CancellationToken` as well?

guillaume-fr:
If so, should we pass the `CancellationToken` to the read/write delegate ? I'm not a fond of delegates in this context and I would have returned the stream...

jonstoneman:
Yes, good idea @Cheesebaron.  I'll update the PR to include `CancellationToken` support on all platforms.

@guillaume-fr: Why wouldn't we pass the `CancellationToken` to to read/write delegate?

jonstoneman:
Included https://github.com/MvvmCross/MvvmCross/pull/975 and added `CancellationToken` support on all platforms.  In order to minimise code duplication, there is a new MvxFileStoreBase, which is used by all implementations, and MvxBaseFileStore in HackFileShare, which is used by the iOS/Android/WPF implementations has beome MvxIoFileStoreBase.

jonstoneman:
Oh, this is **untested!**  I will do so at some point but would be grateful if anyone has any comments or can test in the meantime.

jonstoneman:
This has now been very well tested on WPF, WindowsPhone, WindowsStore & WindowsCommon (store + phone) and a couple of fixes have been applied.

iOS and Android yet to be tested but they should be fine as they share almost all their code with WPF.

jonstoneman:
Also tested on Android.  iOS may be a while... my mac is a bit unwell.  :(

jonstoneman:
Now tested on iOS too - no problems found.  Would be nice if someone else could have a look through the code but otherwise this should be good to merge now...

martijn00:
Looks good to me. Let's merge this in.

jonstoneman:
Thanks :)

PR #996, opened 2015-05-05
Add override to determine if the requested fragment

### LINKED ISSUE #966: MvxCachingFragmentActivity.ShowFragment loading the same cached fragment even with a different bundle
Hello,

**Problem :** 

I first load a fragment X within a container Y and a navigation parameter (id=1). Then I want to load the same fragment Y within the same container Y but with another navigation parameter (id=2). This request loads my cached fragment with data from parameter id=1.

**Cause :** 

The `MvxCachingFragmentActivity.ShowFragment()` method tests if a fragment is already loaded, based on the request parameters tag + contentId. It never tests if the bundle is the same or not.

**Solution (proposals) :** 

1. Provide a way to override `IsContentIdCurrentyShowingFragmentWithTag` and inject some conditions. 
  * Set `IsContentIdCurrentyShowingFragmentWithTag` to protected virtual => low impact
2. Provide a way to register a `Fragment` with a parameter to bypass the test and instantiate a new `Fragment` ignoring the cache. 
  * Modify the `FragmentInfo` and `RegisterFragment` signature => medium impact
3. Store and test if the `Bundle` is different from the one in the current `Fragment`
  * Modify the `_currentFragments` to store the `bundle` => medium impact

What do you think ?

andyci:
is currently showing so that you can force the fragment to be reloaded in situations where the same fragment is requested again but with different parameters

resolve #966

### DIFF
--- a/Cirrious/Cirrious.MvvmCross.Droid.Fragging/MvxCachingFragmentActivity.cs
+++ b/Cirrious/Cirrious.MvvmCross.Droid.Fragging/MvxCachingFragmentActivity.cs
@@ -196,8 +196,9 @@ namespace Cirrious.MvvmCross.Droid.Fragging
             string currentFragment;
             _currentFragments.TryGetValue(contentId, out currentFragment);
 
-            // Only do something if we are not currently showing the tag at the contentId
-            if (IsFragmentCurrentlyShowing(contentId, tag)) return;
+            var shouldReplaceCurrentFragment = ShouldReplaceCurrentFragment(contentId, tag);
+            if (!shouldReplaceCurrentFragment)
+                return;
 
             var ft = SupportFragmentManager.BeginTransaction();
             OnBeforeFragmentChanging(tag, ft);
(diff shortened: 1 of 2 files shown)


andyci:
Note: I renamed the function as there was a typo and I didn't think it was immediately obvious what it does which is important if we make it part of the public API

Cheesebaron:
Needs to be added to FullFragging as well.

andyci:
Thanks

andyci:
I've had another look at this and realised that overriding IsFragmentCurrentlyShowing and returning false won't be enough as the cached fragment will still be used.  Also, I don't think the override is that useful as it was because the user will need to check the _currentFragments dictionary to check what the current fragment actually is.

If this is ok then I will apply to FullFragging as well

slodge:
This looks good to me.... :+1:

But I haven't actually run it... wondering what use case I'd have to add do to https://github.com/MvvmCross/MvvmCross-Tutorials/tree/master/Fragments to make this make sense to me or to others wanting to use it.

kjeremy:
Shouldn't this also key off of `Fragment.IsVisible`?

andyci:
What do you mean?

Issue #821, opened 2014-11-03
Handling async work with MvxCommand


molinch:
This is related to the old task #374 

The idea is to handle asynchronous code with MvxCommand and not with a new class.
There would be a new constructor and some extra handling for Tasks. Then .NET would automatically resolve the correct constructor, ie: the one accepting a Task.

1st benefit: same class for all usages, sync or non async
2nd benefit: we can await the given Task and have a correct CanExecute implementation

Do you think this would make sense?
If yes I will do it and make a PR.

guillaume-fr:
Tried something : https://github.com/MvvmCross/MvvmCross/compare/3.5...guillaume-fr:3.5-asynccommand

How should we handle exceptions ? Last exception in a property, an event or maybe leave it unhandled ?
Should it implement INPC if it has properties like LastError, IsRunning... ?
Use ConfigureAwait(false), when https://github.com/MvvmCross/MvvmCross/issues/954 is fixed ?

lothrop:
Very nice. About the exceptions: I'm wondering if it would be a good idea to expose the `Task`(s) through a getter. That way the caller could retrieve the `AggregateExceptions` from there. Or `await` that `Task` and use try/catch.

guillaume-fr:
We may consume too much memory if we keep a list of completed tasks. I preffer the event over all other solutions (including the one I implemented ;) ).

I'm starting to experiment Cancellation, I'll commit soon.

dbeattie71:
Mytoolkit has asyncommand, might look and see how it handles exceptions.

guillaume-fr:
@dbeattie71 it doesn't handle exceptions https://mytoolkit.codeplex.com/SourceControl/latest#MyToolkit/Command/AsyncRelayCommand.cs

dbeattie71:
Ah, nm, I was thinking of the runasync in the vm base.

PR #997, opened 2015-05-06
Handling async work with MvxCommand

guillaume-fr:
Fixes https://github.com/MvvmCross/MvvmCross/issues/821


guillaume-fr:
@lothrop, finally I feel like a PR is better to discuss code.
So, maybe the `ExecuteAsync` method should rethrow the exception so the caller can catch it as usual ?
Then what about the fire and forget version that will be used in binding `Execute` ?

Non exhaustive list of possible solutions :
 - Keep the code as it is, if the caller want to catch an error, he can subscribe to the event. This is consistent between Async / Fire and forget.
 - Use the event only in Fire and forget mode, let the exception bubble up when calling async method
 - Remove the event, when in fire and forget we cannot observe the result (after all, that's how regular async void and events works) and let the exception bubble up when calling the async method.

Maybe the last one is less confusing as we don't introduce a new concept and the user is responsible for exception handling in his code. We should keep the `Mvx.Error` for debbuging. *Do we agree on removing this event and let the regular exception handling do its job ?*

Cirrious/Cirrious.MvvmCross/ViewModels/MvxAsyncCommand.cs
```diff
@@ -0,0 +1,283 @@
...
+            else
+                return DoCanExecute(parameter);
+        }
+
+        public async void Execute(object parameter)
+        {
+            await ExecuteAsync(parameter).ConfigureAwait(false);
+        }
+
+        public async void Execute()
```
evnik: Exceptions thrown in `async void` methods couldn't be catched. While `DoExecute` calls are wrapped in `try/catch` blocks, `DoCanExecute` are not, so any exception thrown there might crash the application. So, the question is, do we really need synchronous `Execute` methods in a class, which made to be asyncronous? If yes, I believe, it will be better to wrap `DoCanExecute` calls in a `try/catch` block too.

guillaume-fr: I think we need the synchronous (fire and forget) version so it can be bound.
That being said, you are right : if we want custom exception handling on async command we should also catch exceptions from `CanExecute` and chose what to do with it.
But that's another argument against the custom exception handling I implemented. Moreover, nothing special was done for synchronous command. If an exception is raised by your CanExecute (or Execute) while it's used from binding, I think it'll crash the application and that's expected.

evnik: I got that we need synchronous `Execute` methods, while we need `ICommand` interface to be implemented.
Related to synchronous command, as I know, there are no any async code. This means that only library users hold responsibility on handling of exceptions thrown by their code. Introducing `async void` methods, the library takes this responsibility, because users cannot catch that exceptions anymore.
So, even if we can neither rethrow exceptions asyncronously nor return them through an event, I believe all exceptions (including those thrown in `CanExecute` method) should be catched and at least traced as in current implementation.

jamie94bc:
This looks great - should certainly solve a few issues we've encountered.

For consistency `DoCanExecute` and `DoExecute` abstract methods should be named `CanExecuteImpl` and `ExecuteImpl`?

jamie94bc:
Anything preventing this being merged in now?

lothrop:
Great work. Thanks!

PR #998, opened 2015-05-07
Added a ctor to Setup that takes a presenter.

jfoshee:
This helps with the use of storyboards. Also normalized whitespace in this file.


slodge:
Thanks :)

PR #1003, opened 2015-05-11
Load local images asynchronously and scale them down (android)

gentledepp:
Load local images asynchronously and scale them down (android)

== Problem ==
Loading local images on the UI Thread on Android is a serious bottleneck.
In our production application we've got a ListView that displays all images a user has taken with the devices camera. These can be up to 8 or even more megapixels.
As loading is done on the UI thread the UI freezes as the user scrolls through the ListView.

== This change brings two benefits: ==
a) It changes (sorry for that) the interface of the IMvxLocalFileImageLoader<T> to be asynchronous as the others. The LocalImageLoaders of iOS and Android kick off a task (Task.Run()) to load the actual images.
This already improves scolling performance.

b) As Android supports loading a scaled down version of a Bitmap, the IMvxImageHelper<T> got two properties "MaxWidth" and "MaxHeight"
These are set by the MvxImageView, if its "maxWidth" and "maxHeight" properties are set.
So given you specify a maxWidth- and maxHeight on MvxImageView, it loads the scaled-down version.
This speeds up loading but brings a tremendous memory benefit! 

== Note ==

a) Please note that I do not have a license for Xamarin.IOS - so I made the iOS Plugin change in Notepad ++. (I triple checked it, but I'll take the blame :-))

b) I also have to admit that I have very little knowledge of Xamarin.iOS license. I tried to find an approach to load a scaled-down UIImage but could not find anything. Maybe I am blind. That said - if anybody could point me to the right direction or just implement this feature in my branch, that would be pretty awesome.

### DIFF
--- a/Cirrious/Cirrious.MvvmCross.Binding.Droid/Views/MvxImageView.cs
+++ b/Cirrious/Cirrious.MvvmCross.Binding.Droid/Views/MvxImageView.cs
@@ -22,21 +22,11 @@ namespace Cirrious.MvvmCross.Binding.Droid.Views
     public class MvxImageView
         : ImageView
     {
-        private readonly IMvxImageHelper<Bitmap> _imageHelper;
+        private IMvxImageHelper<Bitmap> _imageHelper;
 
         public MvxImageView(Context context, IAttributeSet attrs)
             : base(context, attrs)
         {
-            if (!Mvx.TryResolve(out _imageHelper))
-            {
-                MvxTrace.Error(
-                    "No IMvxImageHelper registered - you must provide an image helper before you can use a MvxImageView");
-            }
-            else
-            {
-                _imageHelper.ImageChanged += ImageHelperOnImageChanged;
-            }
-
             var typedArray = context.ObtainStyledAttributes(attrs,
                                                             MvxAndroidBindingResource.Instance
                                                                                      .ImageViewStylableGroupId);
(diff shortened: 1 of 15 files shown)


Cheesebaron:
Well there are a couple of problems just firing `Task.Run` and forgetting about it.

1. You can't cancel the `Task`
2. When navigating back or forward it will continue using resources until it is done

I guess it is an acceptable compromise, but I would much rather prefer that the entire thing was made async anyways.

gentledepp:
So just to be sure:
You are proposing that I should convert the whold api to an async one, right?

This was also on my mind at first, but the incompatibility to the Microsoft.BCL nuget packages kept me from doing that.
I cannot take the decision to drop platform support of MvvmCross just because I would like to use "await async" in one of the Plugins - can you?
Maybe we should ask Stuart first, but I guess that this is going to be a showstopper

Cheesebaron:
> incompatibility to the Microsoft.BCL nuget packages

What is incompatible?

gentledepp:
I re-tried adding the BCL Nuget package and it worked. Seems I had chosen a wrong one. Sorry for this misunderstanding.
Now, I changed the DownloadCache to load the images asynchronously as you suggested.
I also created another branch where I (radically) changed all interfaces in the "DownloadCache" plugin to use async await instead of callback actions.
I will make a pull request in a sec. 
This pull request is completely comaptible and not such a huge change => It is more unlikely that I break something with it ;-)

evnik:
I believe it is not good to call `Task.Run` without check if the method is already called in a background thread. Such behavior may produce waiting of free thread in the pool and blocking of another thread working.

PR #1005, opened 2015-05-14
3.5 downloadcache all async

gentledepp:
As cheesebaron argued, it would be much nicer if the whole API of the DownloadCache plugin used "Task<T>" and async await instead of the callback actions in the interfaces.

So I just did it. I hope you like what you see ;-)


guillaume-fr:
Task returning methods should have an “Async” suffix.

gentledepp:
Ok... if that makes you happy.
But if I continue working on that code, will it eventually be merged in? 
Or are there any other issues? (I'd prefer not to continue working on this branch bit by bit for several weeks... :))

PR #1010, opened 2015-05-19
Fixed Email plugin crash on Droid when attachments are not specified

SeeD-Seifer:


SeeD-Seifer:
Though the change-set looks big, there is only added additional check:
```
if (attachments != null) {
   ...
}
```

martijn00:
Thanks! This is fixed through pull request #1019

PR #1013, opened 2015-05-21
Build in MvxBindingAttributes.xml + TextFormatted binding

Cheesebaron:


Cheesebaron:
To test TextFormatted binding. Create a string property in a ViewModel populate it with a string such as: `"<b>Bold Lolz</b> <i>italic lolz</i>"`

Create a ValueConverter:

```
public class TextToSpannableValueConverter : MvxValueConverter<string, ISpanned>
{
    protected override ISpanned Convert(string value, Type targetType, object parameter,
        CultureInfo culture)
    {
        return Html.FromHtml(value);
    }

    protected override string ConvertBack(ISpanned value, Type targetType, object parameter,
        CultureInfo culture)
    {
        return Html.ToHtml(value);
    }
}
```

Bind it to a `TextView`:

    local:MvxBind="TextFormatted MyString, Converter=TextToSpannable"

kjeremy:
Looks good to me though I would split out the ```MvxBindingAttributes.xml``` move into another PR.

Cheesebaron:
So you want me to create another branch... This is what cherry picking is for...

kjeremy:
It was just a suggestion so we can track it easier... I don't care strongly enough about it either way.  Both these changes are good IMHO. :+1:

Cheesebaron:
Well you can't make two separate PR on the same branch. Commit history stays when merging, so I don't understand your "track it easier" claim when they are in separate PR's.

kjeremy:
It's not a big deal.  I think this is good to merge.

PR #1017, opened 2015-06-02
Fixed queuing unindexed files

mierzynskim:
Every time when <code>QueueUnindexedFilesForDelete</code> method was called, every file was queued for deletion. It was caused by not using absolute path in <code>DownloadPath</code>.


lothrop:
Thanks!

PR #1026, opened 2015-06-14
Added WCF DataContract attribute to allow MvxNotifyPropertyChanged-de…

cabadam:
I was using this in an application I am developing.  App consists of a WCF service backend which a variety of clients can communicate with (WPF, Android, etc).

For simplicity, developed a PCL base-library that contained the necessary data objects, view models, etc.  Those data objects, derived from MvxNotifyPropertyChanged to make data binding easier, were throwing errors when the WCF layer attempted to serialize them.  All base types needed to have the [DataContract] attribute on them for the serializer to be able to handle them.

### DIFF
--- a/Cirrious/Cirrious.MvvmCross/ViewModels/MvxNotifyPropertyChanged.cs
+++ b/Cirrious/Cirrious.MvvmCross/ViewModels/MvxNotifyPropertyChanged.cs
@@ -9,10 +9,12 @@ using System;
 using System.ComponentModel;
 using System.Linq.Expressions;
 using System.Runtime.CompilerServices;
+using System.Runtime.Serialization;
 using Cirrious.CrossCore.Core;
 
 namespace Cirrious.MvvmCross.ViewModels
 {
+    [DataContract]
     public abstract class MvxNotifyPropertyChanged
         : MvxMainThreadDispatchingObject
         , IMvxNotifyPropertyChanged
--- a/CrossCore/Cirrious.CrossCore/Core/MvxMainThreadDispatchingObject.cs
+++ b/CrossCore/Cirrious.CrossCore/Core/MvxMainThreadDispatchingObject.cs
@@ -6,9 +6,11 @@
 // Project Lead - Stuart Lodge, @slodge, me@slodge.com
 
 using System;
+using System.Runtime.Serialization;
 
 namespace Cirrious.CrossCore.Core
 {
+    [DataContract]
     public abstract class MvxMainThreadDispatchingObject
     {
         protected IMvxMainThreadDispatcher Dispatcher


kjeremy:
This bothers me for some reason.  I don't like the idea of making a change just to accomodate WCF.  Why doesn't `MvxMainThreadDispatchingObject` need this attribute?

cabadam:
Jeremy,
The MvxMainThreadDispatchingObject type DOES need it (and was included in the pull request).  All classes in the type heirarchy have to have the attribute to be usable for WCF in this manner.

I agree it would be nice to not have to add this attribute, but it also defeats the ability to reuse a bunch of code if we have to redefine duplicate data objects for the sole purpose of passing them in service calls.  Then you have to add conversion methods, etc, etc.  A lot more places for something to get missed in the future.

kjeremy:
Oops, I missed the second part :)

Cheesebaron:
WCF, really? Are people using that stuff?

promontis:
lol @Cheesebaron just my reaction

kjeremy:
It sounds to me like you may need to pack/unpack your model objects into other objects that handle serialization.  Surely they can't expect you to use `[DataContract]` all the way down the chain.

kjeremy:
@cabadam 

How about one of these surrogates?

https://msdn.microsoft.com/en-us/library/ms733064.aspx

cabadam:
@kjeremy - Something like that is what I was alluding to in the last paragraph of my previous comment.  If I follow that, and I want to add a new property in the future, now I have to go add it in two different places.  It just seems like a good way to guarantee problems down the road.  Given that you already have an object that represents a particular object/model, why would I want to create another one with basically the exact same properties?

@Cheesebaron - Not sure I understand your reaction.  Yes, people use WCF.  Perhaps you don't, but others do.

Cheesebaron:
@cabadam my reaction is as it is, because you are using ancient technology and expect backwards compatibility. Welcome to 2007!

cabadam:
Sigh.  I'm not trying to get into an argument of what technology should be used, is too old.

I am using it.  I made a change to make it compatible.  I thought others might find it useful.  That's it.

kjeremy:
@Cheesebaron 

There could be many reasons behind @cabadam having to use WCF and there's a good chance they are beyond his control.

That said I don't think this is the right place for this change.  @cabadam maybe you can use that solution with generics to avoid a lot of duplcated code or possibly fill some sort of transport object instead of using your UI-facing class directly.

Maybe there's a way to make a WCF MvvmCross plugin?  I'm not sure.

Cheesebaron:
@kjeremy sure there could be many reasons, none of the very good anymore I would suspect.

Anyways, I strongly agree that annotating those two core classes with `[DataContract]` is a no no. A `ViewModel` should not be serialized and used to communicate with a server, that is the `Model`s responsibility in a traditional MVVM use case.

kjeremy:
@Cheesebaron if it's a customer requirement it's usually a good one :(

cabadam:
@Cheesebaron -  You are absolutely correct - it is the model's responsibility.  Just to clarify - these are my Model objects, not my view models.  Each of my model types derives from MvxNotifyPropertyChanged.  It simplifies a lot of things.  The same change notifications are very useful even on the service back end to respond to changes.  It isn't just a UI thing.  UI is just the most common use.

PR #1031, opened 2015-06-16
Renaming "GroupItemTemplate" android attribute to "MvxGroupItemTemplate"

ljeancler:
"GroupItemTemplate" android attribute has been renamed to "MvxGroupItemTemplate" to match the current naming convention that uses "Mvx" prefix for android attribute.


martijn00:
You address an issue with tabHostView view here too. Please separate this in another PR.

ljeancler:
Sorry about that, I'm new to git. I've reverted the last commit about the tabHostView (testing purpose). Is it ok ?

martijn00:
Thanks! That last commit seemed valid, but should be in another PR to have a separation of concerns. Can you make a new PR for it?

PR #1043, opened 2015-06-26
Add javaReference/transfer constructors to fragments

andyci:
Android sometimes crashes if it cannot find this constructor. Note that ideally we need to warn people to add this constructor to their own fragments as well although I'm not sure what causes Xamarin.Android to call this constructor


martijn00:
This alsi needs to be added to full fragging and the android support fragments

andyci:
Hopefully that's all of them?

martijn00:
Thanks! This is everything for the main repo, but you need to open another PR here: https://github.com/MvvmCross/MvvmCross-AndroidSupport

We are planning to move everything over to that one.

PR #1048, opened 2015-07-01
Prevent view id collision when generating RadioButton ids

kjeremy:
Port of View.generateViewId  (API >= 17).  I've never seen a collision in practice but you never know.


martijn00:
Not sure if this one is needed, but lets ship it!

Issue #1011, opened 2015-05-20
Move support fragments library (Cirrious.MvvmCross.Droid.Fragging) to MvvmCross-AndroidSupport


jamie94bc:
Since we decided components dependent on the Android support library should be moved to a separate repository, shouldn't we do the same with fragments?

Cheesebaron:
I would say yes. I would also say that we move forward and push everything that has to do with `Fragment` is strictly dependent on support packages, meaning no "FullFragging". This will streamline behavior and possibilities between platforms.

kjeremy:
Should we align these with the Android support packages like ```MvvmCross-AndroidSupport.V7```, ```MvvmCross.AndroidSupport.RecyclerView``` (etc.) or something like that?

jamie94bc:
I think `Cirrious.MvvmCross.Droid.Support.RecyclerView` would be better naming (considering we may want to support API 21 recycler view)?

I don't think we should me mimicking the support libraries namespaces exactly though :-1:

martijn00:
Thanks, this is done now.

PR #1049, opened 2015-07-02
Removed Fragging and FullFragging

martijn00:


martijn00:
Fix for: https://github.com/MvvmCross/MvvmCross/issues/1011

martijn00:
@slodge What do you think of this? Should we move forward and remove this stuff? Everything should be compatible with https://github.com/MvvmCross/MvvmCross-AndroidSupport we only need to warn people that they have to fix the packages to the new nugets

martijn00:
@JoeCooper Yes it is now in: https://github.com/MvvmCross/MvvmCross-AndroidSupport

We will make some explanations on this later.

lothrop:
@martijn00 FullFragging is still there... Is that intentional?

martijn00:
@lothrop yes. due to requests of users of full fragging we added it again.

PR #1060, opened 2015-07-13
Fix download cache

Cheesebaron:
Someone broke it. This seems to fix it.


martijn00:
Current one is broken so let's merge it.

PR #1071, opened 2015-07-17
Added back FullFragging

martijn00:


Cirrious/Cirrious.MvvmCross.Droid.FullFragging/Fragments/MvxDialogFragment.cs
```diff
@@ -0,0 +1,69 @@
...
+        protected MvxDialogFragment(IntPtr javaReference, JniHandleOwnership transfer)
+            : base(javaReference, transfer)
+        {
+            this.AddEventListeners();
+        }
+
+        protected MvxDialogFragment()
+        {
+            this.AddEventListeners();
+        }
```
kjeremy: Doesn't this mean that `AddEventListeners()` could happen twice?

See http://developer.xamarin.com/guides/android/under_the_hood/architecture/#Java_Activation

PR #1075, opened 2015-07-21
Accidently deleting all files in DownloadCache

MKuckert:
I just came across an issue in my touch app. The DownloadCache plugin is removing all files from the file cache during the app start, rendering the cache totally useless in an offline situation. The pull request adds a commit, transforming an absolute file path into a relative file path. After that it is comparable to the stored DownloadedPath and the file cache is not fully wiped.

I've tested on iOS only. The cache is not cleaned on app restart after applying the patch.


martijn00:
Can you open this against the https://github.com/MvvmCross/MvvmCross-Plugins repo?

martijn00:
Closing this for now. Please open again on the plugins repo.

PR #1077, opened 2015-07-22
Fix back stack for full fragging

martijn00:
Also fixed for support: https://github.com/MvvmCross/MvvmCross-AndroidSupport/pull/51


martijn00:
Fixed in https://github.com/MvvmCross/MvvmCross/pull/1079

PR #1087, opened 2015-08-07
Fixed GetSizeInBytes() to proper handle memory cache limits for the Images

IlSocio:
Computation for the allocated bytes was overestimated.

### DIFF
--- a/Plugins/Cirrious/DownloadCache/Cirrious.MvvmCross.Plugins.DownloadCache.Droid/Cirrious.MvvmCross.Plugins.DownloadCache.Droid.csproj
+++ b/Plugins/Cirrious/DownloadCache/Cirrious.MvvmCross.Plugins.DownloadCache.Droid/Cirrious.MvvmCross.Plugins.DownloadCache.Droid.csproj
@@ -11,7 +11,8 @@
     <AssemblyName>Cirrious.MvvmCross.Plugins.DownloadCache.Droid</AssemblyName>
     <FileAlignment>512</FileAlignment>
     <GenerateSerializationAssemblies>Off</GenerateSerializationAssemblies>
-    <TargetFrameworkVersion>v2.3</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.0.3</TargetFrameworkVersion>
+    <AndroidUseLatestPlatformSdk>False</AndroidUseLatestPlatformSdk>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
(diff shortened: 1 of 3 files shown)


jamie94bc:
LGTM :+1:

Cheesebaron:
`BitmapCompat` should be OK to use. Is Support v4 already referenced in DownloadCache?

IlSocio:
I just added the Support v4 package... but I also had to change the compilation target from API 10 to API 15 because the Support v4 package contains only binaries for the API 15 (v4.0.3)

I also noticed that some other project (eg. CossUI.Droid) is already targeting the API 15, so, I suppose that the API 15 is the lowest API level supported by MVVMCross.

martijn00:
Shouldn't this be added to the support repo if we add v4 t this?

Cheesebaron:
[**@IlSocio**](https://github.com/IlSocio) commented on [Aug 14, 2015, 4:40 PM GMT+2](https://github.com/MvvmCross/MvvmCross/pull/1087#issuecomment-131129566 "2015-08-14T14:40:53Z - Replied by Github Comment Enhancer"):
> I just added the Support v4 package... but I also had to change the compilation target from API 10 to API 15 because the Support v4 package contains only binaries for the API 15 (v4.0.3)
> 
> I also noticed that some other project (eg. CossUI.Droid) is already targeting the API 15, so, I suppose that the API 15 is the lowest API level currently supported by MVVMCross.

You are confusing target API with minimum target. Just because a library is using API 15 as target, does not mean it won't work on lower API levels. As long as you are not using methods which are not supported by API 10.

Cheesebaron:
[**@martijn00**](https://github.com/martijn00) commented on [Aug 14, 2015, 6:24 PM GMT+2](https://github.com/MvvmCross/MvvmCross/pull/1087#issuecomment-131168135 "2015-08-14T16:24:53Z - Replied by Github Comment Enhancer"):
> Shouldn't this be added to the support repo if we add v4 t this?

Download cache is a plugin. I guess you could make it depend on our Support v4 package and add a helper for it in there. I dunno. What do you think the appropriate way would be to do this?

martijn00:
I guess it will only be more confusing if we do a helper. So i think that indeed because it is a plugin we can have a dependency on v4.

Cheesebaron:
I guess we can merge this, although we need to fix the nuspec, so it pulls in Support v4 for DownloadCache on Droid.

martijn00:
Fixed nuspec here: https://github.com/MvvmCross/MvvmCross/blob/4.0/nuspec/MvvmCross.HotTuna.Plugin.DownloadCache.nuspec#L30

PR #1098, opened 2015-08-17
Fixed wrong result at IMvxReachability.IsConnected when it is called …

SButterfly:
…in timer.

At this case activeConnection.IsConnected throws native exception, which is handled at try-catch construction, and total result is false instead of true.


martijn00:
I am closing this for now. If still applicable you can reopen this pull request to the Plugins repo: https://github.com/MvvmCross/MvvmCross-Plugins

PR #1099, opened 2015-08-17
Use reflection instead of Expression.Compile()

ddunkin:
in MvxPropertyExpressionParser.

Expression.Compile() depends on Mono.Dynamic.Interpreter on iOS, which
adds about 1MB per architecture to iOS apps.

### DIFF
--- a/Cirrious/Cirrious.MvvmCross.Binding/ExpressionParse/MvxPropertyExpressionParser.cs
+++ b/Cirrious/Cirrious.MvvmCross.Binding/ExpressionParse/MvxPropertyExpressionParser.cs
@@ -82,22 +82,21 @@ namespace Cirrious.MvvmCross.Binding.ExpressionParse
 
             try
             {
-                var constExpr = memberExpr.Expression as ConstantExpression;
-                if (constExpr != null)
-                {
-                    var property = memberExpr.Member as PropertyInfo;
-                    if (property != null)
-                    {
-                        var constant = property.GetValue(constExpr.Value);
-                        return Expression.Constant(constant);
-                    }
+                var constExpr = ConvertMemberAccessToConstant(memberExpr.Expression) as ConstantExpression;
+                object value = constExpr != null ? constExpr.Value : null;
 
-                    var field = memberExpr.Member as FieldInfo;
-                    if (field != null)
-                    {
-                        var constant = field.GetValue(constExpr.Value);
-                        return Expression.Constant(constant);
-                    }
+                var property = memberExpr.Member as PropertyInfo;
+                if (property != null)
+                {
+                    var constant = property.GetValue(value);
+                    return Expression.Constant(constant);
+                }
+
+                var field = memberExpr.Member as FieldInfo;
+                if (field != null)
+                {
+                    var constant = field.GetValue(value);
+                    return Expression.Constant(constant);
                 }
             }
             catch
--- a/Cirrious/Cirrious.MvvmCross.Binding/ExpressionParse/MvxPropertyExpressionParser.cs
+++ b/Cirrious/Cirrious.MvvmCross.Binding/ExpressionParse/MvxPropertyExpressionParser.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Linq.Expressions;
+using System.Reflection;
 using Cirrious.CrossCore;
 
 namespace Cirrious.MvvmCross.Binding.ExpressionParse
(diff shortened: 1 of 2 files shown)


martijn00:
Thanks for the pr. Did you test this on all platforms?

jamie94bc:
I'd like to see how the performance of this differs.

Also is a 1MB a massive issue?

How many other libs use `Expression.Compile` that are likely to be used?

JSON.NET for example: https://github.com/JamesNK/Newtonsoft.Json/blob/f4e8c3d2cad5402ff2de73964052178effa55406/Src/Newtonsoft.Json/Utilities/ExpressionReflectionDelegateFactory.cs#L58

Used here (iOS references the Portable40 lib): https://github.com/JamesNK/Newtonsoft.Json/blob/8b19402e39910b4c35d9a58e1c5925dce33bc766/Src/Newtonsoft.Json/Serialization/JsonTypeReflector.cs#L420

Cheesebaron:
Also according to this article, reflection seems to be slower: http://www.palmmedia.de/Blog/2012/2/4/reflection-vs-compiled-expressions-vs-delegates-performance-comparision

jamie94bc:
That said, it doesn't look like the returned expression is reused or cached - so if it's "one-time" faster then it may be worth making the change.

https://github.com/MvvmCross/MvvmCross/blob/4.0/Cirrious/Cirrious.MvvmCross.Binding/ExpressionParse/MvxPropertyExpressionParser.cs#L71

kjeremy:
I'd really like to see some benchmarks before we pull this in.

kjeremy:
@Cheesebaron it looks like reflection is faster than expressions on the first run.  The decimal point in the chart is used as the grouping operator (thousands), not a decimal point.

ddunkin:
The size cost is 1MB per architecture, so 2MB for armv7 and arm64, or 2% of the OTA download budget. The app I'm working on is currently over 100MB, so 2MB is a big deal for me. :)

I'll get some performance numbers.

jamie94bc:
@ddunkin understood but `Mono.Dynamic.Interpreter` if you use other libs such as JSON.NET - which MvvmCross depends on.

ddunkin:
We use Newtonsoft.Json 6.0.4, which doesn't use compiled expressions.

Testing on an iPhone 6 with a release build, I get these measurements:
Compile took 470ms per 10000 expressions
Reflection took 62ms per 10000 expressions

Reflection is an order of magnitude faster, though it's only even measurable with an unrealistic number if iterations. With 20 expressions, more typical of loading and binding a view, it's 1ms vs 0ms.

jamie94bc:
@ddunkin understood! Can you share the source for the perf tests? I'm interested to know how this differs on other platforms such as Android which are JIT compiled.

ddunkin:
@jamie94bc here is the test code https://github.com/ddunkin/ExpressionPerfTest

SamirHafez:
I think this is not so much a case of Reflection vs Expression as it is a case for poor use of Expressions altogether: Notice that the generated expression is being compiled times and again (even when there is the possibility of it being the same expression).

Although this is easy to diagnose, in the end it might be difficult to solve.
Yes, we can solve this by caching the compiled expression but remember that Expressions do not provide us with a comparing algorithm out of the box (we would have to implement it ourselves, and that costs :smile:).

kentpickard:
Looks like this hasn't seen much action lately but I am very interested in anything that can improve binding performance.  I have a requirement to include 20 pieces of data into a UITableViewCell for iOS.  Using 20 fluent bindings with some light weight converters I am seeing noticeable lag when scrolling.  In this case, caching the compiled expression might yield the best overall results.  I tried to remove the converters and that does make a positive impact but the scrolling still isn't smooth.

Cheesebaron:
I thought we had merged this in. If @ddunkin could pull in latest stable so this can be merged it would be awesome. Otherwise I can put it on my todo to merge it manually.

kjeremy:
FWIW I think cached expressions could perform even better (and allow for more flexibility if we ever need it).

Also the benchmarks are only for iPhone.  Can we get numbers for Android as well?

ddunkin:
I've moved on to other things, so I um unlikely to have time to update this in the near future.

Cheesebaron:
@ddunkin OK, thanks for replying.

So this is up for grabs for someone to look at.

Cheesebaron:
Revisited this and tested on Android devices too. Reflection seems a lot faster there too. I think we should try get this added into MvvmCross and maybe make a preview release to see how apps behave.

kjeremy:
Excellent. What are the numbers?

Cheesebaron:
**Nexus 5x:**
Compile took 3192ms per 10000 expressions
Reflection took 109ms per 10000 expressions

**Samsung Galaxy Tab 2 10.2:**
ompile took 17854ms per 10000 expressions
Reflection took 548ms per 10000 expressions

**Nexus 4 Android Player Emulator:**
Compile took 1179ms per 10000 expressions
Reflection took 23ms per 10000 expressions

So seen from this, reflection is **much** faster even on slow and old devices such as the Galaxy Tab 2 I tested on.

jamie94bc:
Wow that's quite an improvement! Would be great to see this merged in.

softlion:
Trying this on iOS, including this in my work on binding optimizations for ios scrolling in collectionviews.

Cheesebaron:
Merged in #1321. Thanks @ddunkin!

PR #1103, opened 2015-08-19
Adding Cirrious.MvvmCross.WindowsUWP.rd.xml in nuspec.

michaeldaw:
The UWP package requires the file "Cirrious.MvvmCross.WindowsUWP.rd.xml" be distributed with it in order for projects using it to build properly. When the project is open in Visual Studio, this file is found under the Properties node of Cirrious.MvvmCross.WIndowsUWP. It's copied to the output folder during Release builds.

PR #1109, opened 2015-08-22
Generate lib layout to fix broken rd.xml reference in nupkg for UWP

krishna-nadiminti:
Hey guys, 

I've just tried the 4.0.0 beta 2 package on a new Windows UWP app.
My app fails to build complaining about a missing xml file from MVVMCross.

On investigating, it looks like it is indeed missing the Cirrious.MvvmCross.WindowsUWP.rd.xml in the nupkg - and it turns out that is because an option called 'Generate library layout' needs to be set to true for the file to be copies across to the output path.

I'm not entirely sure what this rd.xml does - but the solution is to set that option in the UWP project - so when it builds the rd.xml file is copied correctly.

This PR fixes that.
Hope this helps.

reference: https://social.msdn.microsoft.com/Forums/windowsapps/en-US/af9274cf-fb12-45df-aced-656509fed48d/rdxml-file-not-found?forum=wpdevelop


martijn00:
For beta2 it was also missing in the nuspec, but that is fixed here: https://github.com/MvvmCross/MvvmCross/pull/1103

Thanks for looking into this!

krishna-nadiminti:
Actually - just pushed a proper fix: I'd fixed it for Debug - not release.

PR #1117, opened 2015-08-26
Add CodeAnalysis Project with Roslyn Code Analyzer

### LINKED ISSUE #1040: MvvmCross as a Code-Aware Library with Roslyn Code Analyzers
Hi guys,

I've been kind of investigating and writing a sample Code Analyzer and Code Fix Provider with the new .NET Compiler Platform (Roslyn). After watching [this talk](https://channel9.msdn.com/Events/Build/2015/3-725) by Dustin Campbell, I was inspired to try and write one myself. I did and wrote about it on [my blog](http://stephanvs.com/mvvmcross-as-a-code-aware-library-through-code-analyzers/), for detecting this

```csharp
class FirstView : Activity
{
    public new FirstViewModel ViewModel { get { return base.ViewModel as FirstViewModel; } }
}
```

and transforming it into this

```csharp
class FirstView : Activity<FirstViewModel>
{
}
```

It's just one simple example I could think of myself, but I guess there are many more places where MvvmCross has a _certain way of doing things_ that can be detected, where we can provide guidance for developers and push them in the right direction (pit of success).

I'm really curious to find out what you think about it and if you maybe have any ideas on what other kinds of analyzers we can come up with?

Stephanvs:
# Background

With the release of C# version 6 Microsoft has rewritten their C# and VB compilers from the ground up. They are known as the Microsoft .NET Compiler Platform (aka Roslyn). These compilers provide library builders to provide the compiler with a set of so-called Analyzers and CodeFixes. These can be shipped through a nuget package, and will light up in an IDE with support for these Roslyn based compilers.

# What is this (aka TLDR)?

This pull-request is the first attempt to introduce such analyzers and code fix providers for the MvvmCross library. This fixes #1040.

# What does it do?

In short it helps developers use the MvvmCross library the way it was intended to be used.

This first analyzer will identify the following code:

```csharp
[Activity(...)]
class FirstView : MvxActivity
{
    public new FirstViewModel ViewModel => base.ViewModel as FirstViewModel;
}
```

and provide you with a IDE tooltip to transform the code into:

```csharp
[Activity(...)]
class FirstView : MvxActivity<FirstViewModel> { ... }
```

# What's in it?
- A new project type under the CodeAnalysis solution folder named `MvvmCrossCodeAnalysis` (there's no `.` in the name for reasons - no tooling support) I'm wondering what you guys think about that?
- A Vsix package to enable easy debugging of Code Analyzers/Fixes.

# What's still needed?
- The .nuspec file is a default generated one. It should probably be updated and moved into the root `nuspec` folder.

I'm very curious what you guys think about it, and I'm looking forward to some feedback!


martijn00:
Looking good! I can fix the nuspecs if you want.

Stephanvs:
That'd be great! Thanks

azchohfi:
Loved this!
I think there are many little but useful things that could be done, things like good practices with MVVMCross (checking if an public properties are re-setted (after constructor/Start) but does not have a set with a NotifyProperty (Set(ref x, value), which will cause the property to not update the View, or validations for namespace/folder for View/ViewModels.
Roslyn is amazing and we should definitely use it!

Stephanvs:
Superb! Thanks for merging!

PR #1121, opened 2015-08-29
Add initial support for a MvxPageViewController that dynamically crea…

scastria:
…tes UIViewController pages via additional child ViewModels.

Similar PR from before but based off of 4.0 branch


martijn00:
Thanks!

PR #1125, opened 2015-09-01
Fix nunit path in plugins

tekkies:
Since the projects moved, the relative path to nunit reference was incorrect.  Removed "..\"


martijn00:
Thanks!

PR #1126, opened 2015-09-01
Fix SelectionChangedCommand CanExecute not being called

charri:
SelectionChangedCommand for MvxBaseTableViewSource and MvxBaseCollectionViewSource checks if command can be executed (CanExecute). This enforces the same behavior as in android.


martijn00:
Thanks!

PR #1134, opened 2015-09-18
Implemented Fused location plugin for Android

SeeD-Seifer:
**Notes**
* FusedLocationApi used for location service access
* The name of the plugin is the same as default one (Cirrious.MvvmCross.Plugins.Location.Droid) so that plugin loader can find it
* Plugin can not be used with default Location.Droid plugin
* Plugin depends on GooglePlayServices

**Reason**
FusedLocationApi uses any available location provider, and it is now recommended approach.
See:
* https://developer.android.com/training/location/retrieve-current.html
* http://www.tothenew.com/blog/googles-fused-location-api-for-android/

**TODO**
* Include project into solution. My Xamarin Studio makes lots of changes into other .csproj files and .sln
* Migrate plugin to Mvx 4.x

**References**
Plugin fixes the issue #817


martijn00:
This looks good, but i think it belongs here: https://github.com/MvvmCross/MvvmCross-Plugins

SeeD-Seifer:
We've not migrated to Mvx 4.x yet, so we need this plugin for 3.5.x and I just decided to pull-request it.

As soon as we migrate to 4.x, I'll pull-request this plugin into the new repo too.

PR #1150, opened 2015-10-05
Discover CanExecuteXyz properties in ancestors.

dbeattie71:
Currently command builder discovers commands in ancestors but not CanExecuteXyz properties.


martijn00:
Is this safe to add, and backwards compatible?

dbeattie71:
I would think it would be safe unless for some reason I can't think of, it was intended.  Can anyone think of a reason why the same flags for discovering AbcCommand aren't used for discovering CanExecuteAbcCommand ?

martijn00:
Let's see if there is response on this after next update.

Issue #1132, opened 2015-09-14
Calling Init() on ViewModel creates exception without parameters


indazoo:
Hi

I found a small bug(?) in Cirrious.MvvmCross.ViewModels.CallBundleMethod().
If the parameter "bundle" is null, then CallBundleMethod(...) fails. This creates problems for me because this case occurs when I try to use design time data in Visual Studio. The error does not occur at runtime.
I'm not very familiar with the inner MVVMcross source code and do not really know at what place the check should be done. But with an additional line to check for bundle == null, the Designer no longer throws an exception.
Maybe the check should be much earlier in the stack because if there is no bundle there is no need to run CallBundleMethod and it should fail (don't no your coding style). Maybe there should be a IsDesigner check earlier? I found no way to intercept/override a method to do a designer check in my code after Start().

Anyway, that helped:
public static void CallBundleMethod(this IMvxViewModel viewModel, MethodInfo methodInfo, IMvxBundle bundle)
        {
            if (bundle == null) return;  //!!!! new line , DesignMode in Visual Studio needs this check
                        var parameters = methodInfo.GetParameters().ToArray(); 

I created a small visual studio project with latest MvvmCross code (a few days old):
https://github.com/indazoo/MvvmCross_DesignData/commits/master
There are two commits of today. The earlier one throws an Exception in Designer.

Your help would be greatly appreciated.

P.S. There is some confusion on my side about windows.Common and Windows Store and when to use what. The samples on the net and your fast progress with the renaming/moving (lately the plugins) produce a steep learning curve on my side. Maybe I miss something.

indazoo:
This MvvmCross thing is really good! ;-)  I start liking it!
I overrode MvxDefaultViewModelLocator in App.cs as seen here:
https://github.com/MvvmCross/MvvmCross/wiki/Customising-using-App-and-Setup
This works, but has the drawback is that I have to check for every new MvvmCross version if the code is valid/the same. If the check can be implemented into MvvmCross It would be great.
```
	using Cirrious.MvvmCross.ViewModels;
	using Cirrious.CrossCore;
	using Cirrious.CrossCore.Exceptions;

	public class DesignTimeViewModelLocator : MvxDefaultViewModelLocator
	{
		public override IMvxViewModel Load(Type viewModelType,
									IMvxBundle parameterValues,
									IMvxBundle savedState)
		{
			IMvxViewModel viewModel;
			try
			{
[...]

Issue #323, opened 2013-06-20
Design time support for Visibility and Color


slodge:
Following http://stackoverflow.com/questions/17207058/mvvmcross-throws-nullreferenceexception-in-windows-phone-designer

I wonder if we could use a technique like:

```
using System.ComponentModel;
using Cirrious.CrossCore;
using Cirrious.CrossCore.Core;
using Cirrious.CrossCore.IoC;
using Cirrious.CrossCore.UI;

namespace Cirrious.MvvmCross.Plugins.Visibility.WindowsPhone
{
    public class MvxPhoneVisibilityDesignTimeHelper
    {
        public MvxPhoneVisibilityDesignTimeHelper()
        {
            if (!DesignerProperties.IsInDesignTool)
                return;

            if (MvxSingleton<IMvxIoCProvider>.Instance == null)
            {
                var iocProvider = MvxSimpleIoCContainer.Initialise();
                Mvx.RegisterSingleton(iocProvider);
            }

            if (Mvx.CanResolve<IMvxNativeVisibility>())
                return;

            var forceVisibilityLoaded = new Plugin();
            forceVisibilityLoaded.Load();
        }
    }
}
```

This could then be added as a static resource to the app

```
   <whatever:MvxPhoneVisibilityDesignTimeHelper x:Key="DesignTime" />
```

PR #1151, opened 2015-10-07
Fix design time error

Luca7993:
The binding for mvxcommand cause designer error because in mvxcommandbase it use Mvx.TryResolve but the IMvxIoCProvider is not initialized.
One solution is to add the check if ioc is null in TryResolve method.

### DIFF
--- a/CrossCore/Cirrious.CrossCore/Mvx.cs
+++ b/CrossCore/Cirrious.CrossCore/Mvx.cs
@@ -42,12 +42,22 @@ namespace Cirrious.CrossCore
         public static bool TryResolve<TService>(out TService service) where TService : class
         {
             var ioc = MvxSingleton<IMvxIoCProvider>.Instance;
+            if (ioc == null)
+            {
+                service = null;
+                return false;
+            }
             return ioc.TryResolve(out service);
         }
 
         public static bool TryResolve(Type serviceType, out object service)
         {
             var ioc = MvxSingleton<IMvxIoCProvider>.Instance;
+            if (ioc == null)
+            {
+                service = null;
+                return false;
+            }
             return ioc.TryResolve(serviceType, out service);
         }
 


martijn00:
Is this related to https://github.com/MvvmCross/MvvmCross/issues/1132 ?

Luca7993:
I don't think is related to that problem.
the problem is in the constructor of MvxCommandBase:
 public MvxCommandBase()
        {
            if (!Mvx.TryResolve<IMvxCommandHelper>(out _commandHelper))
                _commandHelper = new MvxWeakCommandHelper();

            var alwaysOnUIThread = (MvxSingletonCache.Instance == null)
                                       ? true
                                       : MvxSingletonCache.Instance.Settings.AlwaysRaiseInpcOnUserInterfaceThread;
            ShouldAlwaysRaiseCECOnUserInterfaceThread = alwaysOnUIThread;
        }


The first row cause the crash at design time (because the IMvxIoCProvider is not initialized).
I think that you can change the row in MvxCommandBase, or you can directly change the tryresolve method in Mvx.cs as suggested

slodge:
Not sure why MvxCommandBase is used in the designer?!

The old (possibly out of date) way to avoid design time issues for ioc was to use designtimehelper - see https://github.com/MvvmCross/MvvmCross/issues/323 and pull https://github.com/MvvmCross/MvvmCross/commit/1c527322a600192de13d8bf8c5353ff2e92d0c8f

Luca7993:
Tomorrow i Will create an example Project.
It happens when you attach a design context to the page and set a command binding in a button
________________________________
Da: Stuart Lodge<mailto:notifications@github.com>
Inviato: ‎07/‎10/‎2015 19:21
A: MvvmCross/MvvmCross<mailto:MvvmCross@noreply.github.com>
Cc: Luca7993<mailto:lucapinasco@hotmail.com>
Oggetto: Re: [MvvmCross] Update Mvx.cs (#1151)

Not sure why MvxCommandBase is used in the designer?!

The old (possibly out of date) way to avoid design time issues for ioc was to use designtimehelper - see https://github.com/MvvmCross/MvvmCross/issues/323 and pull https://github.com/MvvmCross/MvvmCross/commit/1c527322a600192de13d8bf8c5353ff2e92d0c8f

---

Luca7993:
Hi,
here you can find a sample project where you can see the error:
http://1drv.ms/1hrWc74

I tried the way to avoid design time issues, but I can't make it work.
For the moment, I use a try catch in the command property of viewmodel to avoid the design exception

martijn00:
Maybe we could find inspiration in MvvmLight: https://marcominerva.wordpress.com/2015/09/23/using-ioc-and-design-time-support-with-mvvm-light-and-template-10/

slodge:
Does a design time helper not work for this? (e.g. like in MvxVisibilityDesignTimeHelper in https://github.com/MvvmCross/MvvmCross/wiki/Value-Converters)

PS Sorry - but I don't download zip's for source code as I once got a nasty surprise in one of them - much prefer to have source shared using repo's (sorry - it's nothing against you - just I don't like downloading unseen dropbox or onedrive content!)

Luca7993:
Hello Stuart,
Today I'm out of office and I can't prepare the small repository for the example, I think I will create it in the weekend, and I will also check the designhelper

Luca

Inviata dal mio Windows Phone
________________________________
Da: Stuart Lodge<mailto:notifications@github.com>
Inviato: ‎09/‎10/‎2015 12:35
A: MvvmCross/MvvmCross<mailto:MvvmCross@noreply.github.com>
Cc: Luca7993<mailto:lucapinasco@hotmail.com>
Oggetto: Re: [MvvmCross] Update Mvx.cs (#1151)

Does a design time helper not work for this? (e.g. like in MvxVisibilityDesignTimeHelper in https://github.com/MvvmCross/MvvmCross/wiki/Value-Converters)

PS Sorry - but I don't download zip's for source code as I once got a nasty surprise in one of them - much prefer to have source shared using repo's (sorry - it's nothing against you - just I don't like downloading unseen dropbox or onedrive content!)

---

Luca7993:
I have created the sample project, you can find it here: https://github.com/Luca7993/App9
the designtimehelper works if I explicit call it in the view model constructor. (look at comment "//******if remove comment from this line, everything is working").

Adding it as a resource in App.xaml doesn't work (or at least I can't make it work). 

Do you know if there is a more clever idea to call the initialization in designtimehelper without call it in the view model?
Or do you know how can I make it work using it as a resource?
I tried to call it in the page constructor, but it does not work...

thank you
Luca

slodge:
If you declare other resources in App.Xaml, can you access them in the page? In theory, the resource in App.xaml should work and get the designer happy... but in practice, not every theory works...

You should also be able to declare it as a page level resource and get it working... but app level would be most convenient...

It might also be a quirk of vs2015 I guess...

martijn00:
Fixed in https://github.com/MvvmCross/MvvmCross/pull/1438

PR #1157, opened 2015-10-18
Add MvxWindowsMultiRegionPresenter

Stephanvs:
Allows multiple Views to be rendered within the same parent view,
through use of named Frame based container views.


Stephanvs:
Read more about intended usage here: http://stephanvs.com/implementing-a-multi-region-presenter-for-windows-10-uwp-and-mvvmcross/

I'm curious what you guys think of it!

martijn00:
I think the idea of this is good! I am just wondering what happens when the frame you target is on another page. I think this is a bit the same as https://github.com/MvvmCross/MvvmCross-AndroidSupport/issues/43 for Android, where we have that problem too.

Stephanvs:
If I'm correct the `FindChild` method looks in runtime for any subview with a name defined as `x:Name` that matches the attribute, thus it should work even in views nested in other pages as long as it's loaded in the current Visual Tree, but you're right that it's an interesting scenario to test out. I was also working on a sample which uses this presenter, and will add the case over there to try it :grin:

martijn00:
Thanks! Merging this for now to have in the next beta. Please open a new PR if you have improvements after your tests and samples.

PR #1163, opened 2015-10-29
Create MvxUIActivityIndicatorViewHiddenTargetBinding

munkii:
The default binding for UIActivityIndicatorView Hidden will only show\hide the view. It needs to be told to spin and that is not a property it is a method. This binding does that calling of the appropriate method as well as the setting of Hidden

### DIFF
--- a/Cirrious/Cirrious.MvvmCross.Binding.Touch/Cirrious.MvvmCross.Binding.Touch.csproj
+++ b/Cirrious/Cirrious.MvvmCross.Binding.Touch/Cirrious.MvvmCross.Binding.Touch.csproj
@@ -106,6 +106,8 @@
   <Target Name="AfterBuild">
   </Target>
   -->
-  <ItemGroup />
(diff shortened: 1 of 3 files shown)


lothrop:
Thanks, this looks like something I could have used in the past.

The file ending `.cs` is missing and the file isn't referenced in any `.csproj` file. Could you please fix those two things?

munkii:
That'll teach me to copy and paste out of our project, sorry

munkii:
Done

Cheesebaron:
Why is this in the 3.2 branch?

munkii:
That was the branch I had forked when I was last making updates. My GitHub skills are bloody terrible. I can Fork the latest if you want and add it there. Probably won't be until Monday now

Cheesebaron:
There is no reason to add it to the 3.2 branch, it isn't even the main branch. How you got it tangled into that branch I have no idea. This goes into 4.0.

Issue #949, opened 2015-04-01
Use Same Bundle for InitFromBundle and SaveStateToBundle


JimWilcox3:
When a ViewModel is possibly being TombStoned, SaveStateToBundle is called.  When the ViewModel is started back up, Init is called before LoadStateFromBundle.  This causes fresh data to be loaded before you even know if you need to restore from a tombstone.  If you used the same bundle for InitFromBundle, then you could check the "parameters" bundle to see if there is saved state data in there and load the saved data from it.  If not, you can use the parameters in that bundle to load fresh data.

Alternatively, you could call LoadStateFromBundle before you call any flavor of Init.  That way you could see if data is present before you load fresh data.

The bottom line is that in the Init, you need the opportunity to load saved data if recovering from a tombstone or load fresh data if it is a new navigation.

cyrilcathala:
You should only load data within the Start() method. Here is what is called (in order) :
- Init() is for initializing navigation parameters.
- LoadStateFromBundle() is for initializing tombstoning parameters.
- Start() should state whether it should load fresh data or not, based on the data retrieved in previous methods

vvolkgang:
@martijn00 about #1165 fixing this, I think @JimWilcox3 needs to change his workflow based on @cyrilcathala comments.

JimWilcox3:
@martijn00 That is what I have done.  I will close this issue.

PR #1165, opened 2015-10-30
Improved ViewModelLoader and ViewModelLocator with viewmodel reloading

vvolkgang:
This should enable cached viewmodels to go through the lifecycle again if they are being showed multiple times with different parameters. My current usecase was to properly support fragment caching in AndroidSupport.


vvolkgang:
More on that discussion in MvvmCross/MvvmCross-AndroidSupport/#108

martijn00:
Look like it could work, but hard to test without the implementation for it in Android. Would it be an idea if you(@vvolkgang) make the PR for Android based on this too, and them we merge them at the same time.

martijn00:
@vvolkgang is it correct that this also fixes: https://github.com/MvvmCross/MvvmCross/issues/949

vvolkgang:
@martijn00 commented there!

PR #1178, opened 2015-11-17
MultipleViewModel cache was not prepared for multiple instances of ViewModel

thefex:
As in title.
It is required to fix multiple instances of same type fragment caching.


martijn00:
Look good to me! Did you test this enough to be sure it doesn't make regressions?

thefex:
Well. There is nothing to cause regression as default behavior is persisted with default viewModelTag string value "singleInstanceCache".

I also tested this on my production app - it's ok.

martijn00:
What about full fragging? That needs the caching changes as well.

thefex:
Yes it is.
Full Fragging is a lot commits behind Support library. 
I gonna take care of that but not today.

PR #1179, opened 2015-11-17
4.0 android search view custom binding

mgj:
Add custom binding for android SearchView, so that a SearchView's query-text can be bound to a string in the viewmodel


martijn00:
Looks good to me! And useful as well!

PR #1186, opened 2015-12-02
Moved caching logic from support library.

thefex:
This fixes a lot of issues and crashes discovered during implementation of SupportFragging Cache.

Moved Cache logic from https://github.com/thefex/MvvmCross-AndroidSupport - can someone test it before merging? I do not use FullFraggings, are there any samples?


martijn00:
There is no sample, but it should work exactly like the support code. If that one work, this should work too. Thanks!

PR #1188, opened 2015-12-02
Fix merge of PR 1186

dbeattie71:
I think this is correct, extension methods possible didn't get added to be tracked?


martijn00:
Thanks!

PR #1195, opened 2015-12-19
Restructure

djrocksmith:
Added missing Droid project to Droid folder and added to solution.
Updated usings and namespaces to remove Cirrious.
Update of Windows project files for assembly names not to use cirrious.

Removal of all Cirrious references in the entire solution.  Including in Assembly.info's and registration of Android Mvx Views.  Updated typeof assembly including to not look for Assemblies with Cirrious in both Droid and iOS.

Mac is still broken as I'm unsure how to fix with Unified
iOS and Android appear to be working
Missing WindowsStore and WPF files, maybe from a copy across?
Major refactor and has the potential to have breaking changes, though it is important to do.


martijn00:
Kind of hard to see with all those changes if it is correct, but i'll merge it and check the project.

PR #1196, opened 2015-12-21
Use XIB and auto layouts for iOS Starter Pack example

lothrop:
Now that there is a visual XIB file editor in both Visual Studio and Xamarin Studio I think the example provided in the Starter Pack should not use a hand-coded UI with absolute coordinates (from the times of 3.5" iPhones). Having an editor should scare away less people who are migrating to MvvmCross from a Windows background.

My changes have not been tested. How can I test this? I'm not sure where to declare how the XIB file is to be included in the csproj. It should show up as

    <InterfaceDefinition Include="Views\FirstView.xib" />

in the csproj. Also, the designer file should show up as a nested file

    <Compile Include="Views\FirstView.designer.cs">
      <DependentUpon>FirstView.cs</DependentUpon>
    </Compile>

Has anyone done this before?


martijn00:
Change itself looks good. I am not sure about your questions, never done that before.

lothrop:
How do I test the nuget package locally?

Cheesebaron:
1. Create the nuget package. 
2. Add the folder to your Nuget sources where the nupkg is in. 
3. Install it
4. ??
5. Profit

lothrop:
Thanks Tomasz! The tests worked. The .designer.cs file is not nested under the .cs file but it still works.

PR #1199, opened 2015-12-27
Windows Store and Windows Phone

djrocksmith:
Re-linked up files that were missing in the csproj files

I believe there may be some compiler directives not laid out in order for it to build correctly, however I'm not sure.


lothrop:
Thanks! I wonder why Visual Studio puts some paths as lowercase into the csproj files?

Issue #856, opened 2014-12-11
Mac and XamMac nugets


tofutim:
Hey Stuart, will you or have you posted the Mac and XamMac nugets? I couldn't find them.

lothrop:
Should we do this before the 3.5 release? I'm not sure of the implications.

lothrop:
@tofutim I'm going to give this another go. Do you have any information on why the previous attempt failed?

lothrop:
@slodge Can you tell me which puzzle piece was missing?

lothrop:
Fixed in #1209.

PR #1207, opened 2016-01-03
Fixes to get pack.ps1 to run again.

lothrop:
Made pack.ps1 run again.
Updated nuspecs to beta8.
Remove outdated MvvmCross.Droid.Maps.


lothrop:
@Cheesebaron, @martijn00: Are we currently not using the Xamarin.Mac nuspecs?

martijn00:
You mean Xamarin.Mac as targetFramework? I didn't knew that exists, but i guess it could be added. I also see MonoAndroid is used and not Xamarin.Android, should that be changed too?

lothrop:
You mean the referenced assembly? I don't know of a Xamarin.Android.dll. Xamarin.Mac.dll and Xamarin.iOS.dll were introduced with the Unified API breaking change. AFAIK no such change was ever done for Android.

lothrop:
Ah, I just got what you were asking. I think we could add the targetFramework for Mac, too. It appears to be `xamarinmac20` for Xamarin.Mac Unified.

What I was asking: There are three nuspecs specifically for Mac. None of them seem to be referenced in pack.ps1. It appears they are not being published?

https://www.nuget.org/packages?q=MvvmCross.HotTuna.Mac.StarterPack

lothrop:
See also #856. I see we haven't followed the blog entry's advice yet. Should I?

martijn00:
@lothrop I think that from the blog would be good to do. I also noticed some of the projects don't get their symbols published, so that needs to be checked as well.

Also the Mac should be added so they will be published.

lothrop:
Might this be the reason?

https://github.com/MvvmCross/MvvmCross/blob/restructure/nuspec/pack.ps1#L15

martijn00:
Hmm, might be. @Cheesebaron would know.

Cheesebaron:
If they are not in the pack.ps1 script, they are not released on NuGet.

lothrop:
Do you know why there are no symbols for those two packages?

https://github.com/MvvmCross/MvvmCross/blob/restructure/nuspec/pack.ps1#L15

lothrop:
Ah, I see, it produces an error message otherwise.

PR #1209, opened 2016-01-04
Nugets for Mac

lothrop:
I re-added MvvmCross_Mac.sln since Xamarin Studio on Mac was trying to build some Windows projects (like WPF) and failing. Perhaps this could also be solved in the Configuration Manager with a new configuration or platform.


lothrop:
This fixes #856.

PR #1211, opened 2016-01-06
Added dummy packages to ease migration from HotTuna.

lothrop:


martijn00:
Maybe we should put this in a separate folder "deprecated" under nuspecs

lothrop:
Done. Should we create the nuspecs for the other projects, too?

martijn00:
I think some others like android an forms only existed in beta. They might not require nuspecs?

lothrop:
Right. That would leave the plugins.

PR #1213, opened 2016-01-07
Made MvvmCross.StarterPack pull in MvvmCross, not the other way around.

lothrop:
If you want example files, install StarterPack (and uninstall it right away).
If you don't want the samples, install MvvmCross.


Cheesebaron:
MvvmCross package itself does not do anything. As discussed on Slack it should probably pull in Core and Binding + the proposed change from @lothrop

lothrop:
That was already in the changeset...

Cheesebaron:
yes, missed that :) Looks good!

Issue #1217, opened 2016-01-15
NotifyPropertyChanged on MvxAdapter is not marshalled to MainThread


CSkoubo:
For instance adding items to an observable collection on another thread. NotifyPropertyChanged will be called on incorrect thread.

```
[mvx] 123.11 Exception masked during Adapter RealNotifyDataSetChanged AndroidRuntimeException: Only the original thread that created a view hierarchy can touch its views.
[mvx] 	  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x0000c] in <filename unknown>:0 
[mvx]   at Android.Runtime.JNIEnv.CallNonvirtualVoidMethod (IntPtr jobject, IntPtr jclass, IntPtr jmethod) [0x00083] in /Users/builder/data/lanes/2098/3efa14c4/source/monodroid/src/Mono.Android/src/Runtime/JNIEnv.g.cs:1010 
[mvx]   at Android.Widget.BaseAdapter.NotifyDataSetChanged () [0x00058] in /Users/builder/data/lanes/2098/3efa14c4/source/monodroid/src/Mono.Android/platforms/android-21/src/generated/Android.Widget.BaseAdapter.cs:345 
[mvx]   at Cirrious.MvvmCross.Binding.Droid.Views.MvxAdapter.RealNotifyDataSetChanged () [0x00000] in <filename unknown>:0 
[mvx]   --- End of managed exception stack trace ---
[mvx] android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
[mvx] 	at android.view.ViewRootImpl.checkThread(ViewRootImpl.java:6334)
[mvx] 	at android.view.ViewRootImpl.requestLayout(ViewRootImpl.java:873)
[mvx] 	at android.view.View.requestLayout(View.java:16754)
[mvx] 	at android.view.View.requestLayout(View.java:16754)
[mvx] 	at android.view.View.requestLayout(View.java:16754)
[mvx] 	at android.view.View.requestLayout(View.java:16754)
[mvx] 	at android.view.View.requestLayout(View.java:16754)
[mvx] 	at android.view.View.requestLayout(View.java:16754)
[mvx] 	at android.widget.RelativeLayout.requestLayout(RelativeLayout.java:352)
[mvx] 	at android.view.View.requestLayout(View.java:16754)
[...]

PR #1218, opened 2016-01-18
Main thread marshal

Cheesebaron:
As described in #1217 NotififyDataSetChanged does not always happen on the UI thread, which will result in throwing an exception.

This has been changed and removed the masking of exceptions in NotifyDataSetChanged.

### DIFF
--- a/MvvmCross/Binding/Droid/Views/MvxAdapter.cs
+++ b/MvvmCross/Binding/Droid/Views/MvxAdapter.cs
@@ -5,6 +5,8 @@
 //
 // Project Lead - Stuart Lodge, @slodge, me@slodge.com
 
+using Android.OS;
+
 namespace MvvmCross.Binding.Droid.Views
 {
     using System;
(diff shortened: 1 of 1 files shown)


slodge:
Surely this shouldn't happen?

martijn00:
Same thing should be done for the RecyclerView i think.

slodge:
Afaik... (and my knowhow may be out of date...) there should never be updates to items in an ObservableCollection done on a background thread... it's up to the ViewModel author to work out how to avoid this - i.e. the ViewModel has to marshal the ViewModel update code to the UI thread - there is a `InvokeOnMainThread` base class method to help with this.

Posting individual changes asynchronously is dangerous - because if you start doing incremental collection changes asynchronously then it's all too easy for the UI and ViewModel to get array-out-by-one errors (or worse) - e.g. by the time the notify update message arrives in the UI that an item was inserted in position 10, then position 10 might be position 12... or it might have been deleted or...

The UI control has to be sure that the answers remain consistent when it asks GetCount, GetItemPositionAt, etc - it can't be surprised by some background thread changing the data underneath it.

There are people who've tried things to workaround this - mainly by holding two copies of the collection (one that's sync updated with the UI and one that can be manipulated on the background thread) - e.g. things like http://stackoverflow.com/a/14602121/373321 and http://www.codeproject.com/Articles/183495/Asynchronous-Multi-threaded-ObservableCollection - and I think some of the newer Windows controls have special options to allow them to update asynchronously... but none of this for Droid or iOS (yet?)

Cheesebaron:
Hmm you might be on to something here. I know for sure properties in a MvxNotifyPropertyChanged type of model, when RaisePropertyChanged is called is marshalled to the UI thread per default. So I guess we have to make it super clear when using ObservableCollection or derivatives that it is up to the dev to fix that themselves.

slodge:
Here's an old Q&A on it - http://stackoverflow.com/a/18469700/373321

PR #1220, opened 2016-01-19
Adapter long-life subscription sub-issue on #1217

slodge:
may need additional fixes in other adapter like classes


martijn00:
Woooh!! @slodge does a PR again!

PR #1225, opened 2016-01-26
Missing changes for adapter long-life subscription sub-issue on #1217

roubachof:
based on stuart correction for MvxListView


martijn00:
Thanks!

Issue #35, opened 2012-10-05
Using MvxBindableListView makes Commands pointing on Models instead of ViewModels


hugoterelle:
Hi Stuart, 

As discussed on stackoverflow (http://stackoverflow.com/questions/12682082/mvvmcross-changing-viewmodel-within-a-mvxbindablelistview), here is my point of view on the ideal axml file:

articles_view.axml

```
   ...
          <Mvx.MvxBindableListView
                android:id="@+id/listviewArticle"
                android:choiceMode="singleChoice"
                android:layout_width="fill_parent"
                android:layout_height="fill_parent"
                android:orientation="vertical" 
                local:MvxItemTemplate="@layout/article_rowlayout"
                local:MvxBind="{'ItemsSource':{'Path':'Articles'}}" />    
   ...


And here is the "article_rowlayout.axml"

   ...
            <TextView
                android:id="@+id/rowArticleLabel"
                android:layout_width="0dip"
                android:layout_weight="14"
                android:layout_height="wrap_content"
                android:textSize="28dip"
                local:MvxBind="{'Text':{'Path':'Item.Label'}}" />
            <ImageButton
                android:src="@drawable/ic_modify"
                android:layout_width="0dip"
                android:layout_weight="1"
                android:layout_height="wrap_content"
                android:id="@+id/rowArticleButtonModify"
                android:background="@null" 
                android:focusable="false"
                android:clickable="true"    
                local:MvxBind="{'Click':{'Path':'MyCommand'}}"
              />
   ...
```

The "Item" word should be a "reserved word" pointing to the "Artcicles[x]"
And without "Wrapper" programming, the "article_rowlayout.axml" should always point on the ViewModel of the MvxBindableListView instead of working in the "Articles[x]" context.  

Another point of view should be to choose the viewmodel of the MvxBindableListView in the parameters: 

```
   ...
          <Mvx.MvxBindableListView
                android:id="@+id/listviewArticle"
                android:choiceMode="singleChoice"
                android:layout_width="fill_parent"
                android:layout_height="fill_parent"
                android:orientation="vertical" 

                local:MvxViewModel="xxx"

                local:MvxItemTemplate="@layout/article_rowlayout"
                local:MvxBind="{'ItemsSource':{'Path':'Articles'}}" />    
   ...

```

Where "xxx" should be the class we would like to use, or maybe a reserved word telling to use the current viewmodel (the one of the MvxBindableListView). 

Other points of view should be nice ...  

[...]

PR #1232, opened 2016-02-07
[Wip] Example of parent binding code - works for Droid collection views currently

slodge:
#35 seems to have been open quite a long time...

This code uses `$parent` to allow a parent data context to be accessed for any view that knows what it's parent is - this is currently is only Droid `MvxListItemView`s hosted inside `MvxAdapter`s, but it could be extended to other views such as fragments (see code in `MvxBaseListItemView` and `MvxAdapter` which provides the additional parent linking...)

There's a sample of this binding working in https://github.com/MvvmCross/MvvmCross-Samples/blob/parent-binding-experiment/ApiExamples/ApiExamples.Droid/Resources/Layout/Item_Parented.axml

```
    <TextView
        android:layout_width="fill_parent"
        android:layout_height="wrap_content"
        local:MvxBind="Text  'From Item:' + . + ' From ParentVM:' + $parent.DebugOutput" />
    <Button
        android:layout_width="fill_parent"
        android:layout_height="wrap_content"
        android:text="Tap Me"
        local:MvxBind="Click $parent.Hello" />
```

The code's not perfect... as it has to abuse some `object` to `MvxEnhancedDataContext` casting - but doing anything else would have required making breaking changes to the binding object interfaces (and are plenty enough breaking changes already so I didn't fancy adding more!)

This is really just at a hack level right now.... To complete this pull request....
- [x] prototype code should be written
- [ ] need to decide if this is wanted.... and if `$parent` is the right approach... (`$parent` is the only thing I've ever really heard people asking for....)
- [ ] code needs to be tidied up, refactored, written properly, etc (there's just something a bit too "hacky-go-lucky" about the code in https://github.com/MvvmCross/MvvmCross/commit/68e556098360216d144c8bea6b67039c111c90e8#diff-2a644c81f6260239ebb50b8f9fe1059dR33 right now :8ball: )
- [ ] Droid code for other controls could be written (is this just fragments? Or others too? Maybe Mvx frame controls would be needed)
- [ ] iOS code needs to be written - should be possible to get something working for tableviewsource and collectionviewsource - beyond that maybe would need to look at how other `MvxView` type controls could support parent access...
- [ ] unit tests should be added (I added a couple of parser ones, but no binding ones)
- [ ] android, ios, winphone, etc would need testing.... what has this hacking broken?
- [ ] field binding, command binding, reactive binding and any other extensions that have been written should be tested...
- [ ] somehow would need to check there are no leaks here.... I think there aren't but I've thought that before :)
- [ ] somehow would need to check there are no gaps in the assumption that parent can't be changed without the control getting a notification. I'm pretty confident this assumption is good for Droid list views... but is it safe for other view types?
- [ ] docs updated
- [ ] resharper code stasi would need to be applied

No hurry on any of this.... #35 is not exactly new ;) Don't know how much time I'll get on any of it... so very happy for others to work on the pull...

### DIFF
--- a/MvvmCross/Binding/Droid/BindingContext/MvxAndroidBindingContext.cs
+++ b/MvvmCross/Binding/Droid/BindingContext/MvxAndroidBindingContext.cs
@@ -21,8 +21,8 @@ namespace MvvmCross.Binding.Droid.BindingContext
         private readonly Context _droidContext;
         private IMvxLayoutInflaterHolder _layoutInflaterHolder;
 
-        public MvxAndroidBindingContext(Context droidContext, IMvxLayoutInflaterHolder layoutInflaterHolder, object source = null)
-            : base(source)
+        public MvxAndroidBindingContext(Context droidContext, IMvxLayoutInflaterHolder layoutInflaterHolder, object source = null, object parentSource = null)
+            : base(source, parentSource)
         {
             this._droidContext = droidContext;
             this._layoutInflaterHolder = layoutInflaterHolder;
--- a/MvvmCross/Binding/Droid/Views/MvxAdapter.cs
+++ b/MvvmCross/Binding/Droid/Views/MvxAdapter.cs
@@ -5,6 +5,11 @@
 //
 // Project Lead - Stuart Lodge, @slodge, me@slodge.com
 
+using System.Collections.Generic;
+using MvvmCross.Binding.BindingContext;
+using MvvmCross.Binding.Bindings;
+using MvvmCross.Binding.Bindings.Source.Construction;
+
 namespace MvvmCross.Binding.Droid.Views
 {
     using System;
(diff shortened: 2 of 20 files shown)


martijn00:
Looks like a good start to me! I've been looking for something like this since i started with MvvmCross!

SergioZgz:
HI,

I think that people who comes from wpf or similar will appreciate if that "parent" but maybe add something like relativesource with ancestorlevel or maybe better elementname to get viewmodel from any ui element that you want.

I will try to improve your shared code and I will try to do that (I think that is out of my knowledge but could be funny try it :) )

Cheesebaron:
Closing this as this PR since nothing has happened in this one for a very long time.

If you want to work on this again, please feel free to open a new PR.

PR #1243, opened 2016-02-15
Fix for forced fade animation on replaced rows

bspinner:


bspinner:
We should consider setting the default value to Fade, so existing apps won't break.

lothrop:
Thanks!

PR #1244, opened 2016-02-18
Removes mandatory parameters from MvxWrappingCommand

### LINKED ISSUE #1240: Shouldn't MvxWrappingCommand implement IMvxCommand?
Since the parameter used in the call to `Execute(object parameter)` [is ignored anyway](https://github.com/MvvmCross/MvvmCross/blob/f7fcf18d960f578b851837f2aaaeb4d0e3b72364/MvvmCross/Core/Binding/ValueConverters/MvxWrappingCommand.cs#L60-L62), wouldn't it be good to have `MvxWrappingCommand` implementing IMvxCommand so we can call `Execute()` instead of `Execute(null)`? 

I can make a PR myself if this sounds like a good idea

willsb:
Fixes #1240


martijn00:
Thanks!

PR #1249, opened 2016-02-28
MvxWindowsMultiRegionViewPresenter - support searching for a frame which exists inside a frame

ErezG:
The current code does not support searching for a frame which exists inside a frame.
I encountered that problem in a project I am working on.
The fix above works in my project.


martijn00:
Thanks!

Stephanvs:
Nice addition! Thanks for contributing.

Issue #955, opened 2015-04-07
Reason for multiple initialisation should not be called simultaneously?


jihlee:
In MvxAndroidSetupSingleton, EnsureInitialized() is designed not to be called simultaneously.
In detail, if one thread is running _setup.Initialize() and other threads call EnsureInitialized() again, it goes into locked critical section, and throw MvxException("Multiple initialize calls made") since initialisation is not ended.

My application has GCM notification service and it updates the app's view. It was developed by IMvxMessenger, and I should call EnsureInitialized() of MvxAndroidSetupSingleton in GCMIntentService.

The thing is that if the application is starting and a GCM message is received at that moment, EnsureInitialized() is called again by GCM Service, and MvxException is thrown.

To resolve the issue, I would like to change EnsureInitilized() work as
- It can be called simultaneously
- Mvx system is initialised only once though.
- If the method is called multiple times, the first caller does initialise the system and others just wait there.
- The first caller notify others when the system is prepared.

Thanks

Cheesebaron:
Reopen this with more information.

monesprit54:
Added description of the issue and reopen the issue.
I submitted first ugly empty issue by mistake :( 

Thanks for fast reply!!

monesprit54:
Please confirm the suggestion would be feasible or not. Is there any reason the method should not be called multiple times?

slodge:
I'm not sure I understand how you intend to implement this.

I'm especially worried about the line "others just wait there" - not sure how they are waiting - is this async awaiting? is it UI thread blocking? something else? Blocking threads often leads to other problems :(

Can you demonstrate what you propose - e.g. it should be possible to write a small wrapper class (your own "setup singleton") which should have the behaviour you are looking for?

Alternatively, it might be possible to throw a more specific error? Or it might be possible to expose an event (or for you to hook into an existing setup event)? Together these things would allow you to implement your own mechanism for pausing and then processing the setup queue.

davidschwegler:
https://github.com/MvvmCross/MvvmCross/blob/3.5/Cirrious/Cirrious.MvvmCross.Droid/Platform/MvxAndroidSetupSingleton.cs#L36

I'm actually in the same boat as monespirit54. 

When the user starts the app via the main launcher/splash screen, it does an async load of MvvmCross. Any calls to EnsureInitialized will cause an exception during this time. 

[...]

PR #1253, opened 2016-03-04
Allow multiple EnsureInitialized calls

andyci:


martijn00:
What is there left to be done for this?

andyci:
Not sure, I have been using it in my application successfully so perhaps it just needs to be reviewed as it is a pretty integral part of the framework

martijn00:
I just came across https://github.com/MvvmCross/MvvmCross/issues/955 There is some interesting code in there. Can we make sure we can combine your and that code, to fix both cases?

andyci:
That code is slightly different.  This pull request makes a similar change but keeps the EnsureInitialised method synchronous so we don't have to update all the places we use EnsureInitialised and turn them async.

davidschwegler:
@martijn00 @andyci 

This PR is a good start, but has a critical bug in it. 

Multiple calls to EnsureInitialize probably work fine. Calling EnsureInitialize and then InitializeFromSplashScreen also probably works. However, calling InitializeFromSplashScreen and *then* EnsureInitialize crashes.

1) App calls InitializeFromSplashScreen (eg app starts up normally)
   - [_initializationStarted is set to true](https://github.com/MvvmCross/MvvmCross/pull/1253/files?diff=split#diff-7df39d8e4f2f85a961e1d3887d24e817R88) and [async init starts](https://github.com/MvvmCross/MvvmCross/pull/1253/files?diff=split#diff-7df39d8e4f2f85a961e1d3887d24e817R94)
2) While initializing async, EnsureInitialized is called on the main thread (lets say some push notification arrives to the device and it needs to process it)
   - [_initializationStarted is NOT checked](https://github.com/MvvmCross/MvvmCross/pull/1253/files?diff=split#diff-7df39d8e4f2f85a961e1d3887d24e817R49) (this is the bug), and this._setup.Initialize() is called anyways
  - [MvxException is thrown](https://github.com/MvvmCross/MvvmCross/blob/4.0/MvvmCross/Core/Core/Platform/MvxSetup.cs#L44) because setup has already started. 

Workaround? None. I can't just catch and ignore this exception, because any later calls will be permanently stuck waiting on IsInitialisedTaskCompletionSource because [SetResult](https://github.com/MvvmCross/MvvmCross/pull/1253/files?diff=split#diff-7df39d8e4f2f85a961e1d3887d24e817R58) will never be called on it.

davidschwegler:
Please do create a separate issue if that's better -- just wanted to initiate the discussion here since that's where the change was made. 

We currently can't upgrade to 4.2 with this code in it, though. (I'm the author of the workaround to the previous issue described in #955, btw).

Issue #1260, opened 2016-03-16
Missing MvxLayoutInflaterCompat+FactoryWrapper2 JNI constructors


thefex:
MvxFragments randomly crash on view inflate call during OnCreateView:
this.BindingInflate(ResourceId null);

### Steps to reproduce

Can't reproduce this. Random crashes especially when "Don't keep activities" is on.

### Expected behavior
Should not crash.

### Actual behavior
Crash with exception:

 System.NotSupportedException: Unable to activate instance of type MvvmCross.Binding.Droid.Binders.MvxLayoutInflaterCompat+FactoryWrapper2 from native handle 0xeda00025 (key_handle 0xb30c2540). ---> System.MissingMethodException: No constructor found for MvvmCross.Binding.Droid.Binders.MvxLayoutInflaterCompat+FactoryWrapper2::.ctor(System.IntPtr, Android.Runtime.JniHandleOwnership) ---> Java.Interop.JavaLocationException: Exception of type 'Java.Interop.JavaLocationException' was thrown.

### Configuration

**Version:** 4.0.0

**Platform:** Android

PR #1261, opened 2016-03-16
Added missing JNI interop constructors in MvxLayoutInflaterCompat

thefex:
When Fragment/Activity is destroyed, it looks like managed peer instance of MvxLayoutInflaterCompat is sometimes disposed. JNI could not recreate instance of inflater because of missing IntPtr, JniHandleOwnership constructor - therefore application crash with UnsupportedOperationException.

Fixes issue #1260

### DIFF
--- a/MvvmCross/Binding/Droid/Binders/MvxLayoutInflaterCompat.cs
+++ b/MvvmCross/Binding/Droid/Binders/MvxLayoutInflaterCompat.cs
@@ -20,6 +20,12 @@ namespace MvvmCross.Binding.Droid.Binders
         {
             protected readonly IMvxLayoutInflaterFactory DelegateFactory;
 
+            public FactoryWrapper(System.IntPtr handle, Android.Runtime.JniHandleOwnership ownership) : base(handle, ownership)
+            {
+                
+            }
+
+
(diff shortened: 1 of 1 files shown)


martijn00:
It seems good to merge, but it seems you didn't merge in the changes of the master into your own fork. So i can't merge this.

thefex:
sorry fixed.

thefex:
shit... VS removed end line, wait a while for a proper pull

martijn00:
Hmm this seems a bit messed up with all those changes. Maybe you can copy your changes into a fresh branch of the current master?

thefex:
sure thing, but later on - do not merge that. I need to dig into my resharper/vs settings cause it automatically changes file encodings - removes ENDLINE -.-

Cheesebaron:
Otherwise we can just cherrypick the change.

kjeremy:
I think I see what you tried to do.  Good catch.  I sometimes forget that that constructor is required.

thefex:
@kjeremy Technically speaking, this constructor should not be used in most cases. It often hides bugs about too-early object disposal (let it be too early explict Dispose call). The only place where it should be mandatory are java-land objects which can be touched by Android Runtime lifecycle methods -> Fragment, Activities, Views.

PR #1262, opened 2016-03-18
Fix Android crash when Bitmap updated after MvxImageView destroyed

SeeD-Seifer:
ImageHelperOnImageChanged method propogates bitmap update on the UI thread,
which leads to crash in case MvxImageView object has been disposed by Android runtime.
This simple fix avoids the crash.


martijn00:
Thanks!

PR #1266, opened 2016-03-22
Added swipe-gesture-recognizer

SimonSimCity:
I've added the swipe gesture recognizer. One of the comments said, that there weren't a use-case, but I see a use-case for it. I didn't want to set a default-swipe-direction so the developer has to think about it. If he get's a default, he may get the intention, that this captures all directions.

Here's an example how you can use it. You can also apply multiple directions to one view.

    set.Bind(MyView.Swipe(UISwipeGestureRecognizerDirection.Right)).For(v => v.Command).To(vm => vm.SwipedRightCommand);
    set.Bind(MyView.Swipe(UISwipeGestureRecognizerDirection.Left)).For(v => v.Command).To(vm => vm.SwipedLeftCommand);


martijn00:
Thanks!

PR #1267, opened 2016-03-22
Add bitmap existance check in Droid's MvxImageView

SeeD-Seifer:
Partial fix of https://github.com/MvvmCross/MvvmCross-Plugins/issues/41
Avoids crashes when Bitmap already disposed.


martijn00:
Thanks!

PR #1268, opened 2016-03-22
Move shared fragment logic to shared lib

martijn00:
- [x] Fix Android support libs to use the new library
- [x] Check and test changes


sescandell:
First of all: what a really good PR this one is :+1: 

Just wondering: aren't these changes BC Breaks?

How do you handle BC Breaks in MvvmCross and Android-Support? I'm totally agree this unification is important. I'm just trying to get information on how I can be prepared to update my apps.

Thanks,

martijn00:
This will indeed break some namespaces, but other then that no real code changes or fixes.

PR #1277, opened 2016-04-01
Added missing constructor

slown1:


SeeD-Seifer:
Shouldn't alternative method be used to obtain styled attributes when defStyle provided? http://developer.android.com/intl/ru/reference/android/content/res/Resources.Theme.html#obtainStyledAttributes(android.util.AttributeSet,%20int%5B%5D,%20int,%20int)

(I've not ever override this constructor, but if it's going to be added, then all the income fields should be used right)

slown1:
I want to extend the MvxImage further and I want to do extra stuff in my subclass. I am porting a class from Java and this was the way the did it. I think that you cand do it both ways.

I am not sure if the init method should be called there.

@martjin what do you say?

slown1:
Indeed I can use the constructor with the context and IAttributeSet only...

 TypedArray styledAttributes = context.ObtainStyledAttributes(attrs, Resource.Styleable.DeventureCircleImageView);

        public MvxRoundedImageView(Context context, IAttributeSet attrs) : base(context, attrs)
        {
            int numStyles = styledAttributes.IndexCount;
            for (var i = 0; i < numStyles; ++i)
            {
                int attributeId = styledAttributes.GetIndex(i);

                if (attributeId == Resource.Styleable.ID)
                {
                    mBorderWidth = styledAttributes.GetDimensionPixelSize(attributeId, DEFAULT_BORDER_WIDTH);
                }
....
           }

          I think it's more elegant with the constructor that has 3 parameters. Maybe someone will actually need to extend that...

martijn00:
I think it is fine to add this. Thanks!

Issue #1136, opened 2015-09-21
Add more async support


martijn00:
I think some parts could support async await to get a better performance. As this could be breaking i would suggest to do this in the 4.0 release.

Some thoughts from the past: https://github.com/MvvmCross/MvvmCross/issues/488

jamie94bc:
I think view models could benefit from exposing `StartAsync` (returning `Task`, not `async void`) but I don't think we'd want to wait for it to complete in [IMvxViewModelLocator](https://github.com/MvvmCross/MvvmCross/blob/4.0/Cirrious/Cirrious.MvvmCross/ViewModels/MvxDefaultViewModelLocator.cs)?

We could keep `Init` synchronous and make `Start` async if we were enforcing a pattern such as:

```c#
private int? _id;

public void Init(NavObject navObject) {
    _id = navObject.Id;
}

public async Task StartAsync() {
    if (_id.HasValue) {
        Item = await Repoistory.GetAsync(_id.Value);
    }
}
```

Or we could make both async, ensuring that `StartAsync` isn't called until `InitAsync` completes.

In terms of perceived performance, we'd probably want to call `StartAsync` after the view model has been assigned to the view?

Were there any other specific areas you had in mind @martijn00?

martijn00:
I was thinking like @Cheesebaron said in Jabbr:

I wonder if we take https://github.com/MvvmCross/MvvmCross/blob/4.0/Cirrious/Cirrious.MvvmCross/Platform/MvxSetup.cs and make everything async in there, if that speeds up the startup process.

I would go for full async with your init proposal.

willsb:
@martijn00 What is the way to benchmark the speed of startup? Will a simple stopwatch suffice? I could give it a shot on this one :)

martijn00:
@willsb  I currently don't have anything to benchmark this with. Maybe you could use something like the Xamarin Profiler?

roubachof:
after digging a while the async/await initialization, it seems to me that we shouldn't give the async responsability to the framework . When doing so (async Task Start() on MvxViewModel), it drips on methods that shouldn't handle async

like this one:

MvxCachingFragmentActivity
=> protected override void OnPostCreate(Bundle savedInstanceState(edited)

I guess it's better to let the dev handle the async loading ​*apart*​ from the mvvmcross framework
saying this, we should make a post in the docs to help the dev loads its data asynchronously.

We could let Stephen Cleary speaks for ourselves:

Async Data Loading
[...]

PR #1284, opened 2016-04-06
Add possibility to pass custom objects with Viewmodel navigation

martijn00:
Inspiration from: http://stackoverflow.com/a/29471996/4222823


benhysell:
That would be nice to have...I've been doing the same thing by hand and it seems to work fine.

SeeD-Seifer:
In short - current implementation is freezy for more then simple objects

In depth. Once upon a time, I had to send non-simple objects between ViewModels. Simple Serialization/Deserialization appeared freeze screen transition a lot on Android devices (for example Nexus 4) and old iOS devices (iPhone 4/4s).

There was few steps on improvement to make object transition smooth.
**Idea:**
1. Store keys to the MvxBundle synchronously.
2. Pass MvxBundle to the ShowViewModel
3. In a new thread serialize objects by keys
4. In the opened ViewModel, await for values by keys

**Concerns:**
Key-value pairs stored to the local DB.
Ensure key-value pairs stored for enough time, to be able reload it when screen restored from tombstone.
Awaiting for object serialisation is not precise for non-nullable types

This is my current implementation which is ugly, but could be used as an example: 
https://gist.github.com/SeeD-Seifer/0df2dfbbe55252fa6d2883b348d0ec68

Didn't expect to share the code, please let me know if it requires more comments around and clean up.

If it appears that this is an overkill and there is more simple solution, I will gladly kill the code and use new approach :)

martijn00:
@SeeD-Seifer i see your concerns, but i think the solution should be more simple to use, like my implementation. Could we somehow mix those two, to make a high performant navigations with parameters?

SeeD-Seifer:
I'm completely agree that solution should be simple, I just didn't find that way yet.

Current implementation cons:
- Serialisation could be heavy
- Only one parameter supported (which means either we can pass one model or create special container per ViewModel transition)

Possible optimisations:
- Make deserialisation asynchronous

martijn00:
The current `IMvxJsonConverter` doesn't have asynchronous support so we shoud add that first i guess.

Cheesebaron:
About async support I think you should read https://github.com/JamesNK/Newtonsoft.Json/issues/66. I think it should be up to the user or the part of the code that requires async deserialization to offload it to a new Task, instead of providing it in MvvmCross.

SeeD-Seifer:
@Cheesebaron good point, thanks. In that case we can only warn developer that he should be careful transferring large objects cos they freeze the app on transition.

benhysell:
Maybe this would be better off as a 'how to' instead of embedding it in the framework because of the risks.  I can already see the issues "I have a giant 5 mb object, why is my screen freezing"

Cheesebaron:
If you are passing around a 5mb object you are already doing it wrong!

willsb:
Awesome proposal, but won't it fall on same the problems mentioned by @roubachof on #1136?

SeeD-Seifer:
How to measure when _dev_ crossed an object limit size? Is 1KB too much? 10KB? 100KB?
If there is a method that could be used wrong, it will be used so.

martijn00:
I've added a description to make the concerns a bit clear to developers. I'll merge this for now, but feel free to open a new PR to improve this. Thanks for all the suggestions and feedback!

PR #1286, opened 2016-04-07
Don't put any special logic in the ACW constructors

### LINKED ISSUE #1285: Investigate whether we should remove all logic from ACW Constructors
After reading [this stackoverflow post](http://stackoverflow.com/questions/10593022/monodroid-error-when-calling-constructor-of-custom-view-twodscrollview/10603714#10603714) it looks like we may want to remove all logic from the Android Callable Wrapper constructors:

```csharp
Constructor(IntPtr handle, Android.Runtime.JniHandleOwnership transfer) : base(handle, transfer){}
```

When one of these classes is created by the Java side of Android the ACW constructor is invoked and then one of the "normal" constructors is invoked.

See: https://developer.xamarin.com/guides/android/under_the_hood/architecture/#Java_Activation

I can not find a case in the xamarin examples where there is logic in the ACW constructor.

kjeremy:
The appropriate constructor gets called after the ACW constructor.
Avoids duplicating work and calling virtual functions until after
the Java side is fully initiated.

Fixes #1285


martijn00:
Thanks!

PR #1290, opened 2016-04-11
Refactor plugin registry

SeeD-Seifer:
**This is a breaking change**

Plugin registry embedded into base MvxPluginManager implementation,
so that MvxFilePluginManager tries to find custom (unconventional) plugin loaders first.

MvxLoaderPluginManager with it's interface deleted.
MvxMacSetup and MvxIosSetup use MvxPluginManager now.
MvxSetup class provide virtual method to access plugin registry.

**TODO**
- [x] Fix MvvmCross-Sample projects _ValueConvertion_ and _WorkingWithCollections_ compilation (because current change break them)
- [x] Update any of MvvmCross-Sample example to demonstrate current feature


Cheesebaron:
Love what is going on here. I've been missing something like this several times.

From what I can see it should still work with how plugins worked previously, the code has just been adjusted slightly.

About the additions to IosSetup: https://github.com/MvvmCross/MvvmCross/pull/1290/commits/ffd2b3390a5121b4030aa38fad2d85cd4fe8b8a4#diff-1f90d540c92c33d192a140d22f7d3402R64 would this also be needed for Android and other platforms?

SeeD-Seifer:
> From what I can see it should still work with how plugins worked previously, the code has just been adjusted slightly.

All previous cases supported. If anybody used ```AddPluginsLoaders``` before, then the code should be simply updated to the new approach.

> About the additions to IosSetup: ffd2b33#diff-1f90d540c92c33d192a140d22f7d3402R64 would this also be needed for Android and other platforms?

Not sure I understand the question.
There is no big change for iOS and Mac, they work as before.
Other platforms use ```MvxFilePluginManager``` which extends ```MvxPluginManager``` and so that it is possible to use approach from iOS/Mac to load plugins in addition to common one.

It is still possible to override ```AddPluginsLoaders``` in Setup.cs to load plugins, but it's just simpler to use ```MvxLoaderPluginBootstrapAction<CoreT, PlatformT>```

SeeD-Seifer:
And there is no more need to use **conventional** naming to register platform plugin.

martijn00:
Thanks! This is looking good!

PR #1294, opened 2016-04-14
Add a way to interact on Content with MvxFrameControl

sescandell:
For now, if you create your custom control based on `MvxFrameControl` on Android, you don't have (or let me know how :) ) any way to write custom code when Content is inflated (except by copy / pasting all the class).

Thanks to this PR, the developper can simply override the `OnContentSet` method and do whatever (s)he wants.


martijn00:
This seems reasonable to add. Thanks!

PR #1295, opened 2016-04-14
add .mailmap file so 'git shortlog' shows sane output

spockfish:
In order to be able to populate an AUTHORS file or something similar it's necessary that we first fix the output of 'git shortlog -s -e'.


martijn00:
Thanks!

PR #1298, opened 2016-04-17
Introduces the IMvxAsyncCommand interface

### LINKED ISSUE #1293: Shouldn't we introduce an `IMvxAsyncCommand` interface?
The commands in my VM are usually of type `IMvxCommand`. However, when testing async commands, I need a way to `await` my code or else I get undefined behaviours. This forces me to make my VM commands `MvxAsyncCommand`s. This, IMO, goes against what's in the [manifesto](https://github.com/MvvmCross/MvvmCross/wiki/The-MvvmCross-Manifesto).

Based on [Stephen Cleary's async command](https://msdn.microsoft.com/en-us/magazine/dn630647.aspx) pattern, shouldn't an `IMvxAsyncCommand` interface be introduced? I'll implement it myself, I just need to know what the community thinks of it beforehand

BONUS QUESTION: Since the `ExecuteAsync` method is [on the Base class](https://github.com/MvvmCross/MvvmCross/blob/3c735adc534a5df2d4730e9d58a08f7863c30cee/MvvmCross/Core/Core/ViewModels/MvxAsyncCommand.cs#L78-L81), there's a problem similar to the one at #1240, where in order to execute a parameterless async command, I need to call `ExecuteAsync(null)`. This could be improved as well, forcing the passing of a parameter only on the `MvxAsyncCommand<T>`

willsb:
Fixes #1293

### DIFF
--- a/MvvmCross/Core/Core/MvvmCross.Core.csproj
+++ b/MvvmCross/Core/Core/MvvmCross.Core.csproj
@@ -43,6 +43,7 @@
     <Compile Include="Parse\StringDictionary\MvxStringDictionaryWriter.cs" />
     <Compile Include="Platform\IMvxFillableStringToTypeParser.cs" />
     <Compile Include="Platform\IMvxStringToTypeParser.cs" />
+    <Compile Include="ViewModels\IMvxAsyncCommand.cs" />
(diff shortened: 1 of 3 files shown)


martijn00:
Thanks!

PR #1306, opened 2016-04-25
New analyzer/codefix for MvxMessengerSubscription return value.

azchohfi:
Total refactor for the analyzers test project to support better test scenarios.


martijn00:
Thanks for helping with this!

PR #1307, opened 2016-04-25
Set image resource directly on the ImageView instead of decoding it first

tiftof:
The current DrawableName binding has a few bad side effects:
* When decoding a large image on a low resolution device, the quality will be bad. I don't know why yet, but it doesn't happen when letting Android decode the image.
* Current implementation decodes the Bitmap => Mono runtime has a reference to the Bitmap => Java GC won't be able to clean up the bitmap unless Mono has already collected this reference => possible memory leak
* There is no automatic reuse of drawables. When setting the same drawable on multiple ImageViews, Android will detect this and decode the drawable only once. When the bitmap is explicitly decoded, we lose this advantage.

### DIFF
--- a/MvvmCross/Binding/Droid/Target/MvxImageViewDrawableNameTargetBinding.cs
+++ b/MvvmCross/Binding/Droid/Target/MvxImageViewDrawableNameTargetBinding.cs
@@ -9,7 +9,6 @@ namespace MvvmCross.Binding.Droid.Target
 {
     using System;
 
-    using Android.Graphics;
     using Android.Widget;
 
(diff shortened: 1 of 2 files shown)


martijn00:
This change looks quite good! I am wondering if there are any side effects or disadvantages to this? And will it break stuff for some users?

sescandell:
Actually, there might be a side effect: `SetImageResource` is [running on the UI Thread](http://developer.android.com/reference/android/widget/ImageView.html#setImageResource%28int%29). If you use it in a list or with a lot of images, you might experience some latency.

It would be better if `SetImageDrawable` is used.

PR #1310, opened 2016-04-27
Fix MvxSpinner with data bound MvxItemTemplate not resizing for item …

royd:
…indexes greater than zero.

Spinner resizes for the first 15 items, but cannot do so accurately when the DataContext is null due to being detached from the window at the time of width calculation.

### DIFF
--- a/MvvmCross/Binding/Droid/Views/MvxBaseListItemView.cs
+++ b/MvvmCross/Binding/Droid/Views/MvxBaseListItemView.cs
@@ -51,12 +51,11 @@ namespace MvvmCross.Binding.Droid.Views
         }
 
         private object _cachedDataContext;
-        private bool _isAttachedToWindow;
 
         protected override void OnAttachedToWindow()
         {
             base.OnAttachedToWindow();
-            this._isAttachedToWindow = true;
+
             if (this._cachedDataContext != null
                 && this.DataContext == null)
             {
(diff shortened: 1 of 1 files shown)


kjeremy:
This looks like a big change.  Is the regular spinner detached during this time as well or just ours?  Also note that you should also update `MvxAppCompatSpinner` in the support project.

royd:
Yes, it is a bit of a fundamental change in terms of the concept of when a list item binding should actually occur. To me it seems like `MvxAdapter.GetView` should return a bound view that can be measured, since that would be possible with `Adapter.GetView`. `Spinner` would normally be attached at this time, as would the first list item since it is the only one displayed. Items two through 15 would not normally be attached until the drop down is opened.

@slodge, I wasn't able to find the particular issue that the [initial caching commit](https://github.com/MvvmCross/MvvmCross/commit/1aa0abbd5042fd06893ded18e19f61ec60efaf04) fixes, but would be curious to find out.

This commit would change the default behavior of `MvxAppCompatSpinner` as well since it uses `MvxAdapter` which uses `MvxListItemView` which derives from `MvxBaseListItemView`.

There are easy work-arounds for this, such as overriding `MvxAdapter.BindBindableView` and setting the `IMvxBindingContextOwner.BindingContext.DataContext`, but it seems like this should work by default.

Example work-around:
```
public class ResizingSpinnerAdapter : MvxAdapter
{
    public ResizingSpinnerAdapter(Context context)
        : base(context)
    {
    }

    protected override void BindBindableView(object source, IMvxListItemView viewToUse)
    {
        base.BindBindableView(source, viewToUse);

        IMvxBindingContext bindingContext = (viewToUse as IMvxBindingContextOwner)?.BindingContext;
        if (bindingContext != null && bindingContext.DataContext == null)
            bindingContext.DataContext = source;
    }
}
```

Cheesebaron:
I would not merge this. The problem is bigger than this and is related to https://github.com/MvvmCross/MvvmCross/issues/1465 We need to change to not wrap ListView items in our own View to fix LayoutParameters for rows. That goes for anything that uses MvxAdapter.

PR #1316, opened 2016-05-02
Use gitlink to inject source references in PDBs back to mvvmcross repo

### LINKED ISSUE #1314: Generate PDBs for source linking with GitLink
[GitLink](https://github.com/GitTools/GitLink) looks like a less error prone way to inject source information into our PDBs.

kjeremy:
This requires the user to have `gitlink.exe` in their `PATH` which is typically installed via `chocolatey`.  See https://github.com/GitTools/GitLink/

Fixes #1314

### DIFF
--- a/nuspec/MvvmCross.AutoView.Droid.nuspec
+++ b/nuspec/MvvmCross.AutoView.Droid.nuspec
@@ -30,9 +30,7 @@ This package contains the 'AutoView' Android libraries for MvvmCross
 	<files>		
 		<!-- droid -->
 		<file src="..\bin\Release\Mvx\Droid\MvvmCross.AutoView.Droid.*" target="lib\MonoAndroid" />
-		<file src="..\MvvmCross\AutoView\Droid\**\*.cs" target="src\MvvmCross.AutoView.Droid" />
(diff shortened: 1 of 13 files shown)


Cheesebaron:
Do we really need to remove source files from the nuspecs? Can't this work without removing it as long as we don't generate symbols?

kjeremy:
I believe we do since gitlink writes the references to the source files.
Maybe it's in the gitlink docs. I can check early in the week.

Cheesebaron:
I've been using GitLink it some of my own Xamarin libraries and it simply does not work. I can't step into anything that generates mdb instead of pdb files.

martijn00:
@kjeremy Do we still want this in?

Cheesebaron:
This is part of the CI task. However, this should be done for all packages, and will simply be a step in the build.

kjeremy:
@Cheesebaron does that mean gitlink works for you again?

Cheesebaron:
Uhm, haven't tried it recently. My hunch is that I shouldn't include mdb files and only pdb files. Will try again soon-ish.

Cheesebaron:
We do this in the new CI builds

https://github.com/MvvmCross/MvvmCross/blob/develop/build.cake#L61

PR #1322, opened 2016-05-11
TaskBasedBindingContext allows smoother scrolling by databinding on a worker threads

softlion:
TaskBasedBindingContext:  datacontext bindings now occurs on a worker thread. Can be replaced by MvxBindingContext in the iOC to get back the previous behavior.

Source binding changes (NotifyPropertyChanged) can now be triggered on any thread. Switching on the main thread is always done before calling the target binding.

### DIFF
--- a/MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs
+++ b/MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs
@@ -1,0 +1,209 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using MvvmCross.Binding.Binders;
+using MvvmCross.Binding.Bindings;
+using MvvmCross.Platform;
+
+namespace MvvmCross.Binding.BindingContext
+{
+    /// <summary>
+    /// OnDataContextChange executes asynchronously on a worker thread
+    /// </summary>
+    public class MvxTaskBasedBindingContext : IMvxBindingContext
+    {
+        private readonly List<Action> _delayedActions = new List<Action>();
+        private readonly List<MvxBindingContext.TargetAndBinding> _directBindings = new List<MvxBindingContext.TargetAndBinding>();
+        private readonly List<KeyValuePair<object, IList<MvxBindingContext.TargetAndBinding>>> _viewBindings = new List<KeyValuePair<object, IList<MvxBindingContext.TargetAndBinding>>>();
+        private object _dataContext;
+        private IMvxBinder _binder;
+
+        public event EventHandler DataContextChanged;
+
+        public IMvxBindingContext Init(object dataContext, object firstBindingKey, IEnumerable<MvxBindingDescription> firstBindingValue)
+        {
+            this.AddDelayedAction(firstBindingKey, firstBindingValue);
+            if (dataContext != null)
+                this.DataContext = dataContext;
+
+            return this;
+        }
+
+        public IMvxBindingContext Init(object dataContext, object firstBindingKey, string firstBindingValue)
+        {
+            this.AddDelayedAction(firstBindingKey, firstBindingValue);
+            if (dataContext != null)
+                this.DataContext = dataContext;
+
+            return this;
+        }
+
+        private void AddDelayedAction(object key, string value)
+        {
+            this._delayedActions.Add(() =>
+            {
+                var bindings = this.Binder.Bind(this.DataContext, key, value);
+                foreach (var b in bindings)
+                    this.RegisterBinding(key, b);
+            });
+        }
+
+        private void AddDelayedAction(object key, IEnumerable<MvxBindingDescription> value)
+        {
+            this._delayedActions.Add(() =>
+            {
+                var bindings = this.Binder.Bind(this.DataContext, key, value);
+                foreach (var b in bindings)
+                    this.RegisterBinding(key, b);
+            });
+        }
+
+        ~MvxTaskBasedBindingContext()
+        {
+            this.Dispose(false);
+        }
+
+        public void Dispose()
+        {
+            this.Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                this.ClearAllBindings();
+            }
+        }
+
+        protected IMvxBinder Binder
+        {
+            get
+            {
+                this._binder = this._binder ?? Mvx.Resolve<IMvxBinder>();
+                return this._binder;
+            }
+        }
+
+        public object DataContext
+        {
+            get { return this._dataContext; }
+            set
+            {
+                this._dataContext = value;
+                this.OnDataContextChange();
+                DataContextChanged?.Invoke(this, EventArgs.Empty);
+            }
+        }
+
+
+        /// <summary>
+        /// Must be called on main thread as it creates the target bindings, and creating target bindings might subscribe to events that
+        /// needs to be done on main thread (like touchupinside).
+        /// </summary>
+        protected virtual void OnDataContextChange()
+        {
+            if (this._delayedActions.Count != 0)
+            {
+                foreach (var action in this._delayedActions)
+                {
+                    action();
+                }
+                this._delayedActions.Clear();
+            }
+
+            Task.Run(() =>
+            {
+                foreach (var binding in this._viewBindings)
+                {
+                    foreach (var bind in binding.Value)
+                    {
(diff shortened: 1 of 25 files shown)


MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs
```diff
@@ -0,0 +1,209 @@
...
+        {
+            this.Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                this.ClearAllBindings();
```
Cheesebaron: Shouldn't all binding tasks also be cancelled here?

MvvmCross/Core/Binding/MvvmCross.Binding.csproj
```diff
@@ -32,6 +32,9 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="System" />
+    <Reference Include="System.Collections.Concurrent">
+      <HintPath>C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETPortable\v4.5\Profile\Profile7\System.Collections.Concurrent.dll</HintPath>
```
Cheesebaron: I think this should be added from NuGet instead.

MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs
softlion: Yes could be. It won't change much though: if a dispose is received while a binding is changing, it may still crash.

MvvmCross/Core/Binding/MvvmCross.Binding.csproj
softlion: Microsoft Concurrent nuget is not available for silverlight apps. And Mvvmcross.Binding it also targeted at silverlight. Should we update the targets to remove silverlight ?

martijn00:
I agree on removing silverlight, and moving forward with this!

Cheesebaron:
Seems to work very well on Android. Created a small app which keeps loading new rows into a ListView with images and text rows. Scrolls smoothly no issues at all.

Will do some more testing on iOS tomorrow.

MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs
kjeremy: Can we protect against that?

MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs
```diff
@@ -0,0 +1,209 @@
...
+                return this._binder;
+            }
+        }
+
+        public object DataContext
+        {
+            get { return this._dataContext; }
+            set
+            {
+                this._dataContext = value;
```
kjeremy: Equality test?

MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs
```diff
@@ -0,0 +1,209 @@
...
+                DataContextChanged?.Invoke(this, EventArgs.Empty);
+            }
+        }
+
+
+        /// <summary>
+        /// Must be called on main thread as it creates the target bindings, and creating target bindings might subscribe to events that
+        /// needs to be done on main thread (like touchupinside).
+        /// </summary>
+        protected virtual void OnDataContextChange()
```
kjeremy: If this function must be called on the main thread can we at least throw or assert if we're on the wrong thread?

MvvmCross/Core/Binding/Bindings/MvxFullBinding.cs
```diff
@@ -94,12 +107,14 @@ private void UpdateTargetOnBind()
         {
             if (this.NeedToUpdateTargetOnBind && this._sourceStep != null)
             {
-                // note that we expect Bind to be called on the UI thread - so no need to use RunOnUIThread here
+                cancelSource.Cancel();
+                cancelSource = new CancellationTokenSource();
```
kjeremy: Shouldn't the old cancelSource be Disposed of right here?

kjeremy:
This is a pretty interesting change.  I'm not too familiar with this area (and this looks like a substantial change).  So some questions:

1. Does every binding spin off a new `Task` or is there one worker thread?
2. Is this entirely thread safe (including cancellation if we need to move to a new view quickly)?
3. Can we use `async`/`await`?  I know that codegen in this area has improved a lot in the later .NET release.
4. You say that `INotifyPropertyChange` now fires from all threads or it's marshalled to the UI thread?  If I recall the default changed so that it was marshalled on the UI thread sometime in the 3.x days of MvvmCross.

MvvmCross/Core/Binding/Bindings/MvxFullBinding.cs
softlion: Well no:  if you dispose it here, the current subscribers of cancelSource will crash (they are async and may be doing other work before retesting the cancelSource.Token variable they have captured using closures).

MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs
softlion: yep

MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs
softlion: where?

MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs
softlion: Well it should not happen, as the view subscriptions to viewmodel changes are weak. 
But it may happen in corner cases if Dispose() is called explicitly on views in a custom presenter.

A good workaround would be to wait for tasks to end, but Dispose is not async so we can not do that.
So code using Dispose in their custom presenter needs to wait for binding tasks to finish before calling Dispose.
This may be documented in the wiki as a change. And we may expose a public Task property to wait for.
Or the code could wait a few seconds before disposing. It should be enough as executing all bindings should always be very fast <20ms.

softlion:
@kjeremy 

> Does every binding spin off a new Task or is there one worker thread?

Each change to a binding value (ie each notifypropertychanged) does not create a new Task. Only a full rebinding (ie assigning DataContext) is done in a task.
But a propertychanged may triggers a thread switch (dispatcher.RequestMainThreadAction). I hope the dispatcher implementation is smart enough to not switch threads if already on main thread.

> Is this entirely thread safe (including cancellation if we need to move to a new view quickly)?

yes

> Can we use async/await? I know that codegen in this area has improved a lot in the later .NET release.

Not currently, the interface signatures and the architecture between mvvmcross components are too heavily non async. The full mvvmcross framework has not been thinked for async at all. Most methods should returns tasks, in its current version i never saw the async keywork anywhere. This would be a very good project which should break all previous mvvmcross apps.
A customer already implemented an async Start() method in its viewmodel. I don't like this approach though as it means the view will not be initialized before appearing.

> You say that INotifyPropertyChange now fires from all threads or it's marshalled to the UI thread? If I recall the default changed so that it was marshalled on the UI thread sometime in the 3.x days of MvvmCross.

There is an exising method in viewmodels that you can use to force all notifypropertychanged calls to switch on the ui thread. This flag should be set to false to benefit from the new behavior of this binder.  The binder will switch to the ui thread when needed, not before it is needed:

            ShouldAlwaysRaiseInpcOnUserInterfaceThread(false);

MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs
kjeremy: As in `if (this._dataContext == value) return;`.  Not a big deal if that's not the pattern.

MvvmCross/Core/Binding/MvvmCross.Binding.csproj
kjeremy: I think it's safe to remove WP 8.0.  I believe ReactiveUI is also dumping it. See https://github.com/reactiveui/ReactiveUI/issues/945

Cheesebaron:
This seems to work fine. I can scroll smoothly with several bindings in a custom cell on both Android and iOS. Will merge this. Thanks @softlion!

PR #1323, opened 2016-05-11
UWP - Fixed 'ViewModel.Close(this)'.

pauloapsantos:
This fixes a issue with Close(this) where it would go back on the root frame instead of the correct frame.


martijn00:
Thanks!

Issue #1308, opened 2016-04-26
IoC doesn't support transparent proxies on Xamarin.Android


yallie:
Hi, I'm using transparent proxies in my MvvmCross application on Android.
Some of my services are in fact transparent proxies doing remote calls:

```c#
Mvx.RegisterSingleton(interfaceType, () =>
{
	var connection = Mvx.Resolve<IConnection>();
	return connection.CreateProxy(interfaceType);
});
```
It worked fine in MvvmCross 3.0.13, but when I updated the project
to MvvmCross 4.1.4, service resolution doesn't work anymore.

On Mono, transparent proxies always report their type as MarshalByRefObject.
As a result, type.IsInstanceOfType(proxy) check always returns false — see
[MvxSimpleIoCContainer.cs, line 448](https://github.com/MvvmCross/MvvmCross/blob/4.0/MvvmCross/Platform/Platform/IoC/MvxSimpleIoCContainer.cs#L448). And so IoC reports an error:

>Resolver returned object type MarshalByRefObject which does not support interface IMyService.

I need a way to override or to bypass this check for MarshalByRefObjects.
As far as I know, the only way to check whether a Mono's transparent proxy
implements an interface is like this:

```c#
var typedProxy = transparentProxy as IMyService;
if (typedProxy == null)
    throw new MvxException("Resolver returned invalid service"...);
```

Mono runtime obviously has more information on the proxy that Reflection provides,
so "as" operator works fine, but [type.IsAssignableFrom(proxy.GetType())](https://github.com/MvvmCross/MvvmCross/blob/4.0/CrossUI/CrossUI.Core/ReflectionExtensions.cs#L44) fails.

### Steps to reproduce
1. Clone and build my sample solution: [MvvmCrossIocProxy](https://github.com/yallie/MvvmCrossIocTransparentProxy)
2. Run WindowsConsoleApp on Windows. Service resolution works fine.
3. Run AndroidApp on Android. Same service resolution code fails.

### Expected behavior
IoC should return a singleton object registered for the given interface, be it a simple class or a transparent proxy created at runtime.

### Actual behavior
IoC throws an MvxException: Resolver returned object type MarshalByRefObject...

### Configuration

**Version:** 4.1.4
**Platform:** Android

thefex:
My personal opinion: MvvmCross IoC is good for simple things. How about changing your IoC provider to for example Ninject?

yallie:
[...]

PR #1325, opened 2016-05-15
Ignore transparent proxies on Mono, #1308.

yallie:
On Mono, there is no easy way to check if a transparent proxy supports a given type.
So I just ignore transparent proxies (they are hardly ever used nowadays anyway).


martijn00:
Thanks!

Issue #1222, opened 2016-01-19
Support for generic MvxAdapter<T>


joashmusyoka:
I am converting a listview that I had written earlier but my adapter is generic and so unable to extend MvxAdapter since it does not allow a type parameter.

Thanks.

joashmusyoka:
I guess I should just create a new Adapater<T> and implements IMvxAdapter

martijn00:
Maybe you could do something like the generics we have for MvxActivity: https://github.com/MvvmCross/MvvmCross/blob/4.0/MvvmCross/Droid/Droid/Views/MvxActivity.cs#L83

If you could add that to the current code and make a PR we would be happy to pull it in.

joashmusyoka:
I will do something about it. Thanks.

On Thu, Feb 18, 2016 at 1:15 PM, Martijn van Dijk <notifications@github.com>
wrote:

> Maybe you could do something like the generics we have for MvxActivity:
> https://github.com/MvvmCross/MvvmCross/blob/4.0/MvvmCross/Droid/Droid/Views/MvxActivity.cs#L83
>
> If you could add that to the current code and make a PR we would be happy
> to pull it in.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/MvvmCross/MvvmCross/issues/1222#issuecomment-185638997>
> .
>



-- 
Kind Regards,
Musyoka.

PR #1332, opened 2016-05-20
Generic MvxAdapter implementation

MarcBruins:
Made a generic mvxadapter as suggested in: https://github.com/MvvmCross/MvvmCross/issues/1222

### DIFF
--- a/MvvmCross/Binding/Droid/Views/MvxAdapter.cs
+++ b/MvvmCross/Binding/Droid/Views/MvxAdapter.cs
@@ -309,5 +309,30 @@ namespace MvvmCross.Binding.Droid.Views
         }
     }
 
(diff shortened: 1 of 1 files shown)


MvvmCross/Binding/Droid/Views/MvxAdapter.cs
```diff
@@ -307,4 +308,31 @@ protected virtual IMvxListItemView CreateBindableView(object dataContext, int te
...
+
+		public MvxAdapter(Context context, IMvxAndroidBindingContext bindingContext) : base(context, bindingContext)
+		{
+		}
+
+		public MvxAdapter(IntPtr javaReference, JniHandleOwnership transfer) : base(javaReference, transfer)
+		{
+		}
+
+		public new IList<TItem> ItemsSource
```
kjeremy: Shouldn't this be `IEnumerable<TItem> ItemsSource`?  The base class returns `IEnumerable` and might not be able to cast to `IList<TItem>>`.

MarcBruins:
@kjeremy you are absolutely right, i changed it to IEnumerable

martijn00:
Thanks!

PR #1333, opened 2016-05-23
Animate UICollectionView changes

softlion:
This is a working version that animates insertions and deletions.

To go further, the observablecollection architecture must be changed to support async and batch operations correctly in the view: an extended collection class have to be created for that, containing a method like PerformBatchUpdates(), and allowing concurrent calls.

### DIFF
--- a/MvvmCross/Binding/iOS/Views/MvxCollectionViewSource.cs
+++ b/MvvmCross/Binding/iOS/Views/MvxCollectionViewSource.cs
@@ -10,6 +10,7 @@ namespace MvvmCross.Binding.iOS.Views
     using System;
     using System.Collections;
     using System.Collections.Specialized;
+    using System.Threading.Tasks;
 
     using Foundation;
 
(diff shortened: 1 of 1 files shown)


Cheesebaron:
Thanks for this PR. Going to review it soon along with the other PR you have open 👍

MvvmCross/Binding/iOS/Views/MvxCollectionViewSource.cs
```diff
@@ -80,14 +81,45 @@ protected override object GetItemAt(NSIndexPath indexPath)
...
+                CollectionView.PerformBatchUpdates(() =>
+                {
+                    var startIndex = args.NewStartingIndex;
+                    var indexes = new NSIndexPath[args.NewItems.Count];
+                    for (var i = 0; i < indexes.Length; i++)
+                        indexes[i] = NSIndexPath.FromRowSection(startIndex + i, 0);
+                    CollectionView.InsertItems(indexes);
+                }, ok => { });
+            }
+            else
```
Cheesebaron: What about MoveItem?

softlion: Already supported or not needed

Benjamin

PR #1336, opened 2016-05-25
Cache Java.Lang.Boolean.True since in Xamarin it results in a call

kjeremy:


martijn00:
Do you have some more info on this, or why this is a problem?

kjeremy:
It's just a micro opt.  Everytime you call `Java.Lang.Boolean.True` it reaches into the JNI layer and that seemed stupid for something that is static and const so I decided to cache it.

martijn00:
Thanks for the explanation and PR!

vvolkgang:
@kjeremy alternative explanation: https://youtu.be/9FnO3igOkOk?t=41s

PR #1337, opened 2016-05-26
Allows binding to EditTextPreference.Text and TwoStatePreference.Chec…

kjeremy:
…ked as

well as one-way binding to any other Preference subclass using the generic
"Value" binding.

```
public class SettingsViewFragment : MvxPreferenceFragment<SettingsViewModel>
{
    public override View OnCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState)
    {
        // Since the view is created in PreferenceFragment's OnCreateView
        // we don't use BindingInflate like a typical MvxFragment.
        var view = base.OnCreateView(inflater, container, savedInstanceState);

        var serverPreference = (EditTextPreference)this.FindPreference("pref_eugs_server");

        var bindingSet = this.CreateBindingSet<SettingsViewFragment, SettingsViewModel>();
        bindingSet.Bind(serverPreference)
            .For(v => v.Text)
            .To(vm => vm.EugsServerAddress);
        bindingSet.Apply();

        return view;
    }

    public override void OnCreatePreferences(Bundle savedInstanceState, string rootKey)
    {
        this.AddPreferencesFromResource(Resource.Xml.preferences);
    }
}
```

### DIFF
--- a/MvvmCross/Binding/Droid/MvvmCross.Binding.Droid.csproj
+++ b/MvvmCross/Binding/Droid/MvvmCross.Binding.Droid.csproj
@@ -81,7 +81,9 @@
     <Compile Include="ResourceHelpers\MvxAndroidBindingResource.cs" />
     <Compile Include="ResourceHelpers\MvxAppResourceTypeFinder.cs" />
     <Compile Include="Resources\Resource.Designer.cs" />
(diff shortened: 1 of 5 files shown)


Cheesebaron:
Looks pretty good to me.

PR #1339, opened 2016-05-29
MultiRegionViewPresenter - Support frame search failure.

ErezG:
Upon search failure (no matching frame found), use the default frame.

It allows flexability in building a responsive app that navigates to a frame on big screens and navigates inside the main frame if the screen is small.


martijn00:
Please use the same formatting so it is easier to see what changed.

ErezG:
Martin,
So what now? is there a problem with the PR?

martijn00:
It seems good to me now, but i haven't used this functionality yet. Maybe @Stephanvs, @ErezG or @pauloapsantos knows?

ErezG:
I've used it and it works. This change comes from the adjustment I had to do in my project.

PR #1345, opened 2016-06-03
Unsubscribe events

Cheesebaron:
Loads of views were subscribing events without unsubscribing. Let's be better citizens and actually do that to help minimizing memory issues.

This also marks TabActivity obsolete, as it has been obsolete in many API versions of Android.

### DIFF
--- a/MvvmCross/Binding/Droid/Views/MvxSpinner.cs
+++ b/MvvmCross/Binding/Droid/Views/MvxSpinner.cs
@@ -114,7 +114,7 @@ namespace MvvmCross.Binding.Droid.Views
 
             if (disposing)
             {
-                ItemSelected += OnItemSelected;
+                ItemSelected -= OnItemSelected;
             }
         }
     }
--- a/MvvmCross/Binding/Droid/Views/MvxSpinner.cs
+++ b/MvvmCross/Binding/Droid/Views/MvxSpinner.cs
@@ -38,8 +38,8 @@ namespace MvvmCross.Binding.Droid.Views
             var dropDownItemTemplateId = MvxAttributeHelpers.ReadDropDownListItemTemplateId(context, attrs);
             adapter.ItemTemplateId = itemTemplateId;
             adapter.DropDownItemTemplateId = dropDownItemTemplateId;
-            this.Adapter = adapter;
-            this.SetupHandleItemSelected();
+            Adapter = adapter;
+            ItemSelected += OnItemSelected;
         }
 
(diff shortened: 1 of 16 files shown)


kjeremy:
d48a93e seems unrelated.  Can you make a separate PR for it?

Cheesebaron:
It is in a separate PR, however I think I derped branching from the other branch. Not going to do anything about it though...

MvvmCross/Binding/Droid/Views/MvxSpinner.cs
```diff
@@ -73,42 +73,49 @@ public new IMvxAdapter Adapter
...
+            HandleItemSelected.Execute(item);
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            base.Dispose(disposing);
+
+            if (disposing)
+            {
+                ItemSelected += OnItemSelected;
```
kjeremy: Did you mean `ItemSelected -= OnItemSelected`?

Cheesebaron: Thanks, will fix this

MvvmCross/Binding/Droid/Views/MvxTableLayout.cs
```diff
@@ -98,14 +98,27 @@ protected set
...
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            base.Dispose(disposing);
+
+            if (disposing)
+            {
+                if (_adapter != null)
+                    _adapter.DataSetChanged += AdapterOnDataSetChanged;
```
kjeremy: Did you mean `-= AdapterOnDataSetChanged`?

MvvmCross/Binding/Droid/Views/MvxTableLayout.cs
```diff
@@ -98,14 +98,27 @@ protected set
...
+
+                ChildViewRemoved += OnChildViewRemoved;
```
kjeremy: Same here?

kjeremy:
This is crazy.  Were we actually leaking all of these views?  I wonder if we should be weak subscribing in all of our bindings.

Cheesebaron:
WeakSubscribing would probably work. But yeah these were leaking...

Cheesebaron:
I think this is all of them. I've been looking through most of the code to see if I could spot more.

But yeah, we should look into weaksubscribing all events instead, this way we are more protected against leaks.

PR #1349, opened 2016-06-06
Use drawable instead of resource, which is not UI thread bound

Cheesebaron:
Improves #1307 not forcing images to be set on UI thread.

### DIFF
--- a/MvvmCross/Binding/Droid/Target/MvxImageViewDrawableTargetBinding.cs
+++ b/MvvmCross/Binding/Droid/Target/MvxImageViewDrawableTargetBinding.cs
@@ -46,12 +46,12 @@ namespace MvvmCross.Binding.Droid.Target
                 imageView.SetImageDrawable(null);
             else
             {
-				var appContext = AndroidGlobals.ApplicationContext;
+                var context = imageView.Context;
 				Drawable drawable;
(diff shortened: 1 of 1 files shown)


MvvmCross/Binding/Droid/Target/MvxImageViewDrawableTargetBinding.cs
```diff
@@ -45,7 +46,15 @@ protected override void SetValueImpl(object target, object value)
                 imageView.SetImageDrawable(null);
             else
             {
-                imageView.SetImageResource(intValue);
+				var appContext = AndroidGlobals.ApplicationContext;
+				Drawable drawable;
+				if (Build.VERSION.SdkInt >= BuildVersionCodes.Lollipop)
+					drawable = appContext.Resources.GetDrawable(intValue, appContext.Theme);
```
kjeremy: Shouldn't this use the `ImageView`'s context?  The app context could have a different theme than the activity's.

Cheesebaron:
@kjeremy you are right. There can be a difference in themes here. I've fixed that :)

IlSocio:
Is it still possible to use a Selector with the SetImageDrawable?

Issue #1309, opened 2016-04-27
MvxWindowsViewPresenter swallowing exceptions thrown in Init


shemseddine:
In the viewmodel when throwing an unhandled exception in the Init() function, it is swallowed by the ViewPresenter. Which means I'm unable to handle any of the exceptions globally using my own exception handling methods.
### Configuration

**Version:** 3.5.1

**Platform:** WPF

thefex:
Update MvvmCross to most recent 4.x and check does problem still exist.

nmilcoff:
Can this issue be closed now that we have a brand new ViewPresenter for WPF? (#2124)

Issue #1006, opened 2015-05-14
Throwing exception in Init() method isn't reported


SimonSimCity:
When I call an exception in an `Init()` method of a ViewModel, it seems like that exception never is reported anywhere ... Or am I just missing a subscription to an MvvmCross internal exception handler?

slodge:
Linked to https://github.com/MvvmCross/MvvmCross/issues/837 ? There was even a test in the pull...https://github.com/MvvmCross/MvvmCross/pull/838/files

SimonSimCity:
@slodge I tried it in my codebase and all exceptions are swallowed ... Is there a good way somehow to test where that exception was swallowed? Would be happy to have more to debug ... but just having the project as DLL makes it hard :wink:.

Can I just load this as project, remove the DLL, loaded in that project and take this as project-dependency? Will try that soon ...

martijn00:
Linked to https://github.com/MvvmCross/MvvmCross/issues/1309

SimonSimCity:
Ok, I reserved some time today to get deeper into this one.

As written in the docs (https://github.com/MvvmCross/MvvmCross/wiki/View-Model-Lifecycle#2-init), the `Init()` method is part of the lifecycle of a ViewModel and is called in https://github.com/MvvmCross/MvvmCross/blob/3c735adc534a5df2d4730e9d58a08f7863c30cee/MvvmCross/Core/Core/ViewModels/MvxDefaultViewModelLocator.cs#L48 that again calls my method here: https://github.com/MvvmCross/MvvmCross/blob/3c735adc534a5df2d4730e9d58a08f7863c30cee/MvvmCross/Core/Core/ViewModels/MvxViewModelExtensions.cs#L55

But at the end, all comes down to that I had a message-signature, that allowed me to return a `Task` and use the method as an _async_ method.

Of that, I finally learned how bad it is to either have a method, that returns `void` in an async method or how equally bad it is of me to call an async method but neither awaiting it nor doing something with the returned `Task`.

Here's a very short, but very good blog-post on it: http://www.axilis.com/2015/07/27/async-exception-handling-in-c/

**What we should do:**

Check if the response, given by the invoked method, is awaitable (f.e. by casting to Task), start logging the error-messages, or even better: inform the programmer by a warning, that those methods shouldn't return anything awaitable, but handle all possible exceptions within themselves and return void.

@martijn00 Am I missing something out here, or would that solve the problem - or at least inform the developer about it?

[...]

PR #1355, opened 2016-06-09
throw exception

DekkkersR:
Suggested solution for issue #1309


martijn00:
As pointed out here https://github.com/MvvmCross/MvvmCross/issues/1006#issuecomment-231084325 it should be up to the developer to correctly handle this. We should think about a more permanent solution then this. Thanks for the effort!

PR #1357, opened 2016-06-10
Move away from MvxAndroidTargetBinding and use MvxConvertingTargetBinding

Cheesebaron:
MvxAndroidTargetBinding is really unnecessary to use in most cases.

Also removed `this.` usage from all the target bindings (why did you introduce so much verboseness @martijn00?)


kjeremy:
LGTM (so many formatting changes!) as discussed on slack.

PR #1361, opened 2016-06-10
Withevent propertyinfo targetbindings

Cheesebaron:
Don't merge until the bindings have been verified working!


Cheesebaron:
This wont work. The `MvxWithEventPropertyInfoTargetBinding` doesn't know what type the EventHandler is and we can't have a single `OnValueChanged` method that covers all EventHandlers. Closing this.

Issue #1245, opened 2016-02-22
Issue when application wakes after a while on Android


chamamo:
Hi,

I have an Exception when my application is restored 

`System.NullReferenceExceptionObject reference not set to an instance of an object
Raw
MvvmCross.Platform.Mvx.Resolve<TService>()
MvvmCross.Binding.Droid.BindingContext.MvxAndroidBindingContextHelpers.Current<T>()
MvvmCross.Binding.Droid.BindingContext.MvxAndroidBindingContextHelpers.Current()
MvvmCross.Binding.Droid.Views.MvxLayoutInflater.Inflate(int resource, ViewGroup root, bool attachToRoot)
Android.Views.LayoutInflater.n_Inflate_ILandroid_view_ViewGroup_Z(IntPtr jnienv, IntPtr native__this, int resource, IntPtr native_root, bool attachToRoot)
at (wrapper dynamic-method) System.Object:fefe575a-38ac-4503-8078-67f0289580b1 (intptr,intptr,int,intptr,bool)`

Any idea please?

Cheesebaron:
Is this the complete stack trace?

chamamo:
Yes

chamamo:
When It happened the debugger is detached

Cheesebaron:
There is nothing much to chase from that trace. If you can give us more specific reproduction steps we are more than happy to help.

JelleDamen:
I get this same error after closing the app and then opening the via a push notification.

`dalvikvm( 2566): threadid=1: thread exiting with uncaught exception (group=0x41619468)
E/AndroidRuntime( 2566): FATAL EXCEPTION: main
E/AndroidRuntime( 2566): java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
E/AndroidRuntime( 2566): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:796)
E/AndroidRuntime( 2566): 	at dalvik.system.NativeStart.main(Native Method)
E/AndroidRuntime( 2566): Caused by: java.lang.reflect.InvocationTargetException
E/AndroidRuntime( 2566): 	at java.lang.reflect.Method.invokeNative(Native Method)
E/AndroidRuntime( 2566): 	at java.lang.reflect.Method.invoke(Method.java:511)
E/AndroidRuntime( 2566): 	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1029)
[...]

Issue #1192, opened 2015-12-15
[Android] Setup initialization issue when app killed and resumed


cyrilcathala:
Hi,

I found an issue on Android apps when an app is killed and then resumed.

**The problem:**
Usually, the MvxSetup is initialized during the splashscreen on Android.
If an app is killed by the OS and you try to resume the app, the splashscreen doesn't launch, so the MvxSetup is initialized during the `Activity.OnCreate()` through the [MvxActivityViewExtensions.OnViewCreate](https://github.com/MvvmCross/MvvmCross/blob/71265521e1b57d1e01cca514dfd434de07660440/Cirrious/Cirrious.MvvmCross.Droid/Views/MvxActivityViewExtensions.cs#L42).
Well, the problem is, when you register fragments BEFORE the `Activity.OnCreate()`, the setup isn't yet initialized and it crashes because Mvx (the IoC container) is null.

**How to reproduce:**
The issue can be reproduced with the XPlatformMenus sample, the MainActivity registers fragments before the `base.OnCreate()` call (check [MainActivity.OnCreate](https://github.com/MvvmCross/MvvmCross-Samples/blob/master/Samples/XPlatformMenus/XPlatformMenus.Droid/Activities/MainActivity.cs)).
Go to Dev Settings, check the "Don't keep activities" and set Background process limit to "No background processes". Then you just have to launch the sample once, go to the launch screen, then go back to the app.

**The solution?**
Here is a workaround, ensuring the setup is initialized before registering fragments:

```
public abstract class BaseFragmentActivity : MvxCachingFragmentActivity
{
[...]
        protected override void OnCreate(Bundle bundle)
        {
            var setup = MvxAndroidSetupSingleton.EnsureSingletonAvailable(ApplicationContext);
            setup.EnsureInitialized();

            RegisterFragments();
            base.OnCreate(bundle);
        }

        protected abstract void RegisterFragments();
}
```

This is not perfect because when the app restarts, it freezes during the setup initialization... Do you think of a better solution?

Please note that it seems to be the same problem as issue #1112.

martijn00:
The registration will change quite a bit next release: https://github.com/MvvmCross/MvvmCross-AndroidSupport/pull/149

This might be affected by the change. Can you check?

andyci:
[...]

PR #1370, opened 2016-06-18
[WIP] Speed up start process

Cheesebaron:
Don't pull yet...

Related to #955, #1192, #1245, #1253 and probably more.

Will probably break stuff when eventually merged.

### DIFF
--- a/MvvmCross/Core/Core/MvvmCross.Core.csproj
+++ b/MvvmCross/Core/Core/MvvmCross.Core.csproj
@@ -121,6 +121,7 @@
     <Compile Include="Views\MvxViewsContainer.cs" />
     <Compile Include="ViewModels\IMvxPageViewModel.cs" />
     <Compile Include="ViewModels\MvxObservableCollection.cs" />
+    <Compile Include="Platform\MvxAsyncSetup.cs" />
   </ItemGroup>
   <ItemGroup />
   <Import Project="$(MSBuildExtensionsPath32)\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.CSharp.targets" />
--- a/MvvmCross/Core/Core/Platform/MvxAsyncSetup.cs
+++ b/MvvmCross/Core/Core/Platform/MvxAsyncSetup.cs
@@ -1,0 +1,422 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using MvvmCross.Core.Platform;
+using MvvmCross.Core.ViewModels;
+using MvvmCross.Core.Views;
+using MvvmCross.Platform;
+using MvvmCross.Platform.Core;
+using MvvmCross.Platform.Exceptions;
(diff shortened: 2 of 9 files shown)


Dexyon:
Nice 👍

Cheesebaron:
Now, making stuff async introduces some problems. Lets say we await the AndroidAsyncSetupSingleton in `OnCreate` and mark `OnCreate` `async`. This would introduce the problem that `OnResume` could potentially be called before `OnCreate` ends executing.

So I am thinking whether we should execute it sync in `OnCreate`, or does someone have a better idea what to do instead? Wrapping it in `Task.Run` wouldn't bode well, because then code can execute while system is not ready to resolve ViewModels etc.

andyci:
I started changing the startup process to async a while ago and came across the same issue as this.  I don't think there is any other choice than to execute it synchronously because otherwise services such as the singleton cache may not be fully initialised when the viewmodels are resolved, etc.

Whilst we can make the individual methods of the setup asynchronous, I suggest we keep the InitializeAsync method synchronous as strange as that might sound.  There isn't much point in letting the user initialise the framework asynchronously and then them getting crashes when they try to use the singleton cache before it is ready

nmilcoff:
Based on this article from Stephen Cleary https://msdn.microsoft.com/en-us/magazine/jj991977.aspx ("Configure context" subtitle, when it references figure 3)... What about setting up `OnCreate` this way:

```
private async Task DelayAsync()
{
    await Task.Delay(1000);
}

public void LifecycleMethod()
{
    // Start an async task.
    var delayTask = DelayAsync().ConfigureAwait(false);
    // Wait for it in a deadlock safe way
    delayTask.Wait();
}
```

Note that because of `ConfigureAwait(false)` before the Wait() line, it won't cause a deadlock.
Of course, this way `OnCreate` will run sync.

nmilcoff:
If we keep a dependency to a SplashScreen and more precisely, to a method that acts as a callback like `InitializationComplete`, I think this is something possible to introduce. We could use a sandbox like this one: 

```c#
private static async void RunInSandbox(Func<Task> myTask)
{
  try
  {
    await Task.Yield();
    await myTask.Invoke();
  }
  catch (Exception ex)
  {
    ...log the exception, maybe close the app
  }
}
```
that would allow us to call `AndroidAsyncSetupSingleton` from a sync context. The call would be something like:

`TaskSandbox.RunInSandBox(() => AndroidAsyncSetupSingleton.InitializeAsync(this));`

And then just call `InitializationComplete` when the setup finishes.

nickrandolph:
@Cheesebaron do we have a feel for where this PR is going? Given the history, we might want to open an issue linking to this PR and close it for the moment. There's some nasty pieces to the Android init (reflection searching for Setup for example) that would need an overhaul.

Cheesebaron:
Yeah, this PR is awfully out of date. I haven't looked at this since the last comment. It ties nicely into the two PRs #2593 #2587.

nickrandolph:
Shall we just abandon this PR and close it off then?

PR #1377, opened 2016-06-24
Adapter tweaks

kjeremy:
MvxSpinner should use the new adapter's DropDownItemTemplateId.  Unassigns old adapter ItemsSource so we don't get multiple change events.

### DIFF
--- a/MvvmCross/Binding/Droid/Views/MvxSpinner.cs
+++ b/MvvmCross/Binding/Droid/Views/MvxSpinner.cs
@@ -62,12 +62,11 @@ namespace MvvmCross.Binding.Droid.Views
                     value.ItemTemplateId = existing.ItemTemplateId;
                     value.DropDownItemTemplateId = existing.DropDownItemTemplateId;
                 }
-                if (existing != null)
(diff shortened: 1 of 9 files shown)


MvvmCross/Binding/Droid/Views/MvxSpinner.cs
```diff
@@ -62,12 +62,11 @@ public new IMvxAdapter Adapter
...
                 }
-                if (existing != null)
-                {
-                    existing.ItemsSource = null;
-                }
 
                 base.Adapter = value;
+
+                if (existing != null)
+                    existing.ItemsSource = null;
```
SeeD-Seifer: How about to use C# 6 constructions:
```existing?.ItemsSource = null```

kjeremy: I wish.  But it doesn't work with assignments.

SeeD-Seifer: You are right. My fault.

PR #1383, opened 2016-07-02
new animation enabled collection source, revert legacy collection source code

softlion:
I reverted to the legacy code of collection view source,
so existing app will continue to work with no change and no hassle.

I created a new animation enabled collection source, 
to be used with care if you use the stock ObservableCollection,
as 2 quick operations that animate on the collection may crash your ios app 
with a "invalid state" like error.


martijn00:
Sounds like a good plan to do this! Thanks!

PR #1384, opened 2016-07-02
When using MvxTaskBasedBindingContext (default) with full link enabled.

softlion:
When using MvxTaskBasedBindingContext (default) with full link enabled,
this code is required to prevent xamarin from optimizing it away,
as it mvvmcross code uses only an interface.


martijn00:
Thanks! I believe there was an issue for this is well?

softlion:
dont know :)

PR #1397, opened 2016-07-14
Fix line endings

kjeremy:
Attempt to force line endings to LF in the repo.  Also uses csharp diff driver for *.cs files.


kjeremy:
This should be good to go.

PR #1404, opened 2016-07-21
Added support for view model instantiation using segue on iOS

chrilith:
Hi there,

MVVMCross is real cool but there were a small piece of code missing to handle segue on iOS using storyboards.

I added `MvxSegueExtensionMethods` in order to handle view model instantiation:

- `GetViewModelType()` has been added to `IMvxView` to retrieve the view model type at runtime
- `InstantiateViewModelForSegue()` has been added to `IMvxEventSourceViewController` to handle view model instantiation. It would be nice to have it on `UIViewController` but we cannot modifiy the class directly and we would have expose the method publicly instead of the current `internal`.

All the `Mvx_ViewController` classes have been modifier to override `PrepareForSegue()` and call `InstantiateViewModelForSegue()`.

We should check that this doesn't interfer  with the hack for TabBar in `LoadViewModel()`.

A sample project showing this in action is available here:
https://www.dropbox.com/s/8gmtm7jkjgmxo0n/MVVMCross_SeguePoC.zip?dl=0

Let me know if you have any question!
Thanks!

Chris

### DIFF
--- a/MvvmCross/iOS/iOS/Views/MvxTabBarViewController.cs
+++ b/MvvmCross/iOS/iOS/Views/MvxTabBarViewController.cs
@@ -73,7 +73,7 @@ namespace MvvmCross.iOS.Views
         public override void PrepareForSegue(UIStoryboardSegue segue, NSObject sender)
         {
             base.PrepareForSegue(segue, sender);
-            this.InstantiateViewModelForSegue(segue);
+            this.ViewModelRequestForSegue(segue);
         }
     }
 }
--- a/MvvmCross/iOS/iOS/Views/MvxTabBarViewController.cs
+++ b/MvvmCross/iOS/iOS/Views/MvxTabBarViewController.cs
@@ -17,7 +17,6 @@ namespace MvvmCross.iOS.Views
     public class MvxTabBarViewController
         : MvxEventSourceTabBarController
           , IMvxIosView
-          , IMvxIosViewSegue
     {
         protected MvxTabBarViewController()
         {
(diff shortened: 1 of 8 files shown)


chrilith:
Hum... I kept the initial method name `InstantiateViewModelForSegue`, very bad name as it doesn't instantiate the view model here. Should be `PrepareRequestForSegue`. Will change this tomorrow.

MvvmCross/iOS/iOS/Views/MvxTabBarViewController.cs
```diff
@@ -47,6 +48,13 @@ public IMvxViewModel ViewModel
         public MvxViewModelRequest Request { get; set; }
 
         public IMvxBindingContext BindingContext { get; set; }
+
+        // CHECKME: won't this interfer with the hack for TabBar in LoadViewModel()
```
martijn00: Can you verify this? After that i think we can merge this!

chrilith: Yes, in my TODO list, sorry, quite busy. :(

martijn00:
Thanks! Let's try to release this and see how it works out.

peterpleva:
Great feature, cannot wait to try out! When do you release it?

ipalfy:
Can anyone please add some (minimal) documentation or an example how to use this?

chrilith:
@ipalfy Basically, all the Mvx-based controller will now internally call the `ViewModelRequestForSegue(segue, sender)` method extension.

This method, will call the `PrepareViewModelParametersForSegue(segue, sender)` extension method which is the one you CAN implement in your own view controller (only) if you want to pass some parameter to the target Mvx view controller. In that case, your controller must implement `IMvxIosViewSegue` which defines this last method.

That's it.

Issue #1403, opened 2016-07-21
Seems there is no way to refresh the MvxPageViewController with new data


chrilith:
Hi @scastria,

This controller is really nice but I didn't found a way to refresh data or invalidate the cache dictionary.

Let's say you have a list of page data which is empty at startup, then you load REST data and update your list. The page view should update accordingly, which mean clear the current page view state and use the new data from the list.

Looking at the code, I don't see ho to perform this.

Is it possible or is this an issue?

Thanks!

scastria:
If I remember correctly, I did not code in that feature.  I created this out of my need which did not require changing the pages during runtime.  sorry.

chrilith:
OK but can you consider adding this feature?

Also the caching system is interesting but should not be the default behavior. Imagine if you have 50 images in a page view, you should really not want to cache everything. Caching should be up to the developer and not forced to a specific implementation.

I think this should be as generic as possible. What is your thought on this?

Thanks!
Chris

scastria:
I agree.  Unfortunately, I just do not have the time to do it.  I would be happy to review a pull request if you want to make the change, though.

chrilith:
OK np. Will have some questions to fully understand your code. First, what is the initial purpose of `NavigateToViewModel()` and is there any reason why it id not used in `InitializePaging()` using `defaultVM`?

scastria:
I think my intent of NavigateToViewModel was to allow the caller to specify a ViewModel to navigate to instead of specifying a ViewController.  Maybe if you had a Command in some VM somewhere that wanted to switch current pages, it could do so by specifying another VM since a VM should not know about the View.  However, since the NavigateToViewModel is on the View (MvxPageViewController), maybe my intent doesn't makes sense.

No reason why I didn't use it in InitializePaging on the defaultVM.  I could have.  I guess I missed it.

chrilith:
OK @scastria thanks for clarifying things. I haven't tested this method but I really think it make sense in the logic of your implementation.

Thanks again.

nickrandolph:
@chrilith has this issue been resolved?

chrilith:
Hi @nickrandolph,

Well since #1675 hasn't been merge I guess it is no,t unless there is a new implementation in 5.x or current 6.x.

I must admit that I currently haven't time to check all this.
I'll be able to check everyhing only after we plan to migrate to latest version of MvvmCross which may not arise before the end of this year :(

PR #1407, opened 2016-07-25
New MvxPageViewController

chrilith:
Hi,

I've just commited a new implementation of the MvxPageViewController which works more like any other controllers in MvvmCross.

The current implementation, based on view models, is really interesting, but one thing that is hardly missing is binding (AFAICT) and a way to refresh the controller. Also, as stated in #1403 we are forced to use an internal cache, when iOS already cache things on its own.

Here, I added a simple data source that implements most of the minimal requirements to  properly handle PageViews.

I will add some events later if this PR is acceptable to the dev team. Don't want to waste too much time right now because i have an app to develop :)

_Added_:
- `MvxPageViewSource`: the important part here is `GetViewControllerAtIndex()` which will provide the expected view conroller if any and `GetPageIndexForController()`.
- `IMvxpageViewModel` and `IMvxPageViewController`. Both define à `PageIndex` that will be used by the default data source (`MvxPageViewSource.GetPageIndexForController()`)

Basically everything works with ViewControllers with iOS page views so I think it is important to keep this logic.

_Moved_:
- Old code has been moved to a Legacy namespace to keep compatibility. I can add a new source that mimic most of the old behavior if needed.

Let me know if you have any questions.
Chris


Cheesebaron:
Can you please merge master into this branch. Will have a look when you do that 👍

chrilith:
OK, will try to do it this week!

MvvmCross/Binding/iOS/Views/MvxBasePageViewSource.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        protected UIPageViewController PageView => this._pageView;
+
+        public virtual void ReloadData()
+        {
+            try
+            {
+                var controller = GetInitialViewController();
+                var list = new UIViewController[] { controller };
+                // FIXME: we are forced here to have animated:false because of a bug in iOS up to 9.x
```
martijn00: Maybe do a check for the iOS version and put that in a bool variable. Then use that to animate on iOS 9 and higher?

martijn00:
I wouldn't leave the legacy code in. We can remove that if this PR gets accepted and tested, and release this as a breaking change for MvvmCross 5.0 (https://github.com/MvvmCross/MvvmCross/issues/1415)

MvvmCross/Binding/iOS/Views/MvxBasePageViewSource.cs
chrilith: OK, but I have to check if this bug have been fixed in iOS 10 first. If fixed, I'll add the check.

chrilith:
@martijn00 OK, will remove the legacy code then.

chrilith:
Did a rebase. Will work then on Legacy removal and check the animated:false problem.

chrilith:
Should be up to date and clean now.

martijn00:
It seems that your merge of upstream didn't went in correct. I can see the changes in the files in this PR.

martijn00:
Is there a sample repo for this available?

chrilith:
Will do a quick sample.

martijn00:
How is the sample coming along? Can you also merge the upstream into your branch?

chrilith:
Will do all this on monday!

chrilith:
Hum, why are there conflicts?

Here is the PoC: https://www.dropbox.com/s/3qqayvku6028h6l/MVVMCross_PageViewPoC.zip?dl=0

Cheesebaron:
Someone changed some csproj files and MvxPageViewController outside of this PR and merged into develop. Hence there are conficts. You need to merge develop into your branch or rebase.

chrilith:
OK, did a merge but the solution doesn't compile with Visual Studio. I have no portable target for tvOS and so I cannot update the new tvOS stuff.

Any clue?

Cheesebaron:
I'll try make a new branch and cherry pick your changes, because this looks very messed up. Lets get this into MvvmCross 😃

chrilith:
OK thanks :( 
Do you know how to get tvOS target in Visual Studio?

Cheesebaron:
You shouldn't have to do anything. It won't appear in a PCL as targets though.

If you need to create an actual tvOS project, I think you need to do it on a Mac.

chrilith:
Back, will try to reset and add again.

chrilith:
OK, my branch is totally messed up. Let's close this PR and create a new one as proposed by Cheesebaron.

PR #1410, opened 2016-07-31
Add support for own base class, ShowViewModel with parameter serialization

Plac3hold3r:
Switches out the concert `MvxViewModel<TInit>` type requirement to use an interface `IMvxViewModelInitializer<TInit>` instead for [`ShowViewModel<TViewModel, TInit>`](https://github.com/MvvmCross/MvvmCross/blob/4.0/MvvmCross/Core/Core/ViewModels/MvxNavigatingObject.cs#L42).

The limitation of forcing the concert type is that it will require projects that implement their own abstract base ViewModel to either always specify an Init parameter type class or maintain multiple base ViewModels with possibly the same code and the only difference being the inherited type.

Switching to an interface allows for any single ViewModel to implement the feature of ViewModel parameter serialization in isolation or implement it in an extended base class type. The developer will have to implement their own deserializer much the same as currently done in  `MvxViewModel<TInit>`.


Cheesebaron:
Looks good thanks!

PR #1417, opened 2016-08-29
MvxImageViewResourceNameTargetBinding (VectorDrawable resources)

fedemkr:
Hi guys,

I've added MvxImageViewResourceNameTargetBinding to Android target bindings, that is like MvxImageViewDrawableNameTargetBinding but it uses SetImageResource instead of SetImageDrawable. Also refactored a bit MvxImageViewDrawableTargetBinding to reuse some code.

I've done this because MvxImageViewDrawableNameTargetBinding doesn't load VectorDrawable resources pre-Lollipop.

I'm not sure if this should be here or in MvvmCross-AndroidSupport, but I think it could be here because SetImageResource doesn't belong specifically/only to AndroidSupport.

I hope it helps.


Cheesebaron:
Looks OK.

When commiting in the future, please set `core.autocrlf = true` on windows and `core.autocrlf = input` on mac, so line endings don't get messed up in commits.

PR #1420, opened 2016-09-01
Fix splash screen initialization

andyci:
This is my proposed fix to an issue @davidschwegler raised in PR #1253

### DIFF
--- a/MvvmCross/Droid/Droid/Platform/MvxAndroidSetupSingleton.cs
+++ b/MvvmCross/Droid/Droid/Platform/MvxAndroidSetupSingleton.cs
@@ -27,7 +27,6 @@ namespace MvvmCross.Droid.Platform
         private static TaskCompletionSource<bool> IsInitialisedTaskCompletionSource;
         private MvxAndroidSetup _setup;
         private bool _initialized;
-        private bool _initializationStarted;
         private IMvxAndroidSplashScreenActivity _currentSplashScreen;
 
         public virtual void EnsureInitialized()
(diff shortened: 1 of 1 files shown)


MvvmCross/Droid/Droid/Platform/MvxAndroidSetupSingleton.cs
```diff
@@ -73,31 +71,29 @@ public virtual void InitializeFromSplashScreen(IMvxAndroidSplashScreenActivity s
...
+                    return;
+                }
+                else
+                {
+                    IsInitialisedTaskCompletionSource = new TaskCompletionSource<bool>();
+                    this._setup.InitializePrimary();
+                    ThreadPool.QueueUserWorkItem(ignored =>
+                    {
+                        this._setup.InitializeSecondary();
+                        IsInitialisedTaskCompletionSource.SetResult(true);
```
davidschwegler: Important: Missing the lock around 91-93, to match 74-76. (See https://github.com/andyci/MvvmCross/blob/bdcb56b5c9261bbaf63205785c7b22f166aa32e2/MvvmCross/Droid/Droid/Platform/MvxAndroidSetupSingleton.cs#L91). This could result in two calls to _currentSplashScreen?.InitializationComplete().

Technically, I think this part should be called back onto the main thread instead of the background. Calling InitializationComplete (which essentially calls IMvxAppStart.Start() and launches a new activity, see https://github.com/MvvmCross/MvvmCross/blob/4.0/MvvmCross/Droid/Droid/Views/MvxSplashScreenActivity.cs#L82) from a background thread doesn't make sense to me.

davidschwegler:
Other than the comment I made, I think this is great, thank you Andy!

martijn00:
@andyci Can you fix the comment of @davidschwegler so we can merge this?

andyci:
Updated and ready for merge if @davidschwegler agrees

davidschwegler:
@martijn00 @andyci I can't test it myself, but the code reads like it should work now. Thanks for making the change Andy! 👍

PR #1421, opened 2016-09-02
Adds decimal handling to Combiners

### LINKED ISSUE #1418: Decimal zero not correctly interpreted when used as part of boolean expression
## Steps to reproduce

1.
Create binding something like:
MvxBind="If(MyProperty==0,'Is zero',,'Isnt zero')"

### Expected behavior
We should get "Is zero" statement.
Should we specifiy 0.0 or 0m or anything else.

Later, I tried different approach:
a) created property called EmptyDecimal using:

public decimal EmptyDecimal { get; private set; } = 0;

then in my axml used:
MvxBind="If(MyProperty==EmptyDecimal,'Is zero',,'Isnt zero')"

and voila it is working, but it is really silly solution.

### Actual behavior
You will get "Isnt zero" statement.

### Configuration

**Version:** 4.2.3

**Platform:** Android

willsb:
Closes #1418


kjeremy:
This is tough to diff.  Can you please switch from tabs to spaces so that the diff is easier to follow?

willsb:
Done ;)

Cheesebaron:
Looks good to me

PR #1434, opened 2016-09-22
Improved support for binding to interfaces

drungrin:
When binding to IMenuItem, the Binding Registry search for IMenuItemInvoker, which isn't registered.

The fallback search only search for base classes. 

The added code make it search the implemented interfaces also.

### DIFF
--- a/MvvmCross/Core/Binding/Bindings/Target/Construction/MvxTargetBindingFactoryRegistry.cs
+++ b/MvvmCross/Core/Binding/Bindings/Target/Construction/MvxTargetBindingFactoryRegistry.cs
@@ -116,6 +116,12 @@ namespace MvvmCross.Binding.Bindings.Target.Construction
             {
                 return factory;
             }
+            var implementedInterfaces = type.GetTypeInfo().ImplementedInterfaces;
+            foreach (var implementedInterface in implementedInterfaces)
+            {
+                factory = this.FindSpecificFactory(implementedInterface, name);
+                if (factory != null) return factory;
+            }
             var baseType = type.GetTypeInfo().BaseType;
             if (baseType != null)
                 return this.FindSpecificFactory(baseType, name);
--- a/MvvmCross/Core/Binding/Bindings/Target/Construction/MvxTargetBindingFactoryRegistry.cs
+++ b/MvvmCross/Core/Binding/Bindings/Target/Construction/MvxTargetBindingFactoryRegistry.cs
@@ -116,15 +116,16 @@ namespace MvvmCross.Binding.Bindings.Target.Construction
             {
                 return factory;
             }
+            var baseType = type.GetTypeInfo().BaseType;
+            if (baseType != null)
+                factory = this.FindSpecificFactory(baseType, name);
+            if (factory != null) return factory;
             var implementedInterfaces = type.GetTypeInfo().ImplementedInterfaces;
             foreach (var implementedInterface in implementedInterfaces)
             {
                 factory = this.FindSpecificFactory(implementedInterface, name);
                 if (factory != null) return factory;
             }
-            var baseType = type.GetTypeInfo().BaseType;
-            if (baseType != null)
-                return this.FindSpecificFactory(baseType, name);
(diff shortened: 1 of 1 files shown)


MvvmCross/Core/Binding/Bindings/Target/Construction/MvxTargetBindingFactoryRegistry.cs
```diff
@@ -116,10 +116,16 @@ private IMvxPluginTargetBindingFactory FindSpecificFactory(Type type, string nam
             {
                 return factory;
             }
+            var implementedInterfaces = type.GetTypeInfo().ImplementedInterfaces;
+            foreach (var implementedInterface in implementedInterfaces)
+            {
+                factory = this.FindSpecificFactory(implementedInterface, name);
+                if (factory != null) return factory;
+            }
```
kjeremy: I'm not too familiar with this code but shouldn't this be done after the base type?  My gut says that we should check the base class before any interfaces that may be implemented.

drungrin: It depends on the usage.

If binding to derived classes 'll happen more often than binding to interfaces, yes.

Since we have MvxViewClickBinding (and others), you are right, we should move that.

MvvmCross/Core/Binding/Bindings/Target/Construction/MvxTargetBindingFactoryRegistry.cs
```diff
@@ -116,15 +116,16 @@ private IMvxPluginTargetBindingFactory FindSpecificFactory(Type type, string nam
             {
                 return factory;
             }
+            var baseType = type.GetTypeInfo().BaseType;
+            if (baseType != null)
+                factory = this.FindSpecificFactory(baseType, name);
+            if (factory != null) return factory;
```
Cheesebaron: Will we ever reach the implementedInterfaces code?

It is only reached in the case that there is a factory. Are we interested in going there if we already have a factory?

drungrin: In my case, it allow me to add a click binding to the IMenuItem, aware of ICommand.CanExecute.

`            registry.RegisterFactory(new MvxCustomBindingFactory<IMenuItem>("Click", menuItem => new MenuItemClickBinding(menuItem)));
`

`            set.Bind(SalvarMenuItem)
                .For("Click")
                .To(vm => vm.SalvarCommand);
`

You don't find a factory because the class searched for a factory is Android.Views.IMenuItemInvoker, which is an internal class, and you cann't register a factory to it.

Then you register a factory to IMenuItem (which is correct).

martijn00:
This seems good to merge to me! Any objections?

kjeremy:
I'm not convinced that @drungrin answered @Cheesebaron's question. @Cheesebaron ?

drungrin:
@kjeremy,

The code 'll be reached. 

When you bind to a MenuItem:

```
        public IMenuItem SalvarMenuItem { get; set; }
  ...
        public override bool OnCreateOptionsMenu(IMenu menu)
        {
            MenuInflater.Inflate(Resource.Menu.salvar, menu);
            SalvarMenuItem = menu.FindItem(Resource.Id.menu_salvar);            
            var set = this.CreateBindingSet<MonitorarProcessoView, MonitorarProcessoViewModel>();            
            set.Bind(SalvarMenuItem)
                .For("Click")
                .To(vm => vm.SalvarCommand);
            set.Apply();
            return base.OnCreateOptionsMenu(menu);
        }
```
When you reach FindSpecificFactory, you get a Android.Views.IMenuItemInvoker.

The _lookup has a 'Android.Views.IMenuItem:Click', so you don't have a factory.

We cannot register a factory for IMenuItemInvoker because it is a internal class.

So it'll search using the interfaces, and find a factory for 'Android.Views.IMenuItem:Click', which is correct.

PR #1436, opened 2016-09-24
it always be 'false'

g0rdan:


martijn00:
What does this fix? It might seem unnecessary, but this code has been in production like this for a long time.

g0rdan:
I just noticed this and decided send pull request. If better way it's not removing lines, it will be cool :) But yeah, it won't fix anything.

martijn00:
Thanks for the effort! Keep up the good work! I will close this for now, until this gets an issue.

Issue #1431, opened 2016-09-19
Crash when rehydrating an activity on Android 2.3.3


RyanBuzzInteractive:
Hi Guys,

Sorry to paste another old version support ticket. I have a crash when resuming a killed activity on Gingerbread. I think I have tracked down the issue and it should be an easy fix.

https://github.com/MvvmCross/MvvmCross/blob/4.0/MvvmCross/Droid/Shared/Caching/FragmentCacheConfiguration.cs

Line 79 of FragmentCacheConfiguration is: 
var jsonSerializedMvxCachedFragmentInfosToRestore = savedInstanceState.GetString(SavedFragmentCacheConfiguration, string.Empty);

The two-parameter version of GetString was only added in API 12. I get a 
--- End of managed Java.Lang.LinkageError stack trace ---
java.lang.NoSuchMethodError: getString

I should fix this myself and submit it but I haven't done any contribution to open source projects before and i'm up against it time wise now. I think I can just get around it for now by implementing my own IFragmentCacheConfiguration and I could submit that to you guys later, but thought I would leave this here for reference.

ref: http://stackoverflow.com/questions/11110423/nosuchmethoderror-bundle-getstring

Stack trace:
`
[...]

PR #1439, opened 2016-09-27
Use the the lower level GetString on older platforms

martijn00:
Fixes https://github.com/MvvmCross/MvvmCross/issues/1431


kjeremy:
Looks good to me.  Do we even support Gingerbread?

martijn00:
I think we do support it, mostly there is no reason not to. If you are using the Android support libraries you will be limited to API level 7.

kjeremy:
Great.  Then :+1:

Issue #1423, opened 2016-09-10
MvxTaskBasedBindingContext issue with dynamic cell's height in iOS


softlion:
The datacontext of each cell is set by MvxListView when it dequeues a cell. The datacontext setter (which is a MvxTaskBasedBindingContext ) triggers OnDataContextChange which sets the DataContext of each registered cell binding asynchronously. 
The cell may not know its height fast enough for autoheight to work. This is an issue in iOS when using dynamic cell's heights, where a cell can have an invalid height. And on Android when scrolling fast down or up where there may be some jump artifacts. 

DataContext setter in
https://github.com/MvvmCross/MvvmCross/blob/master/MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs

GetCell in
https://github.com/MvvmCross/MvvmCross/blob/master/MvvmCross/Binding/iOS/Views/MvxBaseTableViewSource.cs

A solution would be to create a new SynchronousDataContext property, and use this in MvxListView to set the DataContext instead of the original interface. Another solution would be to have a SetDataContextAsync() method and wait for this async task in MvxListView when iOS asks for the cell's height. The second one will have the best performances, as it let the task run between calls to GetCell and GetCellHeight.

softlion:
Note that you can easily switch to the legacy binding context, see the instructions there:

https://mvvmcross.com/v3.2/blog/mvvmcross-42

PelleRavn:
@softlion For some reason (meaning, I haven't found the reason yet), the fix for going back to the legacy binding context doesn't work for 4.3.0, so the issue you are mentioning happens no matter if you use MvxTaskBasedBindingContext or just MvxBindingContext.

MarcBruins:
I'm busy trying to fix this issue but i don't have a good enough understanding what goes on right now. Love some help (i'm on slack). Already made some changes and added a test project: https://github.com/MarcBruins/MvvmCross/tree/feature/dynamicheightios

softlion:
Marc i just reviewed and answered you on slack. Seems very good. Simply remember that you will never achieve 60fps in your list because of this code. Maybe add a small comment over it to explain the case. You'll need to do the same change in collection views.

PelleRavn:
So, now that it's closed should we open a new issue for collection views as @softlion mentioned?

> You'll need to do the same change in collection views.

martijn00:
Yes! Please do! I think @MarcBruins had some concerns on that.

[...]

PR #1442, opened 2016-09-29
Synchronous resizable cells for iOS

MarcBruins:
This PR fixes https://github.com/MvvmCross/MvvmCross/issues/1423 for the iOS autoresizing cells. The problem is that for autoresizing to work the cells need there content, in a async situation this content may arrive later. If this happens iOS cannot estimate the size of the cell correctly and therefore it will give a faulty size. This PR check if autoresizing is on, if so it will run synchronous so iOS can do its resizing properly


martijn00:
Nice work! Looks good to me!

kjeremy:
Yikes! We should squash things like this next time :)

RefuX:
Hi there, just ran into this issue. Any timeline on when this will make it into a release?

Thanks :)

Cheesebaron:
I'll do one soon.

PR #1453, opened 2016-10-07
Add test projects to easily test changes that are made to the project

MarcBruins:
Included a solution folder "TestProjects". This is usefull when making changes to the MvvmCross code base and when you want to test and debug these changes directly. There is only Droid and iOS for now


martijn00:
Seems like a good start! We still need to extend this with many more test cases.

Cheesebaron:
There is a nice APISample project in the Samples repo, why not use that?

martijn00:
You mean copy that over to this repo? @Cheesebaron

Cheesebaron:
Or replace that empty testproject with it. At least it will have contents...

MarcBruins:
Why do you want to have contents? When i created PR #1442, i wanted a test project that had a UITableViewController and some content that was added async. It is easier to create the classes for your specific use-case then its is to change a existing Sample. The biggest hassle was to setup a project so that i could test my changes to the mvvmcross core and that is what this PR does.

Cheesebaron:
If you want to document usage of a PR as a sample, you probably want to add stuff to an API sample. If it is an empty project, I see no reason to add it to begin with. Just create a new using the VS or XS templates...

Issue #1456, opened 2016-10-10
Add MVVMCross ObservableCollection


chrislamont:
I have a class inheriting from ObservableCollection. 

This loses some functionality, specifically the ability to RaisePropertyChanged.

Would it be beneficial/possible to implement this within MVVMCross

kjeremy:
Post this on StackOverflow with some code.

martijn00:
I get the idea, and this would be useful for many users. We could add things like `AddRange` to our MvxObservableCollection, and implement `MvxNotifyPropertyChanged` instead of the default one. The question is if we want this in this repo.

SeeD-Seifer:
There was a try once https://github.com/MvvmCross/MvvmCross/pull/212 :)

kjeremy:
@martijn00 I think I misunderstood the issue when I closed this so feel free to reopen if you want.  If we want something like this (I'm not convinced but then again the description on this issue is pretty vague) we should add a way to suppress change notification.

martijn00:
That try actually looks promising! I think we could try to get this in again for 5.0

PR #1462, opened 2016-10-14
Add a MvvmCross ObservableCollection

martijn00:
Inspiration from:
https://github.com/MvvmCross/MvvmCross/pull/212
https://github.com/MvvmCross/MvvmCross/pull/219

Fixes: https://github.com/MvvmCross/MvvmCross/issues/1456


kjeremy:
This looks reasonable to me.  Note that we could optimize this (esp for bulk operations) by not inheriting from `ObservableCollection` itself and doing something similar to https://github.com/reactiveui/ReactiveUI/blob/006e41bef02c4785d2929cab2f9265ede682fb1f/src/ReactiveUI/ReactiveList.cs where we wrap an inner `List` or similar.

martijn00:
I think we could do that as an improvement later on. For now i would like to just use this and add an up for grabs to optimize it.

kjeremy:
It looks like we're just doing `Resets` when the collection changes and not filling in the indices which can get quite expensive with large lists.

PR #1469, opened 2016-10-19
Enable usage of Multiple MvxAttributes specifying different layouts

martijn00:
For different Android specifications

This will enable you to specify a different frame layout on for example tablet.

### DIFF
--- a/MvvmCross/Droid/Shared/Presenter/MvxFragmentsPresenter.cs
+++ b/MvvmCross/Droid/Shared/Presenter/MvxFragmentsPresenter.cs
@@ -21,8 +21,8 @@ namespace MvvmCross.Droid.Shared.Presenter
     {
         public const string ViewModelRequestBundleKey = "__mvxViewModelRequest";
 
-        private readonly FragmentHostRegistrationSettings _fragmentHostRegistrationSettings;
-        private readonly Lazy<IMvxNavigationSerializer> _lazyNavigationSerializerFactory;
+        protected FragmentHostRegistrationSettings _fragmentHostRegistrationSettings;
+        protected Lazy<IMvxNavigationSerializer> _lazyNavigationSerializerFactory;
 
         protected IMvxNavigationSerializer Serializer => _lazyNavigationSerializerFactory.Value;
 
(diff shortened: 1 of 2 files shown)


MvvmCross/Droid/Shared/Presenter/MvxFragmentsPresenter.cs
```diff
@@ -21,8 +21,8 @@ public class MvxFragmentsPresenter
     {
         public const string ViewModelRequestBundleKey = "__mvxViewModelRequest";
 
-        private readonly FragmentHostRegistrationSettings _fragmentHostRegistrationSettings;
-        private readonly Lazy<IMvxNavigationSerializer> _lazyNavigationSerializerFactory;
+        protected FragmentHostRegistrationSettings _fragmentHostRegistrationSettings;
+        protected Lazy<IMvxNavigationSerializer> _lazyNavigationSerializerFactory;
```
kjeremy: I would leave these as `private readonly` and just expose them as a protected property unless we want them to be assigned outside of the constructor.

kjeremy:
See my one comment.  This looks like a great idea.  Is the idea to specify an attribute per layout where the view id may be different?

MvvmCross/Droid/Shared/Presenter/MvxFragmentsPresenter.cs
martijn00: What do you mean by expose them as a protected property?

martijn00:
Yes that is the idea! This makes it possible to add multiple attributes to a class, where you on lets say a `layout-sw600dp` layout make use of a master detail frame, and on normal layout just a content_frame.

MvvmCross/Droid/Shared/Presenter/MvxFragmentsPresenter.cs
kjeremy: If they are only assigned in the constructor I would leave them as `private readonly` so that they never get assigned again.  Then expose them as say: `protected FragmentHostRegistrationSettings FragmentHostRegistrationSettings => _fragmentHostRegistrationSettings` so that child classes can still access them.  It just prevents people from changing the reference.

martijn00: Well, i made some up the methods in `FragmentHostRegistrationSettings` virtual so people can make their own class and change that behaviour. They would need to set this some way.

kjeremy: Okay.  Fine by me.

PR #1486, opened 2016-11-16
Fixes  #1485 - MvxObservableCollection RemoveRange raises Reset action

### LINKED ISSUE #1485: MvxObservableCollection RemoveRange Exception
### Steps to reproduce

1. Create instance of MvxObservableCollection.

2. Add some range of elements.

3. Try to remove some range of elements.

### Expected behavior
Items must be removed.

### Actual behavior
You will get an exception: Constructor supports only the 'Reset' action. 

### Configuration

**Version:** 4.4

**Platform:** iOS/Android/WPF/Windows Store/Windows Phone Store/Windows UWP

That is because we have to pass only [Reset](https://msdn.microsoft.com/en-us/library/ms653204(v=vs.110).aspx) on this [call](https://github.com/MvvmCross/MvvmCross/blob/master/MvvmCross/Core/Core/ViewModels/MvxObservableCollection.cs#L216).

Prin53:
Fix for  #1485

### DIFF
--- a/MvvmCross/Core/Core/ViewModels/MvxObservableCollection.cs
+++ b/MvvmCross/Core/Core/ViewModels/MvxObservableCollection.cs
@@ -213,7 +213,7 @@ namespace MvvmCross.Core.ViewModels
             {
                 SuppressEvents = false;
 
-                OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Remove));
+                OnCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset));
             }
         }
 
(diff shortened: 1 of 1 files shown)


kjeremy:
Is there a way to fix this without forcing a `Reset`?

kjeremy:
So after staring at this I think `Reset` is the correct thing to do in this instance of `RemoveRange`.  However there should really be an alternate `RemoveRange` that works on an **actual range** (ie contiguous) of elements.

Something like:

1. ```RemoveRange(int start, int count)``` which would remove an actual range from the collection.
2. Then we should take the current `RemoveRange` and rename it to something like `RemoveItems` because the *Range* name describes the shape of the input and not the operation.

martijn00:
Thanks!

PR #1494, opened 2016-11-25
Optimize MvxObservableCollection a bit

kjeremy:
* Makes event suppression re-entrant.
* Fixes a non-intuitive code smell in `RemoveRange` where events would be suppressed and depended on a callee to alter the state.  Make this explicit.
* Raise the least restrictive event args in some places.

There is probably more work to be done (ie `ReplaceRange` is a weird function and can alter the size of the collection).


kjeremy:
Someone review this.  Minimal testing but should make some operations faster.

martijn00:
Looks ready to merge to me

PR #1497, opened 2016-11-28
WeakReference.IsAlive may not be accurate if true

kjeremy:
The correct way to check for live-ness is to test Target for non-null.
If IsAlive is false then we know the target has been GCed.  If true then
it could be GCed between the check and the access.

### DIFF
--- a/MvvmCross/Core/Core/ViewModels/MvxCommand.cs
+++ b/MvvmCross/Core/Core/ViewModels/MvxCommand.cs
@@ -53,8 +53,8 @@ namespace MvvmCross.Core.ViewModels
                 {
                     foreach (var thing in this._eventHandlers)
                     {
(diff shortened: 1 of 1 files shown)


Cheesebaron:
I guess this check should be the same here: https://github.com/kjeremy/MvvmCross/blob/f71a96cee0500208ead864b34b81d6e7ceff11a0/MvvmCross/Core/Core/ViewModels/MvxCommand.cs#L76

kjeremy:
No that check is okay.  If `IsAlive` is `false` then the reference has truly been GCed.  It's the case of:

```csharp
if (ref.IsAlive)
{
  // At this point the GC might run before the next line.
  var target = ref.Target;
}
```

that's a problem.

Cheesebaron:
Roger. Thanks!

Issue #1398, opened 2016-07-18
Bindings in 4.2.x  System.AggregateException caught in HockeyApp


NielsWillems:
When binding an observable collection as ItemSource for a MvxTableViewSource it works without HockeyApp. When HockeyApp is added the app crashes unless I comment out the bindings in a cell. When I want to use both I have to add the following to the Setup.cs file of the iOS app.:

`Mvx.RegisterType<IMvxBindingContext, MvxBindingContext>();`

The stack trace is provided  further down this post.
### Expected behavior

Properties in cell should be rebound
### Actual behavior

App crashes
### Configuration

**Version:** 4.2.2 (I believe also 4.2.0 but not 100% sure)

**Platform:** iOS

**HockeyApp StackTrace:**

```
Application Specific Information:
*** Terminating app due to uncaught exception 'System.AggregateException', reason: 'System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. ---> System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
  at System.ThrowHelper.ThrowInvalidOperationException (ExceptionResource resource) [0x00000] in /Users/builder/data/lanes/3412/3cf8aaed/source/maccore/_build/Library/Frameworks/Xamarin.iOS.framework/Versions/git/src/mono/external/referencesource/mscorlib/system/throwhelper.cs:99 
  at System.Collections.Generic.List`1+Enumerator[T].MoveNextRare () [0x00016] in /Users/builder/data/lanes/3412/3cf8aaed/source/maccore/_build/Library/Frameworks/Xamarin.iOS.framework/Versions/git/src/mono/external/referencesource/mscorlib/system/collections/generic/list.cs:1180 
  at System.Collections.Generic.List`1+Enumerator[T].MoveNext () [0x00050] in /Users/builder/data/lanes/3412/3cf8aaed/source/maccore/_build/Library/Frameworks/Xamarin.iOS.framework/Versions/git/src/mono/external/referencesource/mscorlib/system/collections/generic/list.cs:1174 
  at MvvmCross.Binding.BindingContext.MvxTaskBasedBindingContext.<OnDataContextChange>b__20_0 () [0x0008d] in C:\vcs\git\MvvmCross\MvvmCross\Core\Binding\BindingContext\MvxTaskBasedBindingContext.cs:127 
[...]

PR #1501, opened 2016-12-02
#1398

richardblewett:
Fixed threading issue in MvxTaskBasedBindingContext where collections were being iterated over on a background thread when they could be modified on the main thread. Now take copies of the collections on the main thread before iterating over the copies in the background thread

### DIFF
--- a/MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs
+++ b/MvvmCross/Core/Binding/BindingContext/MvxTaskBasedBindingContext.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using MvvmCross.Binding.Binders;
 using MvvmCross.Binding.Bindings;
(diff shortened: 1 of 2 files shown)


kjeremy:
It looks like the second commit removes most of the changes from the first one.  Can you squash and force push to clean this up?

I THINK this makes sense however if something is modified in the background doesn't it fall through the cracks (ie it doesn't get the change that the copy gets)?

richardblewett:
@kjeremy I've squashed the second commit to clean up. Its only the collections that are copied, the actual bindings are not clones. 

Any inherent issue in the collections changing while the datacontext fixup takes place can be fixed by taking locks around this whole process and the register methods. Those issues were already there without this change. This change just prevents the InvalidOperationException for the collection being modified during iteration.

Would applying any existing datacontext to the binding during registration resolve the issue without locks or would that introduce other issues in the binding process?

Cheesebaron:
Not sure how to test this. Blast through loads of binding changes on a List type of view?

kjeremy:
@richardblewett can you rebase this on master and do a force push please?

kjeremy:
@richardblewett not sure I follow what you mean.

richardblewett:
@kjeremy I thought I had rebased on master and force pushed in the fork already. Are you not seeing this? Or do I need to do something in the pull request (you may have guessed I'm a little new to pull requests)

You were concerned about changes made after the copy had been taken not getting the new datacontext? If, when a binding is registered, any existing datacontext was assigned, would that solve the problem?

Issue #1500, opened 2016-12-02
Move back sub repos into main


Cheesebaron:
To have a much less complex build process we would need to merge back in the sub repos into the main repo. This is due to now the build process would look something like this:

1. Build main repo 
2. Create nuget packages
3. For each of the other repositories:
  a. Restore Nuget Packages
  b. Update Nuget Pacakges
  c. Build repo
  d. Create Nuget Packages
  e. Commit Nuget changes done in b.

Especially the 3.e. would be very problematic. I.e. how do you determine that 3.b. was successful. It is not unseen that NuGet Updates go wrong.

Instead it would be much easier to just have step 1. and 2., nothing more. This requires everything is in the same repository.

So the suggestion is to move, Plugins, AndroidSupport, iOSSupport and Forms back into MvvmCross.

kjeremy:
I'm torn on this.  I agree that it makes builds easier and it will be much more efficient to hack on things like the android support packages but there's something... sucky about having so many projects in the same repo.

An alternative would be to have a repo of submodules that we use for the build but I don't see why we would want that over your proposed solution.

kjeremy:
I suppose another alternative would be to have a stable interface between MvvmCross and the rest of the repos.

willsb:
@kjeremy I think using submodules is better than merging back into main because it keeps the repositories apart; I don't need to download MvvmCross if I just want to change something in a Plugin, for example.

Isn't it why [this](https://github.com/MvvmCross/MvvmCross-Build) repo exists?

kjeremy:
@willsb yes I believe that was the original purpose.

I agree on your plugin example though we would ultimately need a commit to bump the submodule in the main repo when it comes time for a release of the plugin.  Ultimately I think that's one downside we may face: one offs will be hard (though maybe that's not a big deal).

ghuntley:
I highly/strongly recommend merging back to a single repository. Happy to jump on a video chat  with the core team and share some notes/thoughts about the journey RxUI went through.

Cheesebaron:
It has been done!

Now we need to create a bunch of labels for issues and nuke the old repos or redirect them somehow.

PR #1505, opened 2016-12-08
Merge Sub repos into main repo

martijn00:


Cheesebaron:
This is for #1500 

Also, while you are doing this. It would be great to fix references so they are not dependent on NuGet packages for the sub repos.

martijn00:
Yes! I'm working on fixing the references!

kjeremy:
My gut says that we need to move more around.

Having top level directories `MvvmCross-Forms`, `MvvmCross-AndroidSupport`, `MvvmCross-iOSSupport`, `MvvmCross-Plugins` is good. Then I think we may need a `MvvmCross` directory that has most of the other directories: `MvvmCross`, `Cirrious`, the solution files etc.

kjeremy:
Looks like it preserves the history too even if it's a bit crazy.  Not sure about tags etc though.

martijn00:
@kjeremy what do you mean with the structure? It already is like that after this PR?

kjeremy:
My bad.  I have some stale directories

Issue #1503, opened 2016-12-08
Update Android OnAttach


martijn00:
The currently used OnAttach is deprecated:
`public override void OnAttach(Activity activity)`

We should look into using:
`public override void OnAttach(Context context)`

We need to update the eventsource methods to reflect this.

See also: http://stackoverflow.com/questions/32083053/android-fragment-onattach-deprecated

PR #1509, opened 2016-12-11
Update deprecated OnAttach(Activity) to OnAttach(Context)  (#1503)

orzech85:
#1503


martijn00:
Initial changes look good! What i think is that this is still missing in Android Support Fragments.

Also when i look at http://stackoverflow.com/questions/32083053/android-fragment-onattach-deprecated It looks like OnAttach(Context) is not being called when you use FullFragging.

What you could do is Add OnAttach(Activity) back to FullFragging, and cast the Activity to Context on Androids lower than v23.

martijn00:
Changes looks good now!

PR #1514, opened 2016-12-12
Change misleading trace message

gartesk:
For MvxBindingMode.Default binding mode both methods RequireSourceObservation and RequiresTargetObservation return true just like for MvxBindingMode.TwoWay mode, but trace assumes us that default mode is equal to MvxBindingMode.OneWay.


martijn00:
Thanks!

PR #1520, opened 2016-12-12
Remove CrossUI, AutoView and Dialog from Mac solution

MarcBruins:


martijn00:
The Mac Solution needs a good update where the plugins, android support and ios support are included.

Issue #1465, opened 2016-10-16
Switch to ViewHolder pattern instead of wrapping in FrameLayout for MvxListView items


Cheesebaron:
Currently we have a dummy `MvxListItemView` which itself is a `FrameLayout`, meaning we wrap every item in the list in an additional `FrameLayout` adding another layer to the onion.
This exact wrapping also breaks adjusting `RadioButton`s inside of a `RadioGroup`, since the View hierarchy is looking like:

```
RadioGroup (LinearLayout)
    FrameLayout
        RadioButton
    FrameLayout
        RadioButton
```

Instead of

```
RadioGroup (LinearLayout)
    RadioButton
    RadioButton
```

This means any layout attribute set on a `RadioButton` is not actually used within the `RadioGroup`.

Using `MvxListItemView` is kind of using the `ViewHolder` pattern. However, I think it should be possible to remove the view nesting, just like we do in MvxRecyclerView in the Support library.

orzech85:
I am working currently on this issue. My idea is to remove inheritance from FrameLayout and pass parent from adapter instead. The problem is that MvxListItemView has also responsibility for caching data context dependent on own lifecycle (OnAttachedToWindow() and OnDetachedFromWindow()) and implement ICheckable. I am going to move first responsibility to first child, but currently I have no idea what to do with ICheckable... I will think tomorrow about it, but maybe someone has some ideas about that.

Cheesebaron:
@orzech85 try look at the implementation of ViewHolder in MvxRecyclerView.

orzech85:
I have tried, but ViewHolder pattern is build-in in RecyclerView.Adapter in contrast to Adapter for ListView. eg. callback method OnViewAttachedToWindow exists in RecyclerView.Adapter itself.

Cheesebaron:
But does the "ViewHolder" for a `BaseAdapter` need to be a view. Can't it just contain a inflated view, which you subscribe to that views lifecycle events? 
Haven't looked into this much further, but wrapping all rows in an additional view is what is breaking the layout of rows for all `AdapterViews` that use `MvxAdapter`.

orzech85:
Yes, sure. I will try hack on this today.

Issue #1451, opened 2016-10-06
RadioButton's width doesn't use match_parent


iqbalmineraltown:
I was trying to create border around each of radio buttons, but it wouldn't set its `width` to `match_parent` as I specified in XAML. The `width` would always be set to `wrap_content`

But if I specify the `width` in numbers (e.g. 300dp), it works

This is related to the problem mentioned in https://github.com/MvvmCross/MvvmCross/pull/454

Is this behaviour intended?
### Steps to reproduce

My RadioGroup

```
<MvxRadioGroup
        local:MvxBind="ItemsSource OptionList; SelectedItem SelectedOption"
        local:MvxItemTemplate="@layout/radio_button_item"
        android:orientation="vertical"
        android:layout_width="match_parent"
        android:layout_height="wrap_content"
        android:paddingLeft="20dp"
        android:paddingRight="20dp" />
```

RadioButtonTemplate (radio_button_item)

```
<?xml version="1.0" encoding="utf-8"?>
<RadioButton xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:local="http://schemas.android.com/apk/res-auto"
    android:layout_marginTop="10dp"
    android:background="@drawable/style_border_item"
    android:paddingLeft="25dp"
    android:paddingTop="25dp"
    android:paddingBottom="25dp"
    android:layout_width="match_parent"
    android:layout_height="wrap_content"
    local:MvxBind="Text Value"
    android:text="Test"
    android:textSize="16sp" />
```

My RadioButton's Background(style_border_item)

```
<?xml version="1.0" encoding="utf-8"?>
<selector xmlns:android="http://schemas.android.com/apk/res/android">
  <item android:state_checked="true">
    <shape android:shape="rectangle">
      <stroke
         android:width="2dp"
         android:color="@color/colorLightGreen" />
    </shape>
  </item>
  <item>
    <shape android:shape="rectangle">
      <stroke
       android:width="2dp"
       android:color="@color/colorGrey" />
    </shape>
  </item>
</selector>
```
### Expected behavior

RadioButton's `width` should `match_parent`
### Actual behavior

RadioButton's `width` were `wrap_content` regarding to each own's `text`
### Configuration

**Version:** 4.1.0

**Platform:** Android

Cheesebaron:
I guess it is because MvxAdapter wraps stuff in MvxListItemView, hence the the width is not respected.

Cheesebaron:
[...]

PR #1535, opened 2016-12-15
[WIP] Issue 1465 - Switch to ViewHolder pattern instead of wrapping in FrameLayout for MvxListView items

orzech85:
#1465 

I would like to share my idea to solve this issue. I am going to refactor it (eg. remove MvxListItemView at all) etc. I have used View.Tag for store BindingContext (eg. for RadioButtons). It is probably not the best solution, but in ListView View Holder pattern Tag is used to hold the view...
What do you think about it?

Please notice that I am quite fresh MvvmCross contributor, but I am thirst of knowledge about Mvx :)


kjeremy:
If you are going to use tags please use a key: https://developer.android.com/reference/android/view/View.html#setTag(int, java.lang.Object).

If you could fix your indentation to match the rest of the files that would make this a little more readable.

FWIF I'm not sure I like storing the binding context in a tag.  Will review more later.

orzech85:
I have used data from Adapter except pass DataContext in Tag. It is similar like in MvxListViewSelectedItemTargetBinding. What do you think about that now?

martijn00:
The only remark i have is to cleanup the lines that you commented out.

orzech85:
Ok. I would like to clean it up at all before merge.

orzech85:
Unfortunatelly I have no idea how to solve that without Tag. In Adapter `convertView` is of type `View`. Only possibility to add some object to view is `Tag`. But it is common for ViewHolder pattern for ListView (https://developer.android.com/training/improving-layouts/smooth-scrolling.html). I have used `Tag` without key, because `SetTag(int, Object)` is significant slower. @kjeremy - if you think despite this I should use `Tag` with key, please let me know. 

This pull request removes unnecessary view nesting from ListView/RadioGroup items. I have tried to implement ItemTemplates like in RecyclerView, but it is a bigger refactoring :) Maybe it should be a subject to another pull request?

kjeremy:
How is using `SetTag` measurably slower?  That sounds odd to me since they both go through JNI.  We also use `SetTag` when inflating views: https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Binding/Droid/Views/MvxLayoutInflater.cs#L52

orzech85:
I have read here about that: http://www.jaysoyer.com/2014/12/settag-vs-settag/

> However when using the setTag(int, Object) method, there’s a variable cost.  Remember, internally it stores each View in a SparseArray which uses a binary search for retrieval.  That means for every View stored, you get hit on average with a O(log n) look up. On top of that, you need to type cast every found object.  This is all probably negligible given one or two Views to construct, however as View complexity goes up so will the cost.

kjeremy:
Wow!  I didn't know about that.  Yet one more legitimate reason to hate on Android.  Leave the tag as is then.

Cheesebaron:
Is this ready for testing and review?

orzech85:
I think yes.

Cheesebaron:
Cool, will give it a spin and see what I can find.

Cheesebaron:
Just tested ListView a bit. View Hierarchy looks good. No more wrapping and bindings seems to work OK.

Cheesebaron:
I've verified that https://github.com/MvvmCross/MvvmCross/issues/1451 is fixed in this PR.

This seems good to go.

kjeremy:
Can we  clean up the whitespace before pulling it in?

Cheesebaron:
Whitespace looks OK in XS. Haven't checked in VS.

kjeremy:
The diff on github shows some things misaligned.

Cheesebaron:
Will check it in VS. Probably Tabs vs Spaces territory.

PR #1537, opened 2016-12-15
Reduce memory pressure in MvxNotifyPropertyChanged

kjeremy:
For applications that raise a lot of PropertyChanged notifications like https://github.com/inormis/MvxSIGSEVTest with a timer this helps shave off some allocations (thousands of objects).

The first commit avoids boxing allocations when setting value types (int, bool, etc).  The second statically allocates the all property changed notification.  Previously we were allocating an Action every time a property changes.  The third commit only does this if we need it.

Room for improvement: we could cache all PropertyChangedEventArgs in a static Dictionary but that is left as a followup if wanted.


softlion:
Note that ShouldAlwaysRaiseInpcOnUserInterfaceThread is of no use if using the TaskBased binder.

PR #1556, opened 2017-01-11
Xamarin.Forms.2.3.3.180

kjeremy:


martijn00:
Don't forget to update the nuspec dependency on this as well.

Issue #1582, opened 2017-01-31
Virtualize MvxExpandableListAdapter methods


viniciusmaia:
Please, in futre versions of MvvmCross, change the MvxExpandableListAdapter methods to enable that methods be overwriten, because I had situations that I needed override GetChildView method and had problems.

kjeremy:
Can you submit a PR?

viniciusmaia:
What is PR? I'm sorry, my english is very limited.

kjeremy:
Make a pull request with the changes you want and submit it.

PR #1583, opened 2017-01-31
Virtualize MvxExpandableListAdapter methods

viniciusmaia:
Please, in futre versions of MvvmCross, change the MvxExpandableListAdapter methods to enable that methods be overwriten, because I had situations that I needed override GetChildView method and had problems.

### DIFF
--- a/MvvmCross/Binding/iOS/Views/MvxBaseTableViewSource.cs
+++ b/MvvmCross/Binding/iOS/Views/MvxBaseTableViewSource.cs
@@ -106,7 +106,7 @@ namespace MvvmCross.Binding.iOS.Views
             var item = this.GetItemAt(indexPath);
             var cell = this.GetOrCreateCellFor(tableView, indexPath, item);
 
-            var bindable = cell as MvxTableViewCell;
+            var bindable = cell as IMvxBindable;
 
             if (bindable != null)
             {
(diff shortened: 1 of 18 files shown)


martijn00:
It seems that this was done once before by https://github.com/MvvmCross/MvvmCross/commit/3fc1288e3276a4ac989f44d1bfc17f808a782c1f

Was this change lost somehow?

martijn00:
@jamie94bc do you know what to choose here?

martijn00:
After looking into this, it seems good to go. Let's try this and revert if problems come up.

jamie94bc:
I think this once was the case, but was overwritten by another PR. Thanks for merging!

viniciusmaia:
The problem isn't in TableViewCell. Is a problem of MvxExpandableListAdapter of Android.

jamie94bc:
@viniciusmaia can you submit a new PR with the proposed fixes?

viniciusmaia:
Done!

https://github.com/MvvmCross/MvvmCross/pull/1597

viniciusmaia:
Jamie, I created an issue for this times ago.

That's the issue: https://github.com/MvvmCross/MvvmCross/issues/1582

It's a simple change. Just add the keyword virtal MvxExpandableListAdapter methods to allow them be overwritten.

https://github.com/MvvmCross/MvvmCross/blob/4.0/MvvmCross/Binding/Droid/Views/MvxExpandableListAdapter.cs#L21

jamie94bc:
This is already the case in develop, master and the 4.4 release: https://github.com/MvvmCross/MvvmCross/blob/4.4.0/MvvmCross/Binding/Droid/Views/MvxExpandableListAdapter.cs

4.0 is an old branch.

viniciusmaia:
Ow, thanks!

PR #1588, opened 2017-02-06
Fix some broken links in readme

waf:
I noticed the MvvmCross Blog link was not working in the README. I validated the other URLs, while I was at it.

- **MvvmCross blog** - maybe a recent SSL change broke this?
- **N plus 1 Videos Of MvvmCross** - Looks like there are two pages similarly named. "N plus 1 Videos Of MvvmCross" and "N 1 Videos Of MvvmCross". The latter is blank.
- **Customizing App and Setup** - British vs American English typo. The actual URL uses "Customizing" with a "z"
- **Sqlite-net** - case sensitive URL fix


martijn00:
Thanks!

Issue #1162, opened 2015-10-26
MvxAutoCompleteTextView/MvxFilteringAdapter locks when PartialText is the same value


iambacon:
I'm seeing an issue that is caused when retrieving the items for an autocomplete list from the db. Sometimes it can take a couple of seconds to retrieve the data. In that time the user could have deleted and then re-added the same character again. 

It is this method which locks and never returns https://github.com/MvvmCross/MvvmCross/blob/0be52390b80a2a742f815d02beb64afac8d9a7bd/Cirrious/Cirrious.MvvmCross.Binding.Droid/Views/MvxFilteringAdapter.cs#L66

`PartialText` is set to the new value (which will be the same as the last value), but the bound view model property does not get set. This means that you will never get past `_dataChangedEvent.WaitOne()`
#### Repro steps

_This can be repro'd when there is a delay in the `ItemsSource` being set (e.g. data from a db)_
1. type in search term (e.g. mvvm). In the logger you will see:
   
   `Wait starting for mvvm` 
2. Before the wait has finished, delete the last character (m) and re-add the same value
3. In the logger you will see:
   
   ```
   Wait finished with 2 items for mvvm  
   Wait starting for mvvm
   ```
   
   The wait will never finish.

Cheesebaron:
I am not even sure why it is using ManualResetEvent, could probably be made smarter than it is right now. Thanks for reporting this.

vvolkgang:
I didn't researched into the problem but, giving it is a lock in an async context, `Nito.AsyncEx` nuget has some great helpers for async code, for instance, `AsyncManualResetEvent`

LeftOfHere:
Has this issue been resolved or is there a work around for it?

I'm having the same issue at the moment.

Cheesebaron:
Haven't looked into this since. I still don't understand what that ManualResetEvent is doing there. What is it waiting for?

PR #1590, opened 2017-02-06
Fix MvxFilteringAdapter unreliability.

aaerox:
SetConstraintAndWaitForDataChange would execute twice for one input change with identical input, which would cause two waits but only one PartialText binding change. 

This prevents it from triggering events which won't resolve due to identical values.


aaerox:
This should fix #1162.

PR #1591, opened 2017-02-07
Url escaping for phonenumber in IOS

svego:
Escape the phone number. This makes phone numbers starting with + work in the IOS dialer.


Cheesebaron:
Thanks!

Issue #1315, opened 2016-05-02
Support URL based navigation


Cheesebaron:
In order to improve deep linking into MvvmCross applications, it would be very nice with URL based navigation.

I suggest that we create a new Attribute that you can use for ViewModels or maybe for the Page/ViewController/Activity (not sure yet) like:

```
[MvxUrl("whatever")]
public class WhateverViewModel : MvxViewModel
```

Then when someone visits the Url: _http://my.app/whatever_ it will navigate to that ViewModel.

Additionally, if you have a hierarchy of some sort like _http://my.app/one/two/three_ it should create some kind of index to know what should happen when you navigate back in the application. So, from that particular Url, when navigating back you would go three, two and then one.

I would suspect that this would include extending current presenters or make a new one which supports this.
Xamarin.Forms recently added support for this. Currently it seems like you just add a name to the Page and register URLs in the entry point of that Forms application to which URLs you want to support in the app.

Any thoughts on this?

modplug:
This could probably be used for navigating from 3d-touch shortcuts on iOS as well.

Johan-dutoit:
Should the url define the hierarchy? (maybe as a default yes)

`ie.  http://my.app.com/cheese/1231/reviews`
the 1231 (id of the cheese)  portion of the url should not be considered for the back.

perhaps

```
[MvxUrl(http://my.app.com/cheese/1231/reviews, CheeseViewModel)] //Params of ViewModels
public ReviewViewModel...
```

charri:
Our team created something like you're describing in a previous project.

We found that adding an assembly attribute (compared to class attribute) is about 3x faster (as you don't have to loop through all the classes with reflection).

Example:
```
[assembly: Routing(typeof (FlightViewModel), @"foo://flights/(?<index>0|1)")]
[assembly: Routing(typeof(FlightRoutingFacade), @"foo://flight/\?id=(?<id>\d{8}[A-Z0-9]{2,3}\d{2,5})$")]
[assembly: Routing(typeof(FlightRoutingFacade), @"foo://flight/\?id=(?<id>\d{8}[A-Z0-9]{2,3}\d{2,5})&update=(?<update>true|false)$")]
```

[...]

PR #1595, opened 2017-02-09
Move routing to mvvmcross source

martijn00:
Fix for: https://github.com/MvvmCross/MvvmCross/issues/1315

@charri can you review this?


kjeremy:
1. Should the `Regex`s be compiled?  They will potentially be used a lot.
2. Should we also accept `Uri` objects since navigation is based on uris?

kjeremy:
Furthermore: how does this interact with things like Android's deep linking : is there a relationship between the two? Can we make one?

kjeremy:
As for my question above I see in the example that you use linking via an `IntentFilter` so awesome.

PR #1596, opened 2017-02-09
Add lifecycle support to the IMvxViewModel and MvxViewModel

mvanbeusekom:


kjeremy:
Please fix your indentation.

How do you expect to use these?  When do they get called?  How do they interact with our existing lifecycle events? 

Does this tie into a specific issue?

PR #1602, opened 2017-02-11
Fix forms test projects > Example001CSharp

ehuna:
Fixed the nuget paths on all projects in TestProjects\Forms\Example001CSharp

Example001CSharp > fix PCL nuget paths
Example001CSharp > fix Android nuget paths
Example001CSharp > fix Windows Phone nuget paths
Example001CSharp > fix Windows 8.1 nuget paths
Example001CSharp > fix iOS nuget paths

This allowed me to compile without errors - I was able to also run and test out the Android and Windows Phone test apps. 

This is my first pull request for MvvmCross - once it's merged I'll work on the other test forms projects.

Good times!
@ehuna

### DIFF
--- a/TestProjects/Forms/Example001CSharp/Example.Android/Example.Android.csproj
+++ b/TestProjects/Forms/Example001CSharp/Example.Android/Example.Android.csproj
@@ -46,58 +46,58 @@
     <Reference Include="Mono.Android" />
     <Reference Include="Mono.Android.Export" />
     <Reference Include="MvvmCross.Binding, Version=4.0.0.0, Culture=neutral, PublicKeyToken=null">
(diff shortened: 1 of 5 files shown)


martijn00:
First of all, thank you very much for putting effort into this! Unfortunately this change is not right because the nugets are already restored to the correct folder, which is the root. The reason that this didn't build for you is that you used the .sln file in the sample itself to run. I think we should remove those solution files and add it to the root.

We do this so we can debug the Forms samples from the MvvmCross code base. If you open the MvvmCross_All.sln file you can run it already.

Thanks again! If you want you could make a PR which removes the sln files.

ehuna:
Ah that makes sense - the fact that I could open the standalone test solutions but the projects would not build confused me.

I'll ensure I can run the test projects using the  MvvmCross_All.sln and I'll delete the individual solution files accordingly.

Thanks!

Issue #1594, opened 2017-02-09
Improving MvxTargetBindingFactoryRegistry


Nickolas-:
Hi everybody.

What do you think about change MvxTargetBindingFactoryRegistry method 

`priavte static string GenerateKey(Type type, string name)`

to

`private static int GenerateKey (Type type, string name)`

and use there just GetHashCode and not allocate each time strings?  In my case when you need to create a lot of target bindings it allocate for about 40 mb memory.

**Version:** 4.4.0

**Platform:** Core

kjeremy:
Please submit a pull request for this.

PR #1603, opened 2017-02-11
Improve key generation

Nickolas-:
Hello,

Related to https://github.com/MvvmCross/MvvmCross/issues/1594

It always better works with value types.


kjeremy:
LGTM.

In #1594 you said that 40MB of strings are allocated.  Does this PR drop that to 0?

Nickolas-:
It significant improve my application allocation , tested with profiler. It helps to not fill nursery so quickly. I can't give exactly numbers of allocations amount right now , because every run is unique.

PR #1605, opened 2017-02-12
Added destroy lifecycle for viewmodel

Nickolas-:
It very good to handle view lifecycle appearing and disappearing. But it more important to handle when view is no more needed and we can clean viewmodels references etc.

Yes , I know we can always do subscribe on appear and unsubscribe on disappear, but it not enough.Sometimes when next View change some behaviour , we need to tell previous that something has been changed, that why we can't unsubscribe on previous view disappear and still need that action.

ViewModel.Destoy () will always give us event , that we don't need anymore this ViewModel and can clean up.

### DIFF
--- a/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.RecyclerView/ItemSources/MvxRecyclerViewItemsSourceBridge.cs
+++ b/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.RecyclerView/ItemSources/MvxRecyclerViewItemsSourceBridge.cs
@@ -23,7 +23,7 @@ namespace MvvmCross.Droid.Support.V7.RecyclerView.ItemSources
             _subscription?.Dispose();
             _subscription = null;
 
-            _itemsSource = itemsSourceEnumerable ?? Enumerable.Empty<object>();
+			_itemsSource = itemsSourceEnumerable ?? Enumerable.Empty<object>();
 
             if (_itemsSource != null && !(_itemsSource is IList))
                 MvxBindingTrace.Trace(MvxTraceLevel.Warning,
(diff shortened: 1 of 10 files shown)


martijn00:
Do you also have implementation for this on Android and Windows? And can you fix your indentation?

Nickolas-:
I don't use Windows platform. (Just have a mac osx)
For Android i'm using OnDestroy on Activities and Fragments, for MvxRecyclerViewHolder using Dispose(true); (Checkis if DataContext is IMvxViewModel) but this is just my approach that's why maybe you want something else.

What about indentation it looks on the same im XS , I don't know what is going on.

martijn00:
Could you add your code for Android? I can do the Windows part after that.

It seems you use tabs, and we use spaces.

Nickolas-:
Ok, will do it today ,but a bit later.
PS: you was right with indentation.

Nickolas-:
So I added Android view's events , but there are one thing that i didn't change.
`MvxRecyclerViewHolder` view things are complicated.  We can do it in two ways, first we can add logic in `MvxRecyclerViewItemsSourceBridge` when we are setting `null` after detached `MvxRecyclerView` we can check if `DataContext` is `IMvxViewModel` and loop all viewModels and invoke Destroy method. 
The second approach we invoke `Destroy` on `Dispose(true)` like you do right now `_bindingContext?.ClearAllBindings()` but potentially user's can do some long running staff in method `Destroy()` and this will increase GC collection time. 

Maybe you have some others ideas or we need to create some mechanism that we clean up all `MvxViewHolder's` after detached `MvxRecycleView`

martijn00:
Personally i feel for going for the second approach right now. Can you also add the 4 other lifecycle events to Android to all those classes?

Nickolas-:
There are some challenges about viewAppearing and viewDisappearing, right now I'm not using them at all. There are different thoughts about this , we need get some more information.

kjeremy:
There's no reliable place that I could find to do destruction event hooks for ViewModels attached to `MvxRecyclerView`. Previously I tried to keep a `WeakReference` map to `MvxViewHolder`s but that broke things for nested `MvxRecyclerViews`.  Clearing the views is the closest we can get.  That is our current approach and it will most likely fail if we share RecyclerView adapters (which is totally legal on Android).  As an aside I wouldn't worry about the cost of a GC.

PR #1610, opened 2017-02-17
Generic implementation of IMvxTargetBinding

mvanbeusekom:
Adds a generic version of the MvxTargetBinding class. This allows for a type safe and easier way to implement custom target binding.

Also included in this PR is a iOS and Android test project, called "CustomBinding", to prove that the generic implementation is working correctly.


tofutim:
Hi, I'm on Mac, and for some reason this doesn't work for me.

https://github.com/MvvmCross/MvvmCross/issues/2449

SamuelDebruyn:
I have a question about the following code:

```
        protected virtual TValue MakeSafeValue(TValue value)
        {
            var safeValue = (TValue)TargetType.MakeSafeValue(value);
            return safeValue;
        }
```

Doesn't this assume that the `TValue` has to be the same type as the `TargetType`? What if I want to bind a string (with a color inside) to the background color of a view? Then the `TargetType` is `Android.Views.View` and `TValue` is `string`.

@martijn00 @mvanbeusekom

softlion:
@SamuelDebruyn It's a target binding, to the target binding class is responsible for using TValue on the correct target control property. If you need a converter between the viewmodel property and the target property, use WithConversion()

Issue #1614, opened 2017-02-20
Add id parameter to MvxTabPresentation


MarcBruins:
For Xamarin UITest we would like to set a "Accessibility identifier" on a tab bar menu item so we can safely navigate to another menu item. Currently this is not supported in the MvxTabPresentation. Ah solution would be something like this:

` [MvxTabPresentation(MvxTabPresentationMode.Tab, "News", "MY_ID", "ic_reorder_36pt", true)] 
`

This is feature request for the iOS MvxTabPresentation and not a bug.

Cheesebaron:
I think we should simply have a default which generates it from the name. So in your case: NewsTab. Then also have an option to set your own.

Not sure if we need to do anything about cleaning the name when assigning to Accessibility ID.

MarcBruins:
That sounds like a better solution @Cheesebaron

PR #1617, opened 2017-02-21
Add AccessibilityIdentifier to Tabs

nmilcoff:
The id can be set using the attribute over the class. If it is not defined, the default value will be "{Name of the Tab}Tab".

Solves #1614


MvvmCross-iOSSupport/MvvmCross.iOS.Support/Presenters/MvxTabBarPresenter/IMvxTabBarViewController.cs
```diff
@@ -5,13 +5,13 @@
 
 namespace MvvmCross.iOS.Support.Presenters
 {
-	public interface IMvxTabBarViewController
-	{
-		void ShowTabView(UIViewController viewController, bool wrapInNavigationController, string tabTitle, string tabIconName);
+    public interface IMvxTabBarViewController
+    {
+        void ShowTabView(UIViewController viewController, bool wrapInNavigationController, string tabTitle, string tabIconName, string tabAccessibilityIdentifier);
```
nmilcoff: Sure

MvvmCross-iOSSupport/MvvmCross.iOS.Support/Views/MvxBaseTabBarViewController.cs
```diff
@@ -52,8 +52,10 @@ public virtual void ShowTabView(UIViewController viewController, bool wrapInNavi
 			ViewControllers = currentTabs.ToArray();
 		}
 
-		protected virtual void SetTitleAndTabBarItem(UIViewController viewController, string title, string iconName)
+		protected virtual void SetTitleAndTabBarItem(UIViewController viewController, string title, string iconName, string tabAccessibilityIdentifier)
 		{
+			viewController.View.AccessibilityIdentifier = tabAccessibilityIdentifier;
```
nmilcoff: Yes, I probably should also move that line to be called directly from `ShowTabView`

PR #1626, opened 2017-02-25
Removed redundant register of android fragment presenter

Plac3hold3r:
As of PR #1495 the Android presenter will be automatically registered in the [MvxAndroidSetup ](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Droid/Droid/Platform/MvxAndroidSetup.cs#L126)


martijn00:
Please use the same indentation.

Plac3hold3r:
Previously the file was using a mix of tabs and spaces (mostly tabs). I believe MvvmCross repository convention is to use spaces? So I also re-formatted the file in this pull request.

image

Would you still prefer a PR with just the line removed?

kjeremy:
All spaces. All the time :)

PR #1631, opened 2017-02-26
Add generic WithConversion method

mvanbeusekom:
This change will add a generic "WithConversion" method. This will allow developers to strongly type the use of value converters, making refactoring a lot easier and more save. For example:

`set.Bind(textField).To(vm => vm.Counter).WithConversion<SomeValueConverter>();`

Note: I have made the FindName method public (from protected) and added it to the IMvxValueConverterRegistryFiller interface. This allows me to reuse the same code to determine the name of the value converter based on the type.


Cheesebaron:
Great stuff Maurits!

PR #1633, opened 2017-02-27
Solve bug where auto replacement is not updated through the binding

### LINKED ISSUE #1555: iOS Autocorrect UITextView not captured
### Steps to reproduce
(Assuming English with autocorrect turned on)

1. In a bound UITextView write "inages" then click off the field (maybe into different field?).

2. "inages" is autocorrected to "images"

### Expected behavior
The bound view model data is: images

### Actual behavior
The bound view model data is: inages

### Configuration

**Version:** 4.4.0

**Platform:** iOS 

### Comments

This is because in MvxUITextViewTextTargetBinding.cs the code very reasonably does this:
target.Changed += this.EditTextOnChanged;

Surprisingly you won't get a notification if the data is changed by autocorrect!

### Fix

Here is how RxSwift fixed it: https://github.com/sergdort/RxSwift/commit/c555862a67c524de5437fb64e0f666ccc08e82b7

TL;DR:  You have to use the https://developer.xamarin.com/api/type/MonoTouch.UIKit.NSTextStorage/ backing store and use the NSTextStorageDelegate to listen for changes.

Crazy but true!

mvanbeusekom:
Instead of listening to the `TextChanged` event the binding will now listen to the `DidProcessEditing` event of the NSTextStorageDelegate instance associated with the UITextView class. This will make sure that auto correct replacements are also caught and will trigger a change event in MvvmCross. This PR fixes #1555

### DIFF
--- a/MvvmCross/Binding/iOS/Target/MvxUITextViewTextTargetBinding.cs
+++ b/MvvmCross/Binding/iOS/Target/MvxUITextViewTextTargetBinding.cs
@@ -46,7 +46,16 @@ namespace MvvmCross.Binding.iOS.Target
                 return;
             }
 
-            target.Changed += this.EditTextOnChanged;
+			var textStorage = target.LayoutManager?.TextStorage;
+
+			if (textStorage == null)
+			{ 
+			    MvxBindingTrace.Trace(MvxTraceLevel.Error,
+						  "Error - NSTextStorage of UITextView is null in MvxUITextViewTextTargetBinding");
+				return;
+			}
+
+            textStorage.DidProcessEditing += this.EditTextOnChanged;
             this._subscribed = true;
         }
 


MvvmCross/Binding/iOS/Target/MvxUITextViewTextTargetBinding.cs
```diff
@@ -46,7 +46,16 @@ public override void SubscribeToEvents()
...
+			var textStorage = target.LayoutManager?.TextStorage;
+
+			if (textStorage == null)
+			{ 
+			    MvxBindingTrace.Trace(MvxTraceLevel.Error,
+						  "Error - NSTextStorage of UITextView is null in MvxUITextViewTextTargetBinding");
+				return;
+			}
+
+            textStorage.DidProcessEditing += this.EditTextOnChanged;
```
Cheesebaron: This event is marked Deprecated in the Apple documentation: https://developer.apple.com/reference/uikit/nstextstoragedelegate

Cheesebaron: It is strange though, looking deeper into the documentation, it could seem like it is only deprecated for macOS: https://developer.apple.com/reference/objectivec/nsobject/1532645-textstoragedidprocessediting

mvanbeusekom: There is a bit of confusion there I think, there used to be a method `textStorageDidProcessEditing` which is deprecated and replaced with a new class `NSTextStorageDelegate` which has a `DidProcessEditing` function. The later is the one I am using in the PR.

Cheesebaron: I find it strange that it is still marked deprecated, but not fully in the first link I provided, which seems to be the correct documentation for this notification?

Although, creating a implementation in C# of NSTextStorageDelegate, it doesn't look like it is deprecated there.

🤔

mvanbeusekom: The way I read the [documentation](https://developer.apple.com/reference/uikit/nstextstoragedelegate) (although I am also a bit confused) is that the methods listed under "Handling Text-Edited Messages" are the currently supported methods and replace the methods listed under the "Deprecated" header.

Also the Xamarin [documentation](https://developer.xamarin.com/api/type/MonoTouch.UIKit.NSTextStorage/) doesn't mention anything about the events being deprecated.

Issue #1637, opened 2017-03-02
RecyclerView Header/Footer crash on tap when using custom Header/Footer DataContext


thefex:
This issue applies to Header/Footer functionallity.
If we assign custom DataContext (for ex. during MvxViewHolderBound event) then click on Header/Footer crash application.

Working on fix.

PR #1638, opened 2017-03-02
Fixes crash when using custom DataContext of RecyclerView Header/Footer.

thefex:
Fixes issue #1637.


MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.RecyclerView/MvxRecyclerViewHolder.cs
```diff
@@ -136,9 +144,10 @@ private void OnItemViewOnLongClick(object sender, View.LongClickEventArgs args)
             this.ExecuteCommandOnItem(this.LongClick);
         }
 
-        public MvxRecyclerViewHolder(View itemView, IMvxAndroidBindingContext context)
+        public MvxRecyclerViewHolder(View itemView, int viewType, IMvxAndroidBindingContext context)
```
kjeremy: Would it make more sense to switch the order of `viewType` and `context` and default `viewType to 0?

thefex:
@kjeremy right, pushed your suggestions, rewrote history.

PR #1639, opened 2017-03-05
Add binding for View alpha

kjeremy:


Cheesebaron:
Does this even need a binding? Alpha is a public property so should be possible to bind directly?

kjeremy:
Good call. I just assumed I needed it. I'll test it out.  Don't merge yet.

kjeremy:
Yeah this is completely unnecessary.

PR #1656, opened 2017-03-25
Spit out error messages when conversions go wrong. Fixes #1655

### LINKED ISSUE #1655: Android : Exceptions in MvxValueConverter<TFrom, TTo>.Convert are hidden
When the Convert method of an implementation of MvxValueConverter throws an exception, no error is logged, and nothing appears in the log.

The crash is handled silently somehow.

It should log an error to the android log and to the Mvx.Error log instead.

Cheesebaron:


Cheesebaron:
@softlion does this look good to you?

softlion:
yep

PR #1657, opened 2017-03-26
Use new non RC VS2017 image for build

Cheesebaron:
DO NOT MERGE YET


Cheesebaron:
Should be good to merge, but AppVeyor doesn't seem to have Xamarin tools installed for the new image 😢 

https://github.com/appveyor/ci/issues/1448

martijn00:
It think we could just merge this for now, because the VS 2017 RC builds don't even exist anymore. Maybe if we are lucky AppVeyor fixes this in an update.

PR #1658, opened 2017-03-26
Improve MvvmCross.StarterPack

willsb:
Still WIP


martijn00:
How's this going? :) We would love to add this to the release of MvvmCross 5.0 in about two weeks.

nmilcoff:
I think the iOS View file should reflect the changes from the new MvxIosViewPresenter: MainView should have a MvxRootPresentationAttribute (like [this](https://github.com/MvvmCross/MvvmCross/blob/develop/TestProjects/Playground/Playground.iOS/Views/RootView.cs#L12))

willsb:
@martijn00 I'll need someone with a Windows machine to help me with the WPF Views. The rest I can complete early next week, so this can most likely enter the release :)

@nmilcoff That's all that I need to change? if so, consider it done :)

nmilcoff:
@willsb if there is only one view then yes, that's all :)

martijn00:
How's this going? @willsb

willsb:
@martijn00 According to [this](http://blog.nuget.org/20160126/nuget-contentFiles-demystified.html) blog post, the new `contentFiles` feature requires `project.json`, so the multi-language support will have to wait until we migrate to netstandard (which, according to the Roadmap, will only happen for 6.0). I'll close this PR (because it has too many changes that'll only be useful in the future) and open a new one containing only the essential stuff.

PR #1660, opened 2017-03-27
Added ContainsImage to IMvxImageCache and added support for android.resource schema in MvxAndroidLocalFileImageLoader.

Cybrosys:
Added ContainsImage to IMvxImageCache so it's possible to check if a Image is cached or not. In my specific case, I only want to do a fade-in animation when the image isn't in the cache.

I also added support for the android.resource schema in the MvxAndroidLocalFileImageLoader.


Cybrosys:
Was hoping on creating a new pull request, but being a github newb, it seems to have updated this one with more changes.

I'll update this PR's description etc.

martijn00:
You need to create a new branch to open a new PR. But the changes look good!

Issue #1576, opened 2017-01-25
Move APIExampes from MvvmCross-Samples over in main repo


Cheesebaron:
There are loads of great examples in the APIExamples project in the samples repo.

I would like to see it moved over here and expand it with isolated samples of features added to MvvmCross, to easily test out when debugging stuff.

ghuntley:
I'm conflicted (personally) on this one. For RxUI I'm considering :burning: all of the playground/samples in the main repository and instead doubling down on a separate repository - https://github.com/reactiveui/samples (currently empty). Then doubling down on CI (as in integration) tests that run the samples against the nuget packages generated to verify that there are no packaging defects and also to run e2e tests on simulator or on Xamarn Test Cloud. Will be interested in seeing how this progresses in Mvx land.

Cheesebaron:
It is just as much easing the job of testing the code while working on MvvmCross, it can be a bit painful to create a new project and add references to relevant parts and set up all the boiler plate.

nmilcoff:
There are currently __a lot__ of TestProjects in the main repo. I think we should gradually move into a single one (or maybe two, if we want to keep a separate one for X.Forms).

I suggest these milestones (not considering Xamarin Forms samples here):
1) Move APIExamples into main repo

2) TestProjects reduction: Keep only APIExamples and Playground (good for testing navigation and presenters). All other projects should be integrated into these two.

3) Unify Playground and APIExamples into a single, big TestProject.

What do you guys think @MvvmCross/core?

mvanbeusekom:
I am in doubt, when I need to learn something or looking for a particular sample I prefer the project to be small and focused on the particular topic (no distractions). The same goes when the projects are only meant as test projects. The bigger they get the harder it becomes to quickly test functionality. Second it might also be difficult to merge all functionality into one project (although this might be a sign the we need to improve certain things). A good example would be combining the Sidebar example with the other iOS navigation structures, this will be difficult in the current implementation since they both require a different presenter. 

Then I also get that the maintenance is hard when having a lot of projects. And as mentioned earlier with a bigger project it is easier to test if different functionalities play nice together. So I am kinda in a limbo ;)

nickrandolph:
@Cheesebaron do we still want to do this? 
Personally I think there are samples that are in the main repo that need to be upgraded and added back to the main solution, but samples that are in the samples repo should probably just stay there and have some tasks raised to get them maintained. Thoughts?

Issue #1415, opened 2016-08-23
MvvmCross roadmap


martijn00:
Proposal for MvvmCross 5.0 release:
- [x] CI for builds (#1301)
- [x] Automated release builds and nightly builds (#1301)
- [x] WP Silverlight removal (#1460)
- [x] WP8.1 & WindowsStore removal
- [x] AutoView/CrossUI/Dialog removal
- [x] NavigationService with url / deeplink navigation (#1315) (#1595)
- [x] Reproduction samples in main repo
- [x] New event hooks for View > ViewModel loading (#1531 #1601)
- [x] Refactored iOS presenter (#1274) (#1671)
- [x] New website for blog and documentation (#1686)
- [x] Add improved support for Xamarin.Forms

Proposal for 5.x
- [x] More NavigationService improvements
- [x] New Android presenter with support for Dialogs (https://github.com/MvvmCross/MvvmCross/issues/1934)
- [x] New MacOS and tvOS presenters (https://github.com/MvvmCross/MvvmCross/pull/1913)
- [x] Improved first View / ViewModel in StartPack nuget (#1658)
- [ ] Reloading and caching of Views and ViewModels like Fragments (with data) (https://github.com/MvvmCross/MvvmCross-AndroidSupport/issues/194)
- [ ] Handle Android permissions (https://github.com/MvvmCross/MvvmCross-Plugins/issues/82)
- [ ] No dependency on splash screen for loading app

Proposal for 6.0
- [ ] .NET Standard (#1396)
- [ ] Use Bait & Switch with PCL/.NET Standard for plugins (https://github.com/MvvmCross/MvvmCross-Plugins/issues/49)
- [ ] More Async (depends on platform) (#1136)
- [ ] Make it more easy to inherit from base classes when adding new view / class types

Work on MvvmCross 5 will happen on the `develop` branch. From now on we will use this as our main branch, and use the 4.0 and 3.x branches for bugfix releases(if necessary).

Of course we could use your help so get ready to jump in! Please provide feedback so we can get our requirements ready for this release!

Cheesebaron:
I think a better structure for working is. Ditch the 4.0 and 3.x branches entirely. Make a develop branch where magic happens. Then push releases and tags into master from develop.
I am not a big fan of having silly LTS support in a OSS project, where resources to begin with are sparse. If that is something you want to do, go ahead and do it in a fork or something.

[...]

Issue #1634, opened 2017-02-28
NavigationService improvements


martijn00:
The first version of a new navigation service should support:
- [x] Async await enabled
- [x] Return results
- [x] Type save
- [x] Deeplinking / URI navigation enabled
- [ ] CanNavigate check
- [x] Events on navigation steps

The aim for the second update would be
- [x] Bug fixes for first version
- [x] Unit Tests
- [ ] Nested routed navigation
- [x] Support for dialogs / modals
- [ ] Fully async await
- [x] Type safe all the way

Secondary requirements are:
- [ ] NavigationAware, OnNavigatedTo, OnNavigatedFrom
- [ ] Nested ViewModels / ChildViewModels
- [ ] ViewModel backstack
- [ ] Central Attributes for navigation patterns (#1671)
- [ ] Override navigation from specific view
- [ ] Animations
- [ ] Optional: View first navigation support too
- [x] Cancellation support

Other remarks
- [x] ShowViewModel stays for 100% compatibility
- [x] Testability
- [x] Support for Xamarin Forms

Proposal for new interface:
```c#
public interface IMvxNavigationService
{
Task Navigate<TViewModel>() where TViewModel : IMvxViewModel;
Task Navigate<TViewModel, TParameter>(TParameter param) where TViewModel : IMvxViewModel;
Task<TResult> Navigate<TViewModel, TParameter, TResult>(TParameter param) where TViewModel : IMvxViewModel;
Task<TResult> Navigate<TViewModel, TResult>() where TViewModel : IMvxViewModel;
Task Navigate(string path);
Task Navigate<TParameter>(string path, TParameter param);
Task<TResult> Navigate<TResult>(string path);
Task<TResult> Navigate<TParameter, TResult>(string path, TParameter param);
Task<bool> CanNavigate<TViewModel>() where TViewModel : IMvxViewModel;
Task<bool> CanNavigate(string path);
Task<bool> Close(IMvxViewModel viewModel);
}

public static class MvxNavigationExtensions
{
public static Task<bool> CanNavigate(this IMvxNavigationService navigationService, Uri path)
public static Task Navigate(this IMvxNavigationService navigationService, Uri path)
public static Task Navigate<TParameter>(this IMvxNavigationService navigationService, Uri path, TParameter param)
public static Task<TResult> Navigate<TResult>(this IMvxNavigationService navigationService, Uri path)
[...]

PR #1671, opened 2017-04-01
New MvxIosViewPresenter for iOS

### LINKED ISSUE #1274: MvxModalNavSupportIosViewPresenter does not use CurrentTopViewController
### Steps to reproduce
1. Use MvxModalNavSupportIosViewPresenter
2. Show a MvxViewController that inherits from IMvxModalIosView
3. Show a MvxViewController without IMvxModalIosView
### Expected behavior

The final show should use the NavigationController that contains the IMvxModalIosView to Push/Show the normal MvxViewController (CurrentTopViewController), not the MasterNavigationController. A push modal should allow for a new horizontal navigation. Preferable a way to close the modal as well, collapsing the whole navigation chain.

[Presenters in MvvmCross: Nested Modal Navigation in iOS](http://gregshackles.com/presenters-in-mvvmcross-nested-modal-navigation-in-ios/)
### Actual behavior

No navigation happens after showing a IMvxModalIosView and trying to Push/Show a normal MvxViewController.
### Configuration

**Version:** 4.1.0

**Platform:** iOS

nmilcoff:
This PR aims to resolve #1274. It takes the work done by @PelleRavn. This is also part of #1415. 

I have created a brand new view presenter, which means this won't be a breaking change (unless we set it as the default presenter).

Main features of this presenter are:
* Support for stack navigation 
* Support for nested stack navigation (modals)
* Support for modals _not_ wrapped in Navigation Controllers (`IMvxNavModalIosView` and `IMvxModalIosView`)
* Ability to display a "child" ViewModel in the current stack without restrictions
* Ability tu customize the way a modal view is displayed (animation, presentation style and transition style trough `MvxModalDisplayStyleAttribute` (class attribute)
* Support for modal ViewControllers that are presented manually by the user outside the navigation system of MvvmCross

This PR also includes
* Creation of `MvxNavigationController`, from suggestions of the referenced issue
* Creation of a Playground test project for the presenter: Following #1576 we should consider having a single playground for testing the "core" functionalities of the framework. CustomBinding sample and this one could be merged into one. Or we can bring APIExamples here, as @Cheesebaron sugested

-----

Current Status based on discussion and refactoring proposal:
- [x] Add MvxNavigationController
- [x] Add MvxSplitViewController
- [x] Add custom attributes for different presentation modes
- [x] Add support for navigation stack
- [x] Add support for Tabs navigation
- [x] Add support for _modal_ navigation stack
- [x] Add support for modals without navigation stack
- [x] Add support for SplitView as root
- [x] Remove interface `IMvxModalIosView`
- [x] Remove TabsPresenter from iOS-Support
- [x] Remove old presenters (`MvxModalNavSupportIosViewPresenter` and `MvxModalSupportIosViewPresenter`)
- [x] Remove parameter `wrapInNavigationController` from `ShowTabView` in `IMvxTabBarViewController`: Now that configuration is handled in the presenter
- [x] Add MvxSplitViewContainer
- [x] Complete refactor of `MvxIosViewPresenter`
- [x] Merge Tabs sample with Playground
- [x] Add close functionality (pretty hard to implement due to quantity of situations this presenter can handle)
- [x] ~~Add support for SplitView as child of a stack~~ : _Excluded by now, will do in a separate PR to reduce complexity._
- [x] ~~Support modal ViewControllers that are presented manually by the user outside the navigation system of MvvmCross~~: _Excluded, users are encouraged to display all views inside the MvvmCross navigation mechanism. If an app needs a different behavior, then it should use another / override this one._


martijn00:
Fantastic initiative! I few things that i would like to add as goals for this refactoring:

* I would like to introduce a global system with View attributes for the iOS presenters. Currently the iOS tab presenter and sidemenu presenter already use them. This could be unified and for example for the sidemenu presenter, we would ideally only override a method where we control the attribute usage. We could also add the modal presentation to the attribute.
* To do this we can pull in the attributes to the main libs, and just override the new presenter in the iOS support package.
* The tabs are a standard part of iOS. I think we should support this in the default presenter.
* Can we support Segmented control navigation?

In my opinion this should be used as the new default presenter, and all others should be deprecated when the new one supports those use cases. If we do this for 5.0 it would be fine by me to break some behavior.

What do you think of those things?

nmilcoff:
I really like the idea of attribute-based presenters. It keeps the code clean and extensible.

Here is a proposal for refactoring:
* Add support for Tabs to this presenter (we can potentially add support for SplitView too, but the more we add the more it will take to finish it)
* Create an iOS global attribute for presentation with common configuration members (presentation mode: tab, child, root, modal, ...) 
* Create separate attributes that inherit from a base one, to handle specific configurations for each presentation mode (for tabs: tabTitle, tabIconName, ... / for modals: PresentationStyle, TransitionStyle, ...)
* Both attributes will be grabbed in overridable methods from a base presenter 
* Set this presenter as the default presenter. This should be well documented when releasing 5.0

Segmented control navigation won't be so easy to include, since it doesn't work by itself as a navigation control (it can work together with a PageViewController for instance, but the SDK doesn't provide any automatic way to configure it). We can add a control and a featured presenter to support this on the iOS-Support package, but I would consider this out of the scope of this PR.

Async support (and showing-for-result) are super interesting to add too, but I guess both are related to #1634 and should be added during the creation of the new navigation system

nmilcoff:
I have started the work on the refactoring. I've updated the PR description with the current status.

The original idea of having two separate attributes to handle common / specific configuration for each presentation mode has changed into a single attribute for each presentation mode (MvxTabPresentationAttribute, MvxRootPresentationAttribute, ...) that has all its configuration inside it.

Reason for this is that otherwise the "common configuration" attribute would have an enum that defines a PresentationMode... And enums cannot be extended. As a result, the presenter wouldn't be flexible at all.

PelleRavn:
I've only looked at it very briefly, but it looks very promising. I'll have a deeper look at it, and try it out soon.

One thing that everyone (including myself) quietly ignores is the "clear stack" implementation. It's a complicated thing to do (my current implementation is a little hack-ish), but should we maybe find a way to do this as well?

nmilcoff:
The PR is now ready to be reviewed.

It can be seen in action running the Playground project (you can customize the attributes over the View classes). It supports stack, tabs, split and modal navigation patterns. It is extensible, flexible and I hope you find it readable :)

nmilcoff:
I just realized this change makes others iOS TestsProjects crash in runtime because of missing presentation attributes. Will commit this

PelleRavn:
Wouldn't it be better if a view with no Attribute would be treated as MvxChildPresentationAttribute, so it's only views who behaves differently that needs a specific Attribute, instead of every single view needs one? Then it's a little more logical to people, than it just crashing asking them to please remember an attribute?

martijn00:
I think if a view has no attribute the behavior should be the same as it is right now. If that means it's the `MvxChildPresentationAttribute` we should use that as default.

nmilcoff:
Good suggestion! It is done now. In Playground project, `SecondChildView` has no attribute.

martijn00:
Would it be an idea to use more default values for the attribute's? Like saying:
```
public bool WrapInNavigationController { get; set; } = true;
public UIModalPresentationStyle ModalPresentationStyle { get; set; } = UIModalPresentationStyle.FullScreen;
```

If we make some assumptions what is used most, it is easier to implement, while you can still set your own way if you want.

nmilcoff:
I definitely agree with that. I'll add more default values to the attributes.

>Maybe you can improve the sample by adding a way to go back to the first screen, so you can choose a new navigation pattern to test.
Currently it is possible to initiate stack navigation again when Tabs are root, I'll add a way to do it when in Split is root too.

**Important note**, just in case you didn't notice: Once the root is changed the old root is dismissed (otherwise ViewControllers and all resources won't be disposed). There is no way to _go back_ to a previous **root**, only to navigate forward to a new instance of it.

I was also thinking about getting rid of the word `Presentation` in the attributes, since the context already gives an idea what the attribute will do. What do you guys think about this?

New names would be:
`MvxRootAttribute`
`MvxChildAttribute`
`MvxTabAttribute`
`MvxModalAttribute`
`MvxMasterSplitAttribute`
`MvxDetailSplitAttribute`

Only `MvxBasePresentationAttribute` remains the same.

martijn00:
I didn't notice the root is changed, but you are right, that is a good thing. Maybe then just add a button to navigate to the first view again.

I'm not sure if removing `Presentation` from the name makes it more clear.

PelleRavn:
I definitely don't think you should remove 'Presentation' in the names. I think it should be explicit and it also makes it easier to find them with the intellisense.

nmilcoff:
Thanks guys, it makes sense.

I've added some more default values and improved the sample renaming some buttons and adding a way to initiate stack navigation when SplitView is the root.

PR #1675, opened 2017-04-05
New MvxPageViewController, 2nd edition

chrilith:
Hi,

The is a new attempt to bring a new handling for PageViews in MvvmCross.
All details are in the #1407 which has been closed because it was completely messed up...

New PoC for iOS is [here](https://www.dropbox.com/s/s2v58z66ne9i1pa/MVVMCross_PageViewPoC.zip?dl=0). 

I've just commited a new implementation of the MvxPageViewController which works more like any other controllers in MvvmCross.

The current implementation, based on view models, is really interesting, but one thing that is hardly missing is binding (AFAICT) and a way to refresh the controller. Also, as stated in #1403 we are forced to use an internal cache, when iOS already cache things on its own.

Here, I added a simple data source that implements most of the minimal requirements to  properly handle PageViews.

I will add some events later if this PR is acceptable to the dev team. Don't want to waste too much time right now because i have an app to develop :)

_Added_:
- `MvxPageViewSource`: the important part here is `GetViewControllerAtIndex()` which will provide the expected view conroller if any and `GetPageIndexForController()`.
- `IMvxpageViewModel` and `IMvxPageViewController`. Both define à `PageIndex` that will be used by the default data source (`MvxPageViewSource.GetPageIndexForController()`)
- Some extensions to `MvxPageViewSource` to ease the creation of view controllers

Basically everything works with ViewControllers with iOS page views so I think it is important to keep this logic.

Let me know if you have any questions.
Thanks!


MvvmCross/Binding/iOS/Views/MvxPageViewSource.cs
```diff
@@ -0,0 +1,149 @@
...
+
+		protected MvxPageViewSource(UIPageViewController pageView)
+			: base(pageView)
+		{
+		}
+
+		protected MvxPageViewSource(IntPtr handle)
+            : base(handle)
+        {
+			Mvx.Warning("TableViewSource IntPtr constructor used - we expect this only to be called during memory leak debugging - see https://github.com/MvvmCross/MvvmCross/pull/467");
```
nmilcoff: `TableViewSource` should be `MvxPageViewSource` :)

Cheesebaron:
Your PoC doesn't run. Also please make either a separate repo showing off the functionality or add it in your PR in TestProjects.

There is a lot of code changed here, I need a way to actually verify this is working.

martijn00:
@chrilith Can you please update this PR and add the sample to a github repo or include a sample in the PR? We really want to get this in before 5.0.

chrilith:
Hi,

Yes sorry, was really busy. Will try to do this this week.

chrilith:
The branch has been updated.
Here is the git repo for the code example: https://github.com/chrilith/MvvmCross-PoC

PR #1680, opened 2017-04-09
Improvements on the Xamarin Sidebar component

mvanbeusekom:
Improvements on the iOSSupport - Xamarin Sidebar implementation. This PR makes sure that the component locates a left or right menu automatically. This means it is no longer needed to make a "ShowViewModel(...)" call to initialize the menu.

Also added additional configuration options. By implementing the IMvxSidebarMenu interface you can now set the following additional properties:

- AnimateMenu: this will animate showing or hiding the menu;
- DisablePanGesture: this will disable the option to open the menu with a swipe gesture (useful when you also use swipe gestures in a table view for example).

Note: these changes will break existing implementations of the Xamarin Sidebar component.


nmilcoff:
Can you add a property to `IMvxSidebarMenu` to set the [dark overlay](https://github.com/jdehlin/Xamarin-Sidebar/blob/576fe1292fcced355471ee58396ea421c55f8037/Xamarin-Sidebar/Sidebar.cs#L112) as well?

mvanbeusekom:
Good spot, will do

PR #1684, opened 2017-04-11
Remove (stale) JASidePanels projects from iOS-Support and TestProjects

### LINKED ISSUE #1682: Remove JASidePanels
Should we remove  the JASidePanels implementation? It is a wrapper around another package which we don't control with some issues in there. When someone ask about a hamburger menu in MvvmCross i always say that they need to use XamarinSideBar and not JASidePanels. XamarinSideBar gets updated regularly in MvvmCross and JASidepanels does not. Furthermore the same look that JASidePanels has (the topbar) can be easily implemented for XamarinSideBar as wel. So there are no benefits at all in using JASidepanels.

mvanbeusekom:
Removes JASidePanel implementation from iOS-Support and will resolve #1682


mvanbeusekom:
Agree, could this be something like a "Breaking Change" in the release notes of V5.0?

MarcBruins:
We can write this down in the MvvmCross 5 annoucment. Also a tutorial on how to switch from JASidePanels to XamarinSideBar would be nice and would not be that much work.

Issue #1670, opened 2017-04-01
Cleanup: Obsolete MvxFormFactorSpecificAttribute


nmilcoff:
I might be wrong, but it seems to me there are some classes out there about the iOS platform that are obsolete:

- [MvxFormFactorSpecificAttribute](https://github.com/MvvmCross/MvvmCross/blob/8a824c797747f74716fc64c2fd0e8765c29b16ab/MvvmCross/iOS/iOS/Views/MvxFormFactorSpecificAttribute.cs): This one went obsolete when all apps migrated to iOS Universal, take a look at [this PR](https://github.com/MvvmCross/MvvmCross/pull/916)
- [IMvxIosPlatformProperties](https://github.com/MvvmCross/MvvmCross/blob/8a824c797747f74716fc64c2fd0e8765c29b16ab/MvvmCross/iOS/iOS/Platform/IMvxIosPlatformProperties.cs) 
- [MvxIosPlatformProperties](https://github.com/MvvmCross/MvvmCross/blob/8a824c797747f74716fc64c2fd0e8765c29b16ab/MvvmCross/iOS/iOS/Platform/MvxIosPlatformProperties.cs)
-  [MvxIosDisplayDensity](https://github.com/MvvmCross/MvvmCross/blob/8a824c797747f74716fc64c2fd0e8765c29b16ab/MvvmCross/iOS/iOS/Platform/MvxIosDisplayDensity.cs)
- [MvxIosFormFactor](https://github.com/MvvmCross/MvvmCross/blob/8a824c797747f74716fc64c2fd0e8765c29b16ab/MvvmCross/iOS/iOS/Platform/MvxIosFormFactor.cs)

These are marked as obsolete, and there are some plugins to grab this info, like [this one](https://github.com/Cheesebaron/Cheesebaron.MvxPlugins/tree/master/DeviceInfo).

martijn00:
Can you make a PR to remove those and clean it up? As this might be breaking it would be good to do for 5.0.

PR #1691, opened 2017-04-15
Cleanup iOS platform code // Fix XamarinSidebar Presenter

nmilcoff:
This PR solves #1670.

While working on it I made some other changes:
- Fix compilation issue in PluginLoader of `MvvmCross.Plugins.All`
- Re-add deleted interface `IMvxModalIosView`: It was deleted in #1671 because it's not necessary anymore there. But the XamarinSidebar presenter uses it. This will be however a breaking change because of the namespace. We have three options here: 1) Change the namespace to the original one; 2) Let this be a breaking change; 3) Refactor the XamarinSidebar Presenter (will be of course a breaking change here).
- Fix compilation issue in `Eventhooks.Droid` sample (that project is completely empty, it doesn't even have MvvmCross installed!).
- Made method `Show(IMvxIosView view, MvxViewModelRequest request)` of MvxIosViewPresenter virtual.


martijn00:
We should make this a breaking change since this will go into 5.0 anyways. No point in keeping the interface then.

I think the whole sidebar presenter should be refactored to be used with the new attribute system, and inherit from the new base presenter so all other situations except for sidemenu are handled by the base.

@mvanbeusekom might have some thoughts about this one.

PR #1694, opened 2017-04-15
Refactor Sidemenu presenter

martijn00:
Continue the work of https://github.com/MvvmCross/MvvmCross/pull/1691

I think we should do some more renaming of all the panel stuff and clean up the leftovers from that.

We might also be able to use the base splitview implementation to handle the sidemenu on ipads.


MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresenter.cs
```diff
@@ -33,94 +32,77 @@ protected virtual bool PresentationHintHandler(MvxPanelPresentationHint hint)
...
-                view.RemoveFromSuperview();
+                Mvx.RegisterSingleton<IMvxSidebarViewController>(SideBarViewController);
+
+                CloseMasterNavigationController();
+                CloseModalNavigationController();
+                CloseSplitViewController();
 
-            MasterNavigationController = new UINavigationController();
+                return;
+            }
```
nmilcoff: I think there should be an `else` here that releases all references to `IMvxSidebarViewController`, since if the upper `if` is not true, it means the user is changing the root of the app

MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresenter.cs
```diff
@@ -33,94 +32,77 @@ protected virtual bool PresentationHintHandler(MvxPanelPresentationHint hint)
...
+            base.ShowRootViewController(viewController, attribute, request);
+        }
 
-            RootViewController = new MvxSidebarPanelController(MasterNavigationController);
-           
-            SetWindowRootViewController(RootViewController);
+        public override void Close(IMvxViewModel toClose)
+        {
+            // if the current root is a SideBarViewController, delegate close responsibility to it
+            if (SideBarViewController != null && SideBarViewController.CloseChildViewModel(toClose))
```
nmilcoff: Current implementation of `SidebarViewController.CloseChildViewModel` doesn't take into account the `toClose` parameter, maybe we can get rid of that method and just call `base.Close(toClose)` here. Otherwise make `CloseChildViewModel` just return false and mark it as virtual

MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresenter.cs
martijn00: Ok, but how do you handle that in the base presenter?

nmilcoff:
I'd suggest we can also simplify the way the presenter works. 

The hints can be removed and the `MvxPanelHintType` can be handled directly in the `ShowSidebarViewController` method (using protected virtual methods for each value instead of sending a request to change presentation).

From my understanding, hints are _made to_ / _encouraged_ to be used from shared code (aka Core).

MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresenter.cs
martijn00: I would just leave it as is, and fix that it's not taken in account in the next version.

MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresenter.cs
nmilcoff: After the else that releases everything related to `IMvxSidebarViewController`, `base.ShowRootViewController(viewController, attribute, request);` should still be called. That should be enough

MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresenter.cs
```diff
@@ -33,94 +32,77 @@ protected virtual bool PresentationHintHandler(MvxPanelPresentationHint hint)
...
-        public override void Show(MvxViewModelRequest request)
+        protected override void RegisterAttributeTypes()
         {
-            IMvxIosView viewController = Mvx.Resolve<IMvxIosViewCreator>().CreateView(request);
-            Show(viewController);
+            base.RegisterAttributeTypes();
+
+            _attributeTypesToShowMethodDictionary.Add(
+                typeof(MvxSidebarPresentationAttribute),
+                (vc, attribute, request) => ShowSidebarViewController(vc, attribute as MvxSidebarPresentationAttribute, request));
```
mvanbeusekom: I think you should make a direct cast here `(MvxSidebarPresentationAttribute)attribute`, or otherwise make a check and return if the attribute is not of type `MvxSidebarPresentationAttribute`.

Now there is a possibility for a `NullReferenceException` which will only occur in the `ShowSidebarViewController` method.

martijn00: Hmm i don't see how that could happen. The base presenter also does it this way.

martijn00:
Just merged now so @nmilcoff can fix some remaining issue's.

MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresenter.cs
nmilcoff: I think this is safe to do. The handler method is called from [here](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs#L91), based on the attribute type.

mvanbeusekom: Casting the attribute using the `as` keyword could result in a `null` value (meanly if the attribute is not of type `MvxSidebarPresentationAttribute` (which theoretically is possible). 

So either make a direct cast (which is better performance wise) and in case of wrong usage it will generate an exception on the correct place. Or if we don't want the application to crash we should do a check.

nmilcoff: Mmm sorry but I don't see how it is possible for the attribute to be not of type `MvxSidebarPresentationAttribute` here. If the attribute is of another type then another method will be called, and if the attribute is not registered in the dictionary, then a `MvxSidebarPresentationAttribute` will be [thrown](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs#L92).

Anyways, if a direct cast is faster we should change it.

mvanbeusekom: That's my point. I couldn't oversee if the attribute could be null or not. If it can't be null then we should do a direct cast. 

Besides the performance difference (which is minimal), the as operator still suggests it could be null, which could lead other developers on the wrong path (of which this is an example 😝)

PR #1695, opened 2017-04-15
Xamarinsidebar Presenter improvements

nmilcoff:
This PR continues the work done in #1694.

I've made the following changes:
- Removed hints from XamarinSidebar presenter and implemented that logic in `MvxSidebarPresenter` using virtual methods.
- Removed some hints that were used only in the JASidebarPanels presenter (which was removed in #1684).
- Changed namespaces to the corresponding new locations of some files (in MvvmCross.iOS.Support.XamarinSidebar).
- Renamed `MvxPanelHintType.ActivePanel` to `MvxPanelHintType.PushPanel` to correspond with what it actually does.
- Made method `CloseChildViewModel` of `MvxSidebarViewController` virtual.
- Changed `as` to direct cast in registering of attributes in MvxIosViewPresenter to clarify behavior.


mvanbeusekom:
Awesome work @nmilcoff

PR #1697, opened 2017-04-16
Make sure the correct default presentation attribute is returned

mvanbeusekom:
This change ensures the `GetPresentationAttributes` method returns the correct presentation attribute in case there is no presentation attribute found on the ViewController. In case there is no `MvxNavigationController` instance is available it will return a `MvxRootPresentationAttribute` otherwise a `MvxChildPresentationAttribute` is returned.

This will solve the problem where the `MvxException` (with the message "Trying to show View type: CenterPanelView as child, but there is no current Root!") is being thrown.

### DIFF
--- a/MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
+++ b/MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
@@ -64,7 +64,7 @@ namespace MvvmCross.iOS.Views.Presenters
         {
             base.ChangePresentation(hint);
 
-            if(hint is MvxClosePresentationHint)
+            if (hint is MvxClosePresentationHint)
             {
                 Close((hint as MvxClosePresentationHint).ViewModelToClose);
             }
@@ -88,7 +88,7 @@ namespace MvvmCross.iOS.Views.Presenters
             var attribute = GetPresentationAttributes(viewController);
 
             Action<UIViewController, MvxBasePresentationAttribute, MvxViewModelRequest> showAction;
-            if(!_attributeTypesToShowMethodDictionary.TryGetValue(attribute.GetType(), out showAction))
+            if (!_attributeTypesToShowMethodDictionary.TryGetValue(attribute.GetType(), out showAction))
                 throw new KeyNotFoundException($"The type {attribute.GetType().Name} is not configured in the presenter dictionary");
 
             showAction.Invoke(viewController, attribute, request);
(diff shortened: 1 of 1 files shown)


MvvmCross/iOS/iOS/Views/Presenters/Attributes/MvxRootPresentationAttribute.cs
```diff
@@ -3,6 +3,6 @@ namespace MvvmCross.iOS.Views.Presenters.Attributes
 {
     public class MvxRootPresentationAttribute : MvxBasePresentationAttribute
     {
-        public bool WrapInNavigationController { get; set; }
+		public bool WrapInNavigationController { get; set; } = true;
```
nmilcoff: I'm not sure about making this member `true` by default, Because by doing so, when a user tries to set a `TabBarController` or a `SplitViewController` as root, he will have to set this to `false` explicitly

MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
```diff
@@ -371,12 +371,19 @@ protected virtual void SetWindowRootViewController(UIViewController controller)
...
             {
-                MvxTrace.Trace($"PresentationAttribute not found for {viewController.GetType().Name}, assuming animated Child presentation");
-                attributes = new MvxChildPresentationAttribute();
+				return attributes;
             }
-            return attributes;
+
+			if (MasterNavigationController == null)
+			{ 
+				MvxTrace.Trace($"PresentationAttribute nor MasterNavigationController found for {viewController.GetType().Name}. Assuming animated Root presentation");
```
nmilcoff: Root changes are not being animated currently 😬

MvvmCross/iOS/iOS/Views/Presenters/Attributes/MvxRootPresentationAttribute.cs
mvanbeusekom: I see that would not be nice indeed. I have added a new commit correcting this problem.

nmilcoff: Mhm but... the commit does not revert this change. `MvxRootPresentationAttribute` is the one that needs to be modified

mvanbeusekom: Sorry committed the wrong file. Will fix iT now

mvanbeusekom: Fixed, that was a stupid mistake.

mvanbeusekom:
Fixed the spacing issues, tried to squash changes but not sure it worked out correctly.

kjeremy:
Squash all of the commits into one (use `git rebase -i` to help) and then do a `git push --force`

mvanbeusekom:
commits are squashed

PR #1704, opened 2017-04-22
Speed up AddAssemby,  reduces startup time by at least 1 second

softlion:
Speed up AddAssemby (x10) which reduces startup time by at least 1 second on Android when using app compat libs.

Note that Mvvmcross should instead get rid of all its reflection code to obtain the best performance at startup time.

PR #1707, opened 2017-04-23
Speed up ioc type cache

Cheesebaron:
This is a clean version of #1704 where style and actual functional changes are not in same commit.

As mentioned in #1704 this improves the `AddAssembly` method significantly as the `LINQ` query here is much much faster.

### DIFF
--- a/MvvmCross/Platform/Platform/IoC/MvxTypeCache.cs
+++ b/MvvmCross/Platform/Platform/IoC/MvxTypeCache.cs
@@ -5,6 +5,7 @@
 //
 // Project Lead - Stuart Lodge, @slodge, me@slodge.com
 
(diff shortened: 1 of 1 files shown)


Cheesebaron:
What our `MvxTypeExtensions.ExceptionSafeGetTypes()` does is:

1. Call `ReflectionExtensions.GetTypes()`
2. This calls `assembly.DefinedTypes`
3. Yields all those, but calls `AsType()` on all these first

These steps are wrapped in a try/catch.

The query @softlion proposes is probably faster because of the type checking is done before `AsType` is called. I can't see why it should be faster otherwise than maybe using `yield` here imposes some overhead as well.

Cheesebaron:
I guess we should add the same try/catch here. However, I've never seen this exception throw before.

PR #1715, opened 2017-04-24
Fix solution release mode

martijn00:


Cheesebaron:
Not sure what is going on with the CI not picking up PR's.

I know it works if your branch is `features/*`

martijn00:
Hmmm, but does that also work from forks?

Cheesebaron:
You can try and see :)

PR #1716, opened 2017-04-24
update current navigation and move tipcalc explanation to the tutorial section

MarcBruins:
Updated the current navigation for the fundamentals page off the docs and move the TipCalc article to the tutorials section. The tutorials sections still needs to be sorted

### DIFF
--- a/docs/_documentation/fundamentals/navigation.md
+++ b/docs/_documentation/fundamentals/navigation.md
@@ -4,224 +4,55 @@ title: Navigation
 category: Fundamentals
 order: 3
 ---
-This article will cover some of the techniques available within MvvmCross for navigating between Page-level **ViewModels**.
+MvvmCross uses `ViewModel first navigation`. Meaning the we navigate from ViewModel to ViewModel and not from View to View. In MvvmCross the ViewModel will lookup its corresponding View. By doing so we don't have to write platform specific navigation and we can manage everything from within our core.
 
-### What do we mean by Page?
+## Simple ViewModel navigation
 
-MvvmCross was born for making modern Mobile apps - for building apps for iPhone, Android, and Windows Phone.
-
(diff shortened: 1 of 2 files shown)


docs/_documentation/fundamentals/navigation.md
```diff
@@ -4,224 +4,55 @@ title: Navigation
...
 
-MvvmCross was born for making modern Mobile apps - for building apps for iPhone, Android, and Windows Phone.
+To navigate from a ViewModel to another ViewModel you can use `ShowViewModel` command.
+The `ShowViewModel` command will take a `Generic` type which should represent the ViewModel that you want to navigate to.
 
-These apps are generally Page-based - that is they generally involve User-Interfaces which show a single page at a time and which often involve the user experience moving forwards and backwards through the application workflow.
-
-There are variations on this, especially for Tabbed or Pivoting user interfaces; for Dialogs; and for Split displays. We'll introduce some of these briefly at the end of this article.
-
-## The initial navigation
```
martijn00: I would still leave something about the initial navigation in the doc.

MarcBruins: I was thinking about this as wel. However i can imagine if you search for navigation you want to know how to navigate from viewmodel x to viewmodel y. The Initial navigation is more of a project startup setup which is already handled in the MvvmCross overview(https://www.mvvmcross.com/documentation/fundamentals/feature-overview).

martijn00: Right, but the case of using a custom `AppStart` is not there if i look correctly.

MarcBruins: It would be nice to have some docs apart from the overview to tell about setting up your initial setup/navigation. But i don't think it belongs in the "navigation" section.

PR #1733, opened 2017-04-26
Fix null check in Adapter setter

kjeremy:
We need to guard against existingHeaderFooterAdapter being null


kjeremy:
We will need this in 5.0

PR #1736, opened 2017-04-27
Move bind extension methods namespace under MvvmCross.Binding.BindingContext for ease of use

Plac3hold3r:
This PR is related to #1628 and #1609 that added extension methods for Mvx defined target bindings.

I have changed the namespace for all extension method classes to be under the shared `MvvmCross.Binding.BindingContext` namespace. This namespace is the one that contains the `CreateBindingContext` and `CreateBindingSet` extension methods. Therefore once a developer has include the `MvvmCross.Binding.BindingContext` namespace they will automatically get access to all scoped binding extension method, and no long need to figure out any additional namespaces required from platform/android support or plugins.

Unfortunately, there is a side effect that affected Assembly namespaces are no longer nicely conform to there names. For example the Assembly `MvvmCross.Binding.iOS` would now include `MvvmCross.Binding.BindingContext` thereby breaking the naming conversion pattern.

image

I do feel however, that the benefit to the developer is greater.

I'm not sure if there is a more elegant way to achieve this?

### DIFF
--- a/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.AppCompat/BindingContext/MvxAppCompatPropertyBindingExtensions.cs
+++ b/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.AppCompat/BindingContext/MvxAppCompatPropertyBindingExtensions.cs
@@ -6,9 +6,10 @@
 // Project Lead - Stuart Lodge, @slodge, me@slodge.com
 
 using Android.Support.V7.Widget;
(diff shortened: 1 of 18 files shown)


martijn00:
I'm not sure if this is the right thing to do. Can we think about other solutions?

Plac3hold3r:
@martijn00 agreed, it did feel it was bit hacky to mess with the assembly namespaces. I was not sure how else increase the discoverability of all the target bindings that are sitting in various other assemblies. I'll close the PR for now, and see if there is not another alternative.

PR #1739, opened 2017-04-27
Add editorconfig file to help with with formatting and conventions

Plac3hold3r:
Visual Studio 2017 now brings [support](https://www.visualstudio.com/en-us/news/releasenotes/vs2017-relnotes#coding-convention-configuration-and-enforcement). 
Additional current [limitation](https://www.visualstudio.com/en-us/news/releasenotes/vs2017-relnotes#editorconfig-is-not-supported-in-xml-files) (Not working in xml and no new line feature).

Coding style options can be changed. [Microsoft documentation](https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference) on meaning of options.
Currently, all options are set as suggestions only. 
Additionally, I have set the convention to spaces over tabs

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,40 @@
+﻿root=true
+
+[*]
+end_of_line = crlf
+charset = utf-8
+indent_style = space
+trim_trailing_whitespace = true
+
+[*.xml]
+indent_style = space
+
+[*.cs]
+indent_size = 4
+
+dotnet_style_qualification_for_field = false:suggestion
+dotnet_style_qualification_for_property = false:suggestion
+dotnet_style_qualification_for_method = false:suggestion
+dotnet_style_qualification_for_event = false:suggestion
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,40 @@
+﻿root=true
+
+[*]
+end_of_line = crlf
```
kjeremy: If I remember correctly I think we default to line feed. @Cheesebaron?

Cheesebaron: on Windows crlf is fine, it is what is default in most editors. But it is recommended to have crlf = auto in the git config.

Cheesebaron:
As for more options for editor config take a look at coreclr's rules: https://github.com/dotnet/coreclr/blob/master/.editorconfig

A lot of them are still not used in VS afaik.

Also if you need language support for editor config, there is https://marketplace.visualstudio.com/items?itemName=MadsKristensen.EditorConfig

PR #1741, opened 2017-04-28
Added MvxBindings to Forms integration

LeRondPoint:
Similar to the Windows implementations, we use the BindableProperties to hook up the Mvx binding engine in Forms


Cheesebaron:
Nice. Great first time contribution 🙌

LeRondPoint:
Thanks, will look into the La.ng and then I think it's possible to add a markup extension to make thin more inline with the regular bindings, allowing for something like
<,.. Text="{MvxBinding blahblah}" />

martijn00:
Nice work! I'm thinking we should rework the structure of the Forms projects a bit like the other MvvmCross projects are. We have 1 PCL project containing all shared stuff and then the other platform specific projects added to that. Currently we have only projects which contain the presenter, but in the new structure we can add the presenter stuff in a folder, and the bindings in another. This will then also allow to develop other Forms features into the project.

LeRondPoint:
Yes Martin I agree, event the namespaces aren't standardized in the Forms project.  In fact, shouldn't Forms be handled just like a different platform but in it's own Nuget?

martijn00:
Yeah i think we should do that. Let me review this PR, and merge it asap. After that i'll fix up the projects and nuget.

MvvmCross/Droid/Droid/Platform/MvxAndroidSetup.cs
```diff
@@ -158,7 +159,7 @@ protected virtual void RegisterBindingBuilderCallbacks()
             Mvx.CallbackWhenRegistered<IMvxNamespaceListViewTypeResolver>(FillNamespaceListViewTypeResolver);
         }
 
-        protected virtual MvxAndroidBindingBuilder CreateBindingBuilder()
+        protected virtual MvxBindingBuilder CreateBindingBuilder()
```
martijn00: Why does this need to be changed?

LeRondPoint: So we can use the overloaded class without reimplementing the InitializeBindingBuilder.  Look at the Android Setup.cs

martijn00: Aah right, and it should still be compatible.

LeRondPoint: Absolutely

martijn00:
I'll rework the structure a bit now. Are you on Slack by any chance?

MvvmCross/Droid/Droid/Platform/MvxAndroidSetup.cs
LeRondPoint: This leads me to a question...  Should we have a MvxFormsSetup base class?  Not sure if it's possible without reqriting too much code...

LeRondPoint:
Only on my Toptal account  sylvain@toptal.com

MvvmCross/Droid/Droid/Platform/MvxAndroidSetup.cs
martijn00: I was already thinking that we should have that, and possible even more base classes. The problem for me is that i don't have too many real world Forms apps where i could test it. We would really appreciate any contributions on this.

martijn00:
Can you join the MvvmCross channel on https://xamarinchat.herokuapp.com/

PR #1748, opened 2017-04-29
Format all code with resharper

martijn00:


martijn00:
I'll work on something less aggressive.

PR #1753, opened 2017-04-30
More clean up of Warnings

Cheesebaron:
Another batch of fixing warnings


martijn00:
The changes for context have been done in https://github.com/MvvmCross/MvvmCross/pull/1509. We might need to add this to more places.

PR #1754, opened 2017-05-01
MvxForms lang binding

LeRondPoint:
Added La.ng binding for Forms
Added a lightweight Forms solution to make development easier


TestProjects/Forms/Example002CSharp/Example/Example.csproj
```diff
@@ -8,10 +8,11 @@
     <OutputType>Library</OutputType>
     <RootNamespace>Example</RootNamespace>
     <AssemblyName>Example</AssemblyName>
-    <TargetFrameworkProfile>Profile259</TargetFrameworkProfile>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <TargetFrameworkProfile>Profile24</TargetFrameworkProfile>
```
martijn00: I don't think it is good to change this. Normally we use Profile7 with .NET 4.5. This would be fixed when you merge upstream, where this sample is removed.

TestProjects/Forms/MvxBindings/MvxBindingsExample.UWP/project.json
```diff
@@ -1,6 +1,7 @@
 ﻿{
   "dependencies": {
     "Microsoft.NETCore.UniversalWindowsPlatform": "5.2.2",
+    "Newtonsoft.Json": "10.0.2",
```
martijn00: Please use the same version of Json.Net in all projects.

Issue #1743, opened 2017-04-28
 MvxHorizontalGridView is broken


kjeremy:
IMvxRecyclerView.ItemTemplateId doesn't do anything anymore.

Cheesebaron:
@kjeremy is it still broken or was it fixed in #1758?

kjeremy:
Fixed.

Issue #1642, opened 2017-03-07
Grouped Items Observable Child Collection support


thefex:
Currently grouping does not support Observable Collections for Group Items.
Implementing support for Add/Remove/Reset events of Group Items collection...

behzad62:
I needed a grouping support and it seems that Grouping has been delayed till version 5.1. So, I came up with my own solution which I would like to share here: 
In WPF, collection controls create a wrapper around the source collection called 'CollectionView'. Therefore there is a tight coupling between the collection UIControl and the CollectionView. I also think that for grouping/sorting/filtering a good approach is to wrap the source in a middleware collection but at the same time I wanted to use the existing MvxRecyclerView. Hence, inspired by and based on the System.Windows.Data.CollectionView, I have created a CollectionView that makes grouping/sorting/filtering possible with the current MvxRecyclerView implementation: [CollectionView](https://github.com/behzad62/CollectionViews.NETStandard).
Usage is simple; Just wrap the source collection and add one or more group and sort descriptions or a filter to the CollectionView. Then we need a template selector which maps our source items to a item layout and other items to a group-header layout. 
Please note that this library is not heavily tested yet and is just a proof of concept at the moment. Also, footer and header are intrinsic parts of the View and must be implemented separately in the MvxRecyclerView itself.

thefex:
Why not check... https://github.com/thefex/Xamarin.Bindings.AdvancedRecyclerView I explained in past that is the only reasonable way to implement that
MvvmCross binding available there.

PR #1758, opened 2017-05-01
Revert MvxRecyclerView Header/Footer and Grouping Features

### LINKED ISSUE #1730: Custom MvxRecyclerViewAdapters don't work anymore
If you create a dummy `MvxRecyclerAdapter` it's `ItemSource` property gets set to an instance of
`MvvmCross.Droid.Support.V7.RecyclerView.ItemSources.MvxRecyclerViewItemsSourceBridgeHeaderFooterDecorator`

and not something reasonable like a list of stuff.

kjeremy:
This reverts the RecyclerView header/footer grouping support added for 5.0.

There are some issues with breaking existing custom adapters so we are reverting these features for the 5.0 release.

Fixes #1730, #1743
Sort of obsoletes #1642


Cheesebaron:
Looks good to me.

Will give this a spin soon.

Next time, as discussed briefly on Slack, we shouldn't merge features like these if they haven't actually been tested and demonstrated properly. Sure the code looked good and promising, that is however not enough to ensure that we don't make a big regression. Since the author (@thefex) hasn't responded in #1730 and we are soon to release something stable, I agree with @kjeremy that this shouldn't make it into 5.0, not as-is.

MarcBruins:
Lets also not merge new features without having any docs about how they work. It should be really easy for devs to add documentation about there changes with the new website.

kjeremy:
Can we get this in before the next nuget beta?

martijn00:
Is it an idea to pull over the new features to a new adapter instead of completely removing them?

kjeremy:
Not right now.  I think this will need to be redone.

PR #1763, opened 2017-05-02
Direct Forms bindings with MvxBind and MvxLang

LRP-sgravel:
MvxBind & MvxLang markup extensions

** Note : Requires Xamarin Forms 2.3.5 so the Bindings sample project was updated


martijn00:
This looks great, but we should hold off merging until Forms 2.3.5 hits stable. Otherwise we can't reference the dependency on Nuget.

LeRondPoint:
Agreed

martijn00:
We need to update to Forms 2.4 after this.

PR #1764, opened 2017-05-03
change parameter type to TInit

Hobbes1987:
Parameter used to be a string, but this was incorrect. Should be the
generic type TInit.


martijn00:
Thanks!

PR #1765, opened 2017-05-03
Add documentation about the new navigation and fix copy pasta error

martijn00:


docs/_documentation/fundamentals/navigation.md
```diff
@@ -6,6 +6,126 @@ order: 3
 ---
 MvvmCross uses `ViewModel first navigation`. Meaning the we navigate from ViewModel to ViewModel and not from View to View. In MvvmCross the ViewModel will lookup its corresponding View. By doing so we don't have to write platform specific navigation and we can manage everything from within our core.
 
+# MvvmCross 5.x and higher navigation
+
+MvvmCross 5 introduces a new NavigationService! The new navigation enables you to inject it into your ViewModels, which makes it more testable, and give you the ability to implement your own version of the Navigation! Other main features are that it is fully async and type save.
```
kjeremy: Change wording to "your own navigation" and "and type safe".

MarcBruins:
Looks good!

lucian-v:
I'm not sure this is the best place to ask this, but I'm having trouble with deep linking and I couldn't find answers in the docs.

In here it's stated that `The Uri navigation will build the navigation stack if required. This will also enable deeplinking and building up the navigationstack for it.`, but I can't see any examples of how this can be achieved.
I might have misunderstood what this does, what I need is that when a user is "redirected" from a screen (`ViewModelA`) to a screen that is normally part of a hierarchy (`ChildViewModel`), then when she presses the back button is taken up in the hierarchy (`ParentViewModel`), not back where she came from (`ViewModelA`).

What I have done trying to achieve this is:
```
[assembly: MvxNavigation(typeof(SettingsViewModel), @"mvx://settings$")]
namespace ...
public class SettingsViewModel : BaseViewModel
```
```
[assembly: MvxNavigation(typeof(SecuritySettingsViewModel), @"mvx://settings/security")]
namespace ...
public class SecuritySettingsViewModel : BaseViewModel
```
and then in `ViewModelA`:
```
_navigationService.Navigate("mvx://settings/security");
```

I did this because I assumed that sharing a part of the Uri was how the stack was built, but it seems I was wrong because when hitting back on `SecuritySettingsViewModel` I still arrive at `ViewModelA` and not `SettingsViewModel`.

Did I misunderstand the usage?
Can this be achieved?
Should I ask this somewhere else?

Issue #393, opened 2013-08-13
Add abstraction for path in File plug-in


Cheesebaron:
I am thinking that it could be nice with some kind of abstraction, to allow the developer to choose where to store files with the File plug-in.

Right now there is the `FullPath` method which on
- iOS defaults to `﻿Environment.SpecialFolder.MyDocuments`
- Android defaults to `Context.FilesDir.Path`
- WP and WinRT uses IsolatedStorage

Some of these have options on where to store files, for instance Android devices can have SDCards. iOS also has private storage for files only readable by the app etc.

So my thoughts are that it would be nice to be able to pick the actual path where to save the files, as the default location is not always desirable. However, to ease the usage, something like an `enum` with some common paths could be made. On platforms which do not support SDCards etc. they will simply default to something sane.

What are your thoughts about that?

Cheesebaron:
It is possible to add your own BasePath since MvvmCross 5.0.0. This solves the issue.

PR #1778, opened 2017-05-07
Allow user to provide full path in the file plugin

### LINKED ISSUE #1673: MvvmCross.Plugins.File MvxFileStore read using full path
### Steps to reproduce

1. Create `OpenReadFullPath(string fullPath)` method

### Expected behavior
I need to the option to use full path in `MvxFileStore`  instead of always using the default. Now, I only need `OpenRead` to read full path.

### Actual behavior
Now, the plugin always add its default path (base path) to the beginning of supplied path.

### Configuration

**Version:** 4.4

**Platform:** Android

willsb:
Closes #1673

I am using a configuration object instead of adding new methods because we’d basically have to double all the existing methods to provide support for all operations.

I am not running on Windows, so I did not implement UWP (and WPF could use a thorough review, just in case).


willsb:
Now that I read the proposal on #393, maybe it makes sense to add a `DefaultPath` property on the Configuration object as well, allowing consumers to choose whether or not the default path should be used and also select what the default path is. What do you think?

Cheesebaron:
Would be nice to have it configurable.

Another idea is to maybe call it BasePath instead? Makes more sense as you are appending to it.

willsb:
This is ready for a reviewing and merging, @martijn00 :)

willsb:
@Cheesebaron Can you please check this again now? :)

PR #1780, opened 2017-05-08
Changed modifiers so can be used in overrides

eL-Prova:
Changed the private and internal to protected and protected internal so
it can be used by custom view presenters which overrides this class

### DIFF
--- a/MvvmCross/Windows/Uwp/Views/MvxWindowsMultiRegionViewPresenter.cs
+++ b/MvvmCross/Windows/Uwp/Views/MvxWindowsMultiRegionViewPresenter.cs
@@ -21,7 +21,7 @@ namespace MvvmCross.Uwp.Views
     public class MvxWindowsMultiRegionViewPresenter
         : MvxWindowsViewPresenter
     {
-        private readonly IMvxWindowsFrame _rootFrame;
+        protected readonly IMvxWindowsFrame _rootFrame;
 
         public MvxWindowsMultiRegionViewPresenter(IMvxWindowsFrame rootFrame)
             : base(rootFrame)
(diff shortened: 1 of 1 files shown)


MvvmCross/Windows/Uwp/Views/MvxWindowsMultiRegionViewPresenter.cs
```diff
@@ -70,14 +70,14 @@ public override void Close(IMvxViewModel viewModel)
...
-        private static Type GetViewType(MvxViewModelRequest request)
+        protected static Type GetViewType(MvxViewModelRequest request)
         {
             var viewFinder = Mvx.Resolve<IMvxViewsContainer>();
             return viewFinder.GetViewType(request.ViewModelType);
         }
 
         // Implementation from: http://stackoverflow.com/a/1759923/80186
-        internal static T FindChild<T>(DependencyObject reference, string childName) where T : DependencyObject
+        protected internal static T FindChild<T>(DependencyObject reference, string childName) where T : DependencyObject
```
martijn00: Should this still be internal?

eL-Prova: @martijn00 I have not enough knowlege where this method is used MvvmCross near this class. So to prevent breaking applications I added protected and MvvmCross should work without any problems

Cheesebaron: This won't break things and is a pretty frequent thing to implement in XAML based apps. Whether or not it should be visible or not it not a huge deal. Will of course clash with implementations with same signature.

eL-Prova: I dont know if we understand each other or we have the same vision but translated in a different way. Internal will normally be visible within his namespace. By changing it to protected it will be only visible to itself and subclasses. So to prevent a crash for classes in the same namespace which (can be) used this method, we support the old behaviour and subclasses (in other namespaces). Is this what you also ment to @Cheesebaron?

Cheesebaron: Yes, that is what I meant. If you subclass and implement a method with the same signature, it will clash. Which probably will not be a huge deal.

PR #1568, opened 2017-01-20
Feature/android support/recycler view header footer grouping support

thefex:
Implements issue #1565 - please check issue for more info about PR.

PR #1786, opened 2017-05-10
Added overridable methods to MvxAdapter for index modification.

fela98:
Added these protected virtual methods to give more customisability. There are no behavioural changes. The additional customisability allows you to create custom adapters that add items in between or at the top or bottom of the list. Here an example of an adapter that adds a header by overriding these virtual methods.

``` C#
public class HeaderAdapter : MvxRecyclerAdapter
{
    public HeaderAdapter(IMvxAndroidBindingContext bindingContext) : base(bindingContext)
    {
    }

    public override int ItemCount => (ItemsSource as IList).Count + 1;

    protected override int GetViewPosition(int itemsSourcePosition)
    {
        return itemsSourcePosition + 1;
    }

    protected override int GetItemsSourcePosition(int viewPosition)
    {
        return base.GetItemsSourcePosition(viewPosition) - 1;
    }

    public override object GetItem(int viewPosition)
    {
        if (viewPosition == 0)
            return null;

        return base.GetItem(viewPosition);
    }

    public override int GetItemViewType(int position)
    {
        if (position == 0)
            return 6;

        return base.GetItemViewType(position);
    }

    public override void OnBindViewHolder(Android.Support.V7.Widget.RecyclerView.ViewHolder holder, int position)
    {
        if (position == 0)
        {
            ((IMvxRecyclerViewHolder)holder).DataContext = BindingContext.DataContext;
        }
        else
            base.OnBindViewHolder(holder, position);
    }
}
```

And then you could use the integer returned from GetItemViewType inside a template selector and return the corresponding layout.

This should also make it easier to reimplement header and footer support inside MvvmCross properly with custom adapter support.

### DIFF
--- a/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.RecyclerView/IMvxRecyclerAdapter.cs
+++ b/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.RecyclerView/IMvxRecyclerAdapter.cs
@@ -21,7 +21,7 @@ namespace MvvmCross.Droid.Support.V7.RecyclerView
         ICommand ItemClick { get; set; }
         ICommand ItemLongClick { get; set; }
 
-        object GetItem(int position);
+        object GetItem(int viewPosition);
 
         int ItemTemplateId { get; set; }
     }
--- a/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.RecyclerView/MvxRecyclerAdapter.cs
+++ b/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.RecyclerView/MvxRecyclerAdapter.cs
@@ -167,9 +167,32 @@ namespace MvvmCross.Droid.Support.V7.RecyclerView
 
         public override int ItemCount => _itemsSource.Count();
 
-        public virtual object GetItem(int position)
+        public virtual object GetItem(int viewPosition)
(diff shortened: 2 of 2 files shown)


martijn00:
Did you see the work in https://github.com/MvvmCross/MvvmCross/pull/1568?

martijn00:
Let merge this until we get a contribution to add the headers/footers and grouping back in a stable shape.

Cheesebaron:
Thanks @fela98 looks good!

fela98:
Yes @kjeremy, you are probably right, there is probably a better term for this. I used the term ViewPosition because it indicates the index that is closer to the view in the scope of this component. The idea is to be able to create a mapping between the view that is displayed and the data from ItemsSource, so you then are able to add in arbitrary items. If you have a better name for this, feel free to create a new pull request and update the naming.

kjeremy:
Yep I think that it makes sense everywhere but in that method. Do you
actually need to modify NotifyDatasetChanged for this PR to work.  Also can
you add an example of your custom adapter to the examples project so we can
see it in use?

fela98:
Yes, that's exactly the thing with this pull request, that those indexes also have to be updated. Otherwise one could do this without my modifications to the MvxAdapter I tried first to do this without changing NotifyDatasetChanged events, but weird things would happen where the list would be populated backwards and so on.

PR #1789, opened 2017-05-10
- Fix Android crash when app restart

thongdoan:
- When app restart, IMvxChildViewModelCache will be re-create, so there are no cached viewModels


thongdoan:
Yes, they maybe need add some events into IMvxEventSourceActivity

PR #1792, opened 2017-05-11
Install System.Windows.Interactivity from nuget instead of requiring …

kjeremy:
…the Expression Blend SDK


Cheesebaron:
Hehe, I think I reverted this in a previous commit from exactly this.

I think there were some build errors, where it couldn't find System.Windows.Interactivity. Can't remember the exact reason. If this works, fine with me.

kjeremy:
Yep that's exactly the error I was hitting.

PR #1794, opened 2017-05-11
Fix for MvxRecyclerAdapter indexes

fela98:
Small fix for mistake made in #1786


fela98:
**Yes I am 100% certain that the starting location needs to go through `GetViewPosition`.**

I do not fully understand what you mean by adapter positions and visual positions when you write.
> It looks like incorrect behavior if the starting value can be modified but not the count, especially because we are talking about changes in the adapter positions and not visual positions.

What i know, is that even though it might not be the "visual position" in the adapter, that is what it results in. The views get affected by it, and it gets displayed to the user.

The reason the starting value has to be modified is to be able to offset the whole `ItemsSource` by a number. So even though there are only 3 items in the ItemsSource a header can been added resulting in an ItemCount of 4. Additionally now the item at index 0 is now at index 1, and when the item in the collection at index 0 is removed, we have to remove the item in the adapter that is at index 1. The count of the changed items however doesn't need to be modified because it's not a change in how many items changed in the ItemsSource but only their position (they're offset, moved one or more step down).

PR #1796, opened 2017-05-12
Update release blog

martijn00:


MarcBruins:
For the pieces of code we should just refer to the docs i think. Add a few lines about the new feature  and add a link to the real documentation.

MarcBruins:
We removed several plugins including sqlite we should tell people how to get the same functionality without using the plugin.

docs/_documentation/fundamentals/data-binding.md
```diff
@@ -439,7 +439,83 @@ where `$ViewObject$` is the view target for binding.
 
 where `$ViewProperty$` is the property on the view for binding.
 
-If `For` is not provided, then the default view property is used - e.g. for a `UILabel` the default is `Text`
+If `For` is not provided, then the default view property is used.
+
+**Android**
```
Cheesebaron: These lists are placed without much context in this document, please back it up with some description.

Issue #1776, opened 2017-05-07
the mvvmcross.com's Doc isn't too good show in mobile phone


HeMinzhang:
To help us fix your issue, please provide the information in the below template. 

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.
>
> Please only file bugs verified on latest version of MvvmCross and _stable_ Xamarin tools

### Steps to reproduce

![qq 20170507152230](https://cloud.githubusercontent.com/assets/2683269/25778712/1e2f487e-3339-11e7-897a-6e6b72f59608.png)

thanks a million.

### Expected behavior
Tell us what should happen

### Actual behavior
Tell us what happens instead

### Configuration

**Version:** 4.x

**Platform:** iOS/Android/Windows Phone

MarcBruins:
Hello HeMinzHang! Those issues are know: #1727 thanks for reporting however, if you would like to help out resolving these issues that would be great!

PR #1802, opened 2017-05-16
Lost dependency gem 'jekyll-avatar'

Garfield550:


Garfield550:
Fix search positioning is incorrect and it doesn't display the icon #1776
Fix hamburger icon flies to the left side once clicked #1776

mvvmcross - fix

PR #1808, opened 2017-05-16
Fix Android Fragment caching for new navigation

martijn00:
This fixes fragments not using the already created viewmodel.

The problem was that when showing a ViewModel which is in a Fragment the it serializes the MvxViewModelInstanceRequest, in there the instance is lost, so when the Activity is shown the Fragment creates a new type based on the ViewModelType, but there is already an existing instance of it available. This makes sure that the correct instance is restored.

### DIFF
--- a/MvvmCross/Droid/Shared/Presenter/MvxFragmentsPresenter.cs
+++ b/MvvmCross/Droid/Shared/Presenter/MvxFragmentsPresenter.cs
@@ -67,6 +67,11 @@ namespace MvvmCross.Droid.Shared.Presenter
             var serializedRequest = Serializer.Serializer.SerializeObject(request);
             bundle.PutString(ViewModelRequestBundleKey, serializedRequest);
 
+            if (request is MvxViewModelInstanceRequest)
+            {
+                Mvx.Resolve<IMvxChildViewModelCache>().Cache(((MvxViewModelInstanceRequest)request).ViewModelInstance);
+            }
+
             if (!_fragmentHostRegistrationSettings.IsActualHostValid(request.ViewModelType))
             {
                 Type newFragmentHostViewModelType =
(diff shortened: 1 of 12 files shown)


kjeremy:
Can you explain what the problem was and what the behavior is now?

MvvmCross/Droid/Shared/Presenter/MvxFragmentsPresenter.cs
```diff
@@ -67,6 +67,11 @@ protected virtual void ShowFragment(MvxViewModelRequest request)
             var serializedRequest = Serializer.Serializer.SerializeObject(request);
             bundle.PutString(ViewModelRequestBundleKey, serializedRequest);
 
+            if (request is MvxViewModelInstanceRequest)
+            {
+                Mvx.Resolve<IMvxChildViewModelCache>().Cache(((MvxViewModelInstanceRequest)request).ViewModelInstance);
```
kjeremy: Should we cache the result of `Mvx.Resolve<IMvxChildViewModelCache>()` to avoid a lookup each time?

martijn00: The result is only cached until the fragment is shown and the result is attached as viewmodel. It doesn't make sense to cache it here because it will never reach this point again.

kjeremy: So the `IMvxChildViewModelCache` is only used once?

martijn00: In this case it is. But other then that `IMvxChildViewModelCache` is also used for Fragment lifecycle situations.

PR #1814, opened 2017-05-17
Fix MvxTabBarViewController.VisibleUIViewController

### LINKED ISSUE #1812: MvxTabBarViewController.VisibleUIViewController throws for tabs without a navigation controller
Problem is on line 115:

    var topViewController = (SelectedViewController as UINavigationController).TopViewController;

This assumes every tab is wrapped in a UINavigationController, which can be disabled using the presentation attribute:

        [MvxTabPresentation(
            TabName = "Targets",
            TabIconName = "TabBarItemTargets",
            WrapInNavigationController = false)]

If we do, the above lines throws an NRE when the selected tab is one without a navigation controller.

nmilcoff:
Fixes #1812

### DIFF
--- a/MvvmCross/iOS/iOS/Views/MvxTabBarViewController.cs
+++ b/MvvmCross/iOS/iOS/Views/MvxTabBarViewController.cs
@@ -112,9 +112,7 @@ namespace MvvmCross.iOS.Views
         {
             get
             {
-                var topViewController = SelectedViewController is UINavigationController
-                                        ? ((UINavigationController)SelectedViewController).TopViewController
-                                        : SelectedViewController;
+                var topViewController = (SelectedViewController as UINavigationController)?.TopViewController ?? SelectedViewController;
 
                 if(topViewController.PresentedViewController != null)
                 {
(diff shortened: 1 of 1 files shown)


MvvmCross/iOS/iOS/Views/MvxTabBarViewController.cs
```diff
@@ -112,7 +112,10 @@ public virtual UIViewController VisibleUIViewController
         {
             get
             {
-                var topViewController = (SelectedViewController as UINavigationController).TopViewController;
+                var topViewController = SelectedViewController is UINavigationController
```
Cheesebaron: Couldn't you just have done:

```
var topViewController = (SelectedController as UINavigationController)?.TopViewController ?? SelectedViewController;
```

That seems much cleaner.

Cheesebaron: Also please rebase instead of merging develop in.

nmilcoff: Yes, you're right! Thanks

PR #1816, opened 2017-05-18
Added missing calls to view model lifecycle methods. Fixes #1813.

### LINKED ISSUE #1813: MvxTableViewController does not call MvxViewModel.Appearing() et. al
`MvxViewController` and `MvxBaseSplitViewController` both call the following methods on the view model:

- Appearing()
- Appeared()
- Disappearing()
- Disappeared()

However this is not done from `MvxTableViewController` and `MvxCollectionViewController`.

DanielStolt:


DanielStolt:
Fixes #1813.

PR #1823, opened 2017-05-19
Add sidebar menu open/close callbacks

jamespettigrew:
This PR adds `MenuWillOpen()`, `MenuDidOpen()`, `MenuWillClose()` and `MenuDidClose()` callbacks to the view controller implementing `IMvxSidebarMenu`.

As Xamarin Sidebar doesn't integrate with view controller lifecycle events, the only way to receive notifications about the menu opening/closing is through its `StateChangeHandler`, which MvvmCross currently doesn't do much with. For some reason, `StateChangeHandler` is only invoked after the menu open/close animation has completed, which I find undesirable (I'd like to update viewmodel properties prior to menu open). As such, I've included callbacks that will execute as soon as menu toggle/open/close are requested as well as callbacks that will execute upon animation completion. 

Due to the interface modifications, this is a minor breaking change.

### DIFF
--- a/MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/Extensions/ViewControllerExtensions.cs
+++ b/MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/Extensions/ViewControllerExtensions.cs
@@ -45,6 +45,14 @@ namespace MvvmCross.iOS.Support.XamarinSidebar.Extensions
                     , UIBarButtonItemStyle.Plain
                     , (sender, args) =>
                     {
+                        if (sidebarController.IsOpen)
+                        {
+                            mvxSidebarMenu.MenuWillClose();
+                        }
+                        else
+                        {
+                            mvxSidebarMenu.MenuWillOpen();
+                        }
                         sidebarController.MenuWidth = mvxSidebarMenu.MenuWidth;
                         sidebarController.ViewWillAppear(false);
                         sidebarController.ToggleMenu();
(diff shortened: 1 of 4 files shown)


mvanbeusekom:
This looks very useful thanks. I have however a question, you mention that the `StateChangeHandler` event only fires after the "open/close animation has completed". Does it still fire when you don't use animation (i.e. when you set the `AnimateMenu` property of the `IMvxSidebarMenu` to false)?

mvanbeusekom:
Another question I have, wouldn't it be better to turn the callbacks into events? This way you can handle the events in the view that is currently open. Maybe I am just making it to complex but I can imagine it would be useful to know on other places as well that the menu is about to open / close.

jamespettigrew:
> This looks very useful thanks. I have however a question, you mention that the StateChangeHandler event only fires after the "open/close animation has completed". Does it still fire when you don't use animation (i.e. when you set the AnimateMenu property of the IMvxSidebarMenu to false)?

It still fires, yes. However, from what I can tell, the `AnimateMenu` property doesn't doesn't actually do anything. It gets passed into a `ViewWillAppear` call on the sidebar controller which only serves to ensure that the sidebar view is correctly positioned--it has no effect on whether the menu open/close is actually animated or not.

> Another question I have, wouldn't it be better to turn the callbacks into events? This way you can handle the events in the view that is currently open. Maybe I am just making it to complex but I can imagine it would be useful to know on other places as well that the menu is about to open / close.

I had that thought but, if you're talking about EventHandlers, I'm not sure how to answer the question of where they'd be put. Where would they go such that they can be both subscribed to and invoked without creating lots of coupling? I imagine it would be simpler to leave it up to the user--you could publish a message in one of the callbacks using MvxMessenger.

mvanbeusekom:
That makes sense indeed. Thanks for the explanation.

Issue #1759, opened 2017-05-02
PictureChooser.iOS scales incorrectly on high DPI devices


ElteHupkes:
### Steps to reproduce
Take a picture with the picture chooser task on iOS with a pixel dimension larger than the `maxPixelDimension` (which should be passed), on a retina display that has a default scaling > 1.0. Save the picture somewhere - upload it or put it in the device photo album.

### Expected behavior
The maximum dimension of the picture should equal the passed `maxPixelDimension`

### Actual behavior
The maximum dimension equals `maxPixelDimension * s` where `s` is the device scaling factor.

### Configuration

**Version:** 4.x

**Platform:** iOS

### Cause
The code at line

https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross-Plugins/PictureChooser/MvvmCross.Plugins.PictureChooser.iOS/UIImageHelpers.cs#L58

initialises an image context with the scaling factor set to 0, which means the default device scaling factor will be used for the image. The target image size however does not take this scaling factor into account. Therefore the resulting image's actual size will depend on the scaling factor of the device.

ben-dean:
+1 on this issue.

I've noticed this myself today. I'm assuming that this is a bug as Android doesn't scale images based off the device's screen density.

My use case for the images, is uploading them to an API, so I'm getting inconstant image dimensions depending on the users device. The current code would make sense if I was just showing the image on screen and never uploading it.

Cheesebaron:
If anyone wants to fix this I have marked it up for grabs.

ben-dean:
It's a one character (3rd parameter on BeginImageContextWithOptions) change telling the image context not to automatically determine scaling.

It currently has 0 specified, which means it will look at the device and change the width and height based on the device. x2 for a retina device and x3 for a plus device. 

By specifying 1 it wont scale the width or height at all.

Cheesebaron:
But if you don't use scaling factor to multiply the width and height, wouldn't that cancel it out?

ben-dean:
The image.CurrentScale property is set to tell you if the image is designed for retina screens and the size property will always be set to the computed size rather than the actual size.

E.g.

| Actual Size | Filename | Size Property | CurrentScale Property |
| --- | --- | --- | --- |
| 100x100 | image.png | 100x100 | 1 |
| 200x200 | image@2x.png | 100x100 | 2 |
| 300x300 | image@3x.png | 100x100 | 3 |

[...]

PR #1824, opened 2017-05-19
Fix for #1759 (PictureChooser iOS incorrect scaling)

ben-dean:
The image when resized was using an image context with “use device
scale” specified. 2 on @2x devices and 3 on @3x devices.

UIGraphics.BeginImageContextWithOptions(destRect.Size, false, 0);

Instead it should have been fixed at 1.

UIGraphics.BeginImageContextWithOptions(destRect.Size, false, 1);


Cheesebaron:
Just gave this a spin.

It doesn't seem like images in photo gallery or newly taken images have a scaling factor of anything else than 1. So I believe we can remove Line 25 & 26 too.

However, this should be good to go.

PR #1825, opened 2017-05-19
Fix close result on Android and add initialise for non parameter views

martijn00:


MvvmCross/Core/Core/ViewModels/MvxViewModel.cs
```diff
@@ -102,42 +108,23 @@ public void SetClose(TaskCompletionSource<TResult> tcs)
 
         public virtual Task<bool> Close(TResult result)
         {
+            _isClosing = true;
```
DanielStolt: I would reset this to `false` in a `finally` block, because Close() might return false, or exceptions might be thrown.

PR #1826, opened 2017-05-20
MvxEditTextEditableTargetBinding

tritter:
The ‘Editing’ binding can be used to enable or disable focus/clickable
In Android EditTextViews. This enables to use the new Editable binding like I changed in the sample:
`local:MvxBind="Text Password; Editable(IsLoading == false)"`

To invert a boolean value the '!' operator is usually used, under water this is a '^' XOR operation, I saw that it already exists but is not included in the list of combiners nor called anywhere, so I added the XOr combiner and a new InverterCombiner. 

Adding the Xor combiner allows XOR binary operations like:
`local:MvxBind="Text Password; Editable(IsLoading ^ true)"`

The Invertercombiner is able to convert a boolean which enables: `local:MvxBind="Text Password; Editable(!IsLoading)"`


kjeremy:
I'm on mobile so will review later.  However why use editable over enabled?

tritter:
@kjeremy Because editable is deprecated, I also tried the inputType = none, but it interferes with password fields.  (see http://stackoverflow.com/questions/4297763/disabling-of-edittext-in-android)

tritter:
@kjeremy Yes enabled does exactly what I want let me create a new PR for the combiners

Issue #1155, opened 2015-10-15
File Plugin: GetSize(), GetListWriteTimeUtc()


tekkies:
It would be useful to have some file metadata methods in the file plugin, e.g.
- GetSize(string path)
- GetLastWriteTimeUtc(string path)

Cheesebaron:
If you have the courage to implement it, a pull request would be much appreciated.

tekkies:
Now I got my mac mini build box, I'll see if I can sort it.

promontis:
@tekkies are you using a service for the build box? Or just within your home network?

tekkies:
I have a mac mini under my desk at work.

willsb:
This was closed by #1827

PR #1827, opened 2017-05-20
Add GetSize and GetLastWriteTimeUtc #1155

jz5:
I also want these methods.


martijn00:
Doesn't this need to be implemented on the other platforms as well?

jz5:
It works on iOS/Android/WPF/UWP.
MvxIoFileStoreBase.cs is linked from iOS/Android/WPF Plugins.File project.
Only UWP required special implementation (MvxWindowsCommonFileStore.cs)

PR #1828, opened 2017-05-21
Added support for ! and ^ operators in Tibet parsing

tritter:
Added support for ! and ^ operators in Tibet parsing.

To invert a boolean value the '!' operator is usually used, under water this is a '^' XOR operation, I saw that it already exists but is not included in the list of combiners nor called anywhere, so I added the XOr combiner and a new InverterCombiner. 

Adding the Xor combiner allows XOR binary operations like:
`local:MvxBind="Text Password; Enabled(IsLoading ^ true)"`

The Invertercombiner is able to convert a boolean which enables: `local:MvxBind="Text Password; Enabled(!IsLoading)"`


MarcBruins:
Can we add some docs on this?

PR #1829, opened 2017-05-21
Add support for all UIControlEvents

tritter:
It was not possible before to use all possible control events on controls. This limitation made it impossible to use for example the TouchDown event which fires directly on a button touch, which is required in some designs. I added all control events that are possible and merged the current target binders into one MvxUIControlTouchUpInsideTargetBinding.cs


MvvmCross/Binding/iOS/Target/MvxUIControlTargetBinding.cs
```diff
@@ -0,0 +1,209 @@
...
+                if (this._canExecuteSubscription != null)
+                {
+                    this._canExecuteSubscription.Dispose();
+                    this._canExecuteSubscription = null;
+                }
+            }
+            base.Dispose(isDisposing);
+        }
+
+		private void AddHandler(UIControl control)
```
Cheesebaron: Can we use WeakEvents instead here?

TestProjects/iOS-Support/XamarinSidebar/MvvmCross.iOS.Support.Sidebar/Views/CenterPanelView.cs
```diff
@@ -42,7 +42,7 @@ public override void ViewDidLoad()
             var bindingSet = this.CreateBindingSet<CenterPanelView, CenterPanelViewModel>();
             bindingSet.Bind(label).To(vm => vm.ExampleValue);
             bindingSet.Bind(rightPanelInstructions).To(vm => vm.RightPanelInstructions);
-            bindingSet.Bind(masterButton).To(vm => vm.ShowMasterCommand);
+			bindingSet.Bind(masterButton).For("TouchDown").To(vm => vm.ShowMasterCommand);
```
Cheesebaron: Can we still have TouchUpInside as default if nothing specificed, such that you don't break all Apps.

Plac3hold3r: I believe this PR would not affect the default binding for a [UIButton](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Binding/iOS/MvxIosBindingBuilder.cs#L191). It would just default to the `TouchUpInside` version of this custom binding.

Plac3hold3r:
@tritter could you also update the [iOS extension method bindings](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Binding/iOS/MvxIosPropertyBindingExtensions.cs#L12) with these additions? As well as the [documentation](https://github.com/MvvmCross/MvvmCross/blob/develop/docs/_documentation/fundamentals/data-binding.md#mvvmcross-defined-custom-bindings)?

TestProjects/iOS-Support/XamarinSidebar/MvvmCross.iOS.Support.Sidebar/Views/CenterPanelView.cs
tritter: @Cheesebaron @Plac3hold3r is right, the default binding is/are not affected.

tritter:
@Plac3hold3r just updated the extensions and docs

MvvmCross/Binding/iOS/Target/MvxUIControlTargetBinding.cs
tritter: I just used the ones that were implemented before. I'm also not 100% sure if WeakEvents will use here as we attach to an existing event handler and we don't replace one.

PR #1831, opened 2017-05-21
Docs - Added margin to bottom of tables

Plac3hold3r:
I'm not to sure how to run and test the documentation site locally. But my hope is that by adding a bottom margin to table elements it will improve separation of content areas. From this:

image

to this:

image

### DIFF
--- a/docs/assets/main.scss
+++ b/docs/assets/main.scss
@@ -415,6 +415,7 @@ a:visited {
 table {
     padding: 0;
     border-spacing: 0px;
+    margin-bottom: 15px;
 }
 
 table tr {


MarcBruins:
To run the website locally [install Jekyll](https://jekyllrb.com/docs/installation/) and go to docs folder and run `jekyll serve`. Why did you wanna change the margin if we can just add an empty line? If the formatting is not good CSS wise i would suggest to do a margin top and bottom to keep it consistent.

MarcBruins:
Ah i see, all elements have a default 15px bottom from the minima theme except a few including the table. Then it is consistent with rest

Plac3hold3r:
Thanks @MarcBruins.

PR #1832, opened 2017-05-21
Docs - Updates for data binding

Plac3hold3r:
Add descriptions for default and custom binding as suggested by @Cheesebaron in #1796 .

### DIFF
--- a/docs/_documentation/fundamentals/data-binding.md
+++ b/docs/_documentation/fundamentals/data-binding.md
@@ -559,7 +559,7 @@ Using this syntax, an example binding set is:
 
 ### MvvmCross Defined Custom bindings
 
-In addition to the `Expression` based Fluent bindings, `String` and `Extension Method` based Fluent bindings are also available. This is particularly useful for situations where bindings are needed to View events or to binding targets which are not fully exposed as C# properties. For example, even though a `UIButton` does not have a `Title` property in C#, a 'Title' property can still be set via the use of custom bindings:
+In addition to the `Expression` based Fluent bindings, `String` and `Extension Method` based Fluent bindings are also available. `String` and `Extension Method` based bindings are particularly useful for situations where bindings are needed to View events or to binding targets which are not fully exposed as C# properties. For example, even though a `UIButton` does not have a `Title` property in C#, a `Title` property can still be set via the use of custom bindings:
 
     set.Bind(okButton)
        .For("Title")
@@ -569,7 +569,9 @@ In addition to the `Expression` based Fluent bindings, `String` and `Extension M
         .For(c => c.BindText())
         .To(vm => vm.Caption);
 
- **Note:** `Extension Method` based Fluent bindings require MvvmCross 5+.
+The following tables shows all the custom bindings offered by MvvmCross. 
+
+ **Note:** When using extension method based bindings you will have to include the relevant using namespace to access the extension methods and is only available in MvvmCross 5.
 
 **Android - `using MvvmCross.Binding.Droid`**
 
--- a/docs/_documentation/fundamentals/data-binding.md
+++ b/docs/_documentation/fundamentals/data-binding.md
@@ -1,4 +1,4 @@
----
+﻿---
 layout: documentation
 title: Data binding
 category: Fundamentals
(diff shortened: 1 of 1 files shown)


docs/_documentation/fundamentals/data-binding.md
```diff
@@ -569,9 +569,11 @@ In addition to the `Expression` based Fluent bindings, `String` and `Extension M
         .For(c => c.BindText())
         .To(vm => vm.Caption);
 
- **Note:** `Extension Method` based Fluent bindings require MvvmCross 5+.
+The following tables shows all the custom bindings offered by MvvmCross. 
+
+ **Note:** When using extension method based bindings you will have to include the relevant using namespace to access the extension methods and is only available in MvvmCross 5.
```
MarcBruins: This sentence seems incorrect "and is".

docs/_documentation/fundamentals/data-binding.md
```diff
@@ -439,7 +439,7 @@ where `$ViewObject$` is the view target for binding.
 
 where `$ViewProperty$` is the property on the view for binding.
 
-If `For` is not provided, then the default view property is used.
+If `For` is not provided, then the default view property is used. The following tables shows all the default bindings used against the platforms controls.
```
MarcBruins: Shoudn't it be "used for" instead of "used against"?

Plac3hold3r: Sure, I'll make the change

docs/_documentation/fundamentals/data-binding.md
Plac3hold3r: I'll revise it

Plac3hold3r:
It should read better now :)

docs/_documentation/fundamentals/data-binding.md
```diff
@@ -439,7 +439,7 @@ where `$ViewObject$` is the view target for binding.
 
 where `$ViewProperty$` is the property on the view for binding.
 
-If `For` is not provided, then the default view property is used.
+If `For` is not provided, then the default view property is used. The following tables shows all the default bindings used for each platforms control type.
```
Cheesebaron: _show_ not shows. Equivalent to say "they show" contra "they shows".

docs/_documentation/fundamentals/data-binding.md
```diff
@@ -559,7 +559,7 @@ Using this syntax, an example binding set is:
 
 ### MvvmCross Defined Custom bindings
 
-In addition to the `Expression` based Fluent bindings, `String` and `Extension Method` based Fluent bindings are also available. This is particularly useful for situations where bindings are needed to View events or to binding targets which are not fully exposed as C# properties. For example, even though a `UIButton` does not have a `Title` property in C#, a 'Title' property can still be set via the use of custom bindings:
+In addition to the `Expression` based Fluent bindings, `String` and `Extension Method` based Fluent bindings are also available. `String` and `Extension Method` based bindings are particularly useful for situations where bindings are needed to View events or to binding targets which are not fully exposed as C# properties. For example, even though a `UIButton` does not have a `Title` property in C#, a `Title` property can still be set via the use of custom bindings:
```
Cheesebaron: Why repeat it when implied by the line just before? Could just be replaced by "they".

Cheesebaron:
These tables you added. They are still very strangely placed inside of the doc. I suggest we move them into a separate document instead.

docs/_documentation/fundamentals/data-binding.md
Plac3hold3r: I get what you are saying I was just trying to be specific so that no body could read it as including expressions. But I can replace it with "They are".

Plac3hold3r:
@Cheesebaron sure. What would be the best structure. To move the tablet to some reference section or to rather split up binding documentation in to more sections?

Cheesebaron:
I think we should have a section in the docs dedicated to bindings. This way we could dive deep in to the inner workings. These tables would be a default binding targets reference.

Other topics would be how to make own target binding definitions. How we do bindings in MvvmCross through reflection etc.

PR #1839, opened 2017-05-22
Added Visibility.Hidden

tritter:
To allow more flexibility in hiding views, many platforms also support another type of Hidden. The default implementation of Windows as well(https://msdn.microsoft.com/en-us/library/system.windows.visibility(v=vs.110).aspx), but it was never implemented in MVVMCross (https://github.com/MvvmCross/MvvmCross/blob/8a824c797747f74716fc64c2fd0e8765c29b16ab/MvvmCross/Platform/Platform/UI/MvxVisibility.cs).

I added the missing Hidden attribute and tidied up the converters and implemented the Android variant. Which allows now to just make an item 'Invisible' instead of 'gone' which also increases performance on Android. The overload happens in the converter and needs to be set to true to set visibility to Visibility.Hidden instead of Visibility.Collapse.

`Visibility Visibility(IsLoading, true)`


martijn00:
Can you use spaces instead of tabs?

tritter:
@martijn00 done

kjeremy:
Can you squash and force push? We don't want the tab to space changes in
the history.

PR #1840, opened 2017-05-22
Fix a couple of navigation issue's and add support for viewmodel instances

martijn00:
This fixes navigation by url not being triggered in the right order. Also you can now pass in instances of a viewmodel to navigate to. This can be useful for devs that what to do stuff on the viewmodel before navigating.


kjeremy:
Can you lay out what the issues are that you have fixed?  instances of what?

martijn00:
This fixes navigation by url not being triggered in the right order. Also you can now pass in instances of a viewmodel to navigate to. This can be useful for devs that what to do stuff on the viewmodel before navigating.

kjeremy:
Great!  Can you put that in the commit message so it's in the history?

Issue #1618, opened 2017-02-22
MvxUIDatePickerDateTargetBinding should not set default date to DateTime.Now if MaximumDate < Now


softlion:
https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Binding/iOS/Target/MvxUIDatePickerDateTargetBinding.cs

Line 31:

      if (value == null)
               value = DateTime.UtcNow;

Instead should check if UIDatePicker.MaximumDate is set and < at now

softlion:
Something like this.
I don't like static either, but there is no way to specify a default date for a specific instance of a target binding in the actual framework. Something like WithConversion([some class instance]) would be a good idea.

        public static DateTime DefaultDate { get; set; } = DateTime.Now;

        protected override object MakeSafeValue(object value)
        {
            var date = (DateTime)(value ?? DefaultDate);

            if (View.MaximumDate!=null && date > (DateTime) View.MaximumDate)
                date = (DateTime) View.MaximumDate;
            else if(View.MinimumDate!=null && date < (DateTime)View.MinimumDate)
                date = (DateTime)View.MinimumDate;

            return (NSDate)date;
        }

martijn00:
Can you make a PR?

PR #1843, opened 2017-05-22
Makes sure binding respects the configured max date

mvanbeusekom:
This PR fixes issue #1618. Credits go to @softlion since I simply took the code he already mentioned in the issue (however never created a PR).


mvanbeusekom:
Yes I noticed, working on it (push was a mistake)

mvanbeusekom:
Fixed, apparently pasting code with tabs in Visual Studio for Mac will preserve the tabs even when you select "Convert Tabs to Spaces" in the preferences.

Issue #1847, opened 2017-05-22
PhoneCall Plugin targeting Android Marshmallow


bazookajoey:
Attempting to utilize the MvxPhoneCallTask from the MvvmCross.Plugins.PhoneCall.Droid project, I found a missing API level check that causes a System.Reflection.TargetInvocationException. The Android level that raises this exception is Marshmallow (API 23). There is an SDK check on Lollipop to use the correct method to format a phone number, but not to get the default Locale. The Locale method attempting to be utilized does not exist until Nougat (API 24).

The SDK check should look like the following:
`if (Buidl.VERSION.SdkInt >= BuildVersionCodes.N)
	phoneNumber = PhoneNumberUtils.FormatNumber(number, Locale.GetDefault(Locale.Category.Format).Country);
if (Build.VERSION.SdkInt >= BuildVersionCodes.Lollipop)
	phoneNumber = PhoneNumberUtils.FormatNumber(number, Locale.Default.Country);
else
	phoneNumber = PhoneNumberUtils.FormatNumber(number);`

I found this bug utilizing this library through the NuGet Gallery on version 5.0.0-beta11.

kjeremy:
Please submit a PR.

bazookajoey:
New to using Github for more than browsing a repo... I'll see if I can figure that out quickly

PR #1850, opened 2017-05-23
Update API check for MvxPhoneCallTask.cs

bazookajoey:
#1847 API level check that prevents a System.Reflection.TargetInvocationException on Marshmallow devices.

### DIFF
--- a/MvvmCross-Plugins/PhoneCall/MvvmCross.Plugins.PhoneCall.Droid/MvxPhoneCallTask.cs
+++ b/MvvmCross-Plugins/PhoneCall/MvvmCross.Plugins.PhoneCall.Droid/MvxPhoneCallTask.cs
@@ -22,8 +22,10 @@ namespace MvvmCross.Plugins.PhoneCall.Droid
         public void MakePhoneCall(string name, string number)
         {
             string phoneNumber;
-            if (Build.VERSION.SdkInt >= BuildVersionCodes.Lollipop)
+            if (Build.VERSION.SdkInt >= BuildVersionCodes.N)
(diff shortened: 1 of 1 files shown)


MvvmCross-Plugins/PhoneCall/MvvmCross.Plugins.PhoneCall.Droid/MvxPhoneCallTask.cs
```diff
@@ -22,8 +22,10 @@ public class MvxPhoneCallTask
         public void MakePhoneCall(string name, string number)
         {
             string phoneNumber;
-            if (Build.VERSION.SdkInt >= BuildVersionCodes.Lollipop)
+            if (Build.VERSION.SdkInt >= BuildVersionCodes.N)
                 phoneNumber = PhoneNumberUtils.FormatNumber(number, Locale.GetDefault(Locale.Category.Format).Country);
+            else if (Build.VERSION.SdkInt >= BuildVersionCodes.Lollipop)
+                phoneNumber = PhoneNumberUtils.FormatNumber(number, Locale.Default.Country);
```
Cheesebaron: Both the N and the Lollipop version call the same method signature here, what are you protecting here?

bazookajoey: Android Marshmallow devices throws a System.Reflection.TargetInvocationException because the Locale.GetDefault(Locale.Category.Format) method was added at API Level 24 (Nougat). Pre-N devices needs to call Locale.GetDefault().

PR #1853, opened 2017-05-24
Fix not being able to use NavigationService from the appstart

martijn00:
Also fixes:
- ctor doesn't need IMvxViewDispatcher anymore
- Make navigation methods virtual
- Make IMvxViewDispatcher accessible

### DIFF
--- a/MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
+++ b/MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
@@ -10,15 +10,15 @@ using MvvmCross.Core.Platform;
 using MvvmCross.Core.ViewModels;
 using MvvmCross.Core.Views;
 using MvvmCross.Platform;
+using MvvmCross.Platform.Core;
 using MvvmCross.Platform.Exceptions;
 using MvvmCross.Platform.Platform;
 
 namespace MvvmCross.Core.Navigation
 {
-    public class MvxNavigationService : IMvxNavigationService
+    public class MvxNavigationService : MvxNotifyPropertyChanged, IMvxNavigationService
     {
-        //TODO: Should we get the dispatcher via MvxMainThreadDispatcher.Instance; ?
-        private readonly IMvxViewDispatcher _viewDispatcher;
+        protected IMvxViewDispatcher ViewDispatcher => (IMvxViewDispatcher)base.Dispatcher;
 
         private static readonly Dictionary<Regex, Type> Routes = new Dictionary<Regex, Type>();
 
@@ -42,9 +42,8 @@ namespace MvvmCross.Core.Navigation
         public event BeforeCloseEventHandler BeforeClose;
         public event AfterCloseEventHandler AfterClose;
 
-        public MvxNavigationService(IMvxViewDispatcher viewDispatcher)
+        public MvxNavigationService()
         {
-            _viewDispatcher = viewDispatcher;
         }
 
         public static void LoadRoutes(IEnumerable<Assembly> assemblies)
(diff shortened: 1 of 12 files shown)


MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
```diff
@@ -10,15 +10,15 @@
...
 using MvvmCross.Core.Views;
 using MvvmCross.Platform;
+using MvvmCross.Platform.Core;
 using MvvmCross.Platform.Exceptions;
 using MvvmCross.Platform.Platform;
 
 namespace MvvmCross.Core.Navigation
 {
-    public class MvxNavigationService : IMvxNavigationService
+    public class MvxNavigationService : MvxNotifyPropertyChanged, IMvxNavigationService
```
kjeremy: Why is this an `MvxNotifyPropertyChanged`? I don't see `SetProperty` called anywhere and we don't want to bake `INotifyPropertyChanged` into any implementation of `IMvxNavigationService`.

MvvmCross/Core/Core/ViewModels/MvxViewModelExtensions.cs
```diff
@@ -32,6 +32,10 @@ public static void CallBundleMethods(this IMvxViewModel viewModel, string method
         public static void CallBundleMethod(this IMvxViewModel viewModel, MethodInfo methodInfo, IMvxBundle bundle)
         {
             var parameters = methodInfo.GetParameters().ToArray();
+
+            if (bundle == null && parameters.Count() > 0)
```
kjeremy: This check looks odd to me.  I can't quite grok it.  Why is it needed now?  What's changed?

MvvmCross/Core/Core/ViewModels/MvxViewModelExtensions.cs
```diff
@@ -50,7 +54,7 @@ public static void CallBundleMethod(this IMvxViewModel viewModel, MethodInfo met
             }
 
             // call method using named method arguments
-            var invokeWith = bundle.CreateArgumentList(parameters, viewModel.GetType().Name)
+            var invokeWith = bundle?.CreateArgumentList(parameters, viewModel.GetType().Name)
```
kjeremy: Suddenly we get here and `bundle` can be null?  This seems like a contract violation since previously in our history `bundle` could never be null.

MvvmCross/Core/Core/ViewModels/MvxViewModelExtensions.cs
martijn00: Because we have `public async Task Init(string parameter)` in the MvxViewModel<TParameter> class. CallBundleMethod will try to call this, but the bundle is null. So when the bundle is null it should only call Init methods without parameters.

MvvmCross/Core/Core/ViewModels/MvxViewModelExtensions.cs
martijn00: This makes sure that `Init` still is called without a bundle.

MvvmCross/Core/Core/ViewModels/MvxViewModelExtensions.cs
kjeremy: I think a comment is needed right above the `if` switch... something like "Make sure we have a bundle that matches function parameters" or whatever.  It just seems to me like we're getting in this function when we shouldn't be.

MvvmCross/Core/Core/ViewModels/MvxViewModelExtensions.cs
kjeremy: Maybe a comment then about how this called `Init()`.

PR #1854, opened 2017-05-24
Add MvxStrongCommand

willsb:
This allows consumers of `MvxCommandBase` to provide their own `IMvxCommandHelper` instead of relying on the IoC. It also creates the `MvxStrongCommand`, which, unlike the default implementation of `MvxCommand`, uses a `MvxStrongCommandHelper`.

The behavior of `MvxCommand` remains the same.

### DIFF
--- a/MvvmCross/Core/Core/ViewModels/MvxCommand.cs
+++ b/MvvmCross/Core/Core/ViewModels/MvxCommand.cs
@@ -110,9 +110,13 @@ namespace MvvmCross.Core.ViewModels
         private readonly IMvxCommandHelper _commandHelper;
 
         public MvxCommandBase()
+            : this(Mvx.CanResolve<IMvxCommandHelper>() ? Mvx.Resolve<IMvxCommandHelper>() : new MvxWeakCommandHelper())
         {
-            if (!Mvx.TryResolve<IMvxCommandHelper>(out this._commandHelper))
-                this._commandHelper = new MvxWeakCommandHelper();
+        }
+
+        public MvxCommandBase(IMvxCommandHelper commandHelper)
+        {
+            _commandHelper = commandHelper;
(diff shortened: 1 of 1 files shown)


kjeremy:
Out of curiosity what's the use case for a strong command?  Are your commands getting GCed when they shouldn't?

willsb:
Precisely. More specifically, I've got a `MvxTargetBinding<UIBarButtonItem, IMvxCommand>` where I subscribe to `CanExecuteChanged` and the event is not firing unless I'm using a `MvxStrongCommandHelper ` when creating my `MvxCommand`.

kjeremy:
That sounds like a bug in our `MvxCommand` implementation.

willsb:
Not in `MvxCommand`, but in `MvxWeakCommandHelper`, though I'm not sure how much of a bug it is; I am trying to bind it with something that is not a `UIView`, so I'm not sure about its lifecycles. 

One could investigate the weak references, but I think this PR gives additional flexibility for `MvxCommand`. Currently you need to either register your own `IMvxCommandHelper ` or implement `IMvxCommand` from scratch in order to change the behavior of a command.

Bowman74:
In my humble opinion this seems like a memory leak waiting to happen. Perhaps I'm missing something?

If you are creating a strong command binding to a non-view item and know when to release it then there are other ways to keep it alive and release it when needed, making the current bindings work just fine. If you don't know when it needs to be released and are relying on the bindings to keep it alive then it seems what you are binding to would stay alive forever because you wouldn't know when to release the strong binding either which may represent a memory leak.

kjeremy:
I think this should be closed.  Having Weak and Strong commands is a big code smell and is likely to confuse people.  It looks like the problem is somewhere else.

PR #1864, opened 2017-05-26
Allow for setting of tab title and icon through interface

mvanbeusekom:
This PR will enable developers to update the tabbar name and icon by implementing an interface on the ViewController that represents the actual tab. This (in my opinion) has some benefits over the current method to programmatically determine the tabbar name and icon:

1. it allows you to access the ViewModel of the tabbar item. Meaning you can access ViewModel to determine the icon or title (makes it more cross-platform compatible);
2. implementation stays with the ViewController that implements the tab bar item (which is also decorated with the MvxTabPresentationAttribute. This eliminates the need of implementing a set of if (or a switch) statements in the `SetTitleAndTabBarItem` override in the ViewController implementing the `MvxTabBarViewController` (this is the host of the whole tab bar control).

NOTE: implementing the `IMvxTabBarItemViewController` is optional, however when implemented the values of the `TabName` and `TabIconName` take precedence over values supplied with the `MvxTabPresentationAttribute`.


nmilcoff:
Great improvement!!

PR #1873, opened 2017-05-26
Replaced `Init` and `Start` with `Initialize` method

mvanbeusekom:
Feature - update the MvvmCross Starter Pack code to make use of new MvvmCross 5 features.

I am wondering if I should also already inject the `MvxNavigationService` into the `MainViewModel`, it will not be used directly but changes are that most people will do navigation from here. And it will be a nice introduction to the new navigation service.

In the current version the generated `MainViewModel` uses the `Init` and `Start` methods to initialise the viewmodel.

Use the new `Initialize` method to allow for async support.

Nope (if released with version 5.0.1 or later). Earlier versions don't call `Initialize` from the `AppStart` method.

See breaking change.

none

- [x] All projects build

### DIFF
--- a/nuspec/CoreContent/MainViewModel.cs.pp
+++ b/nuspec/CoreContent/MainViewModel.cs.pp
@@ -1,3 +1,4 @@
+using System.Threading.Tasks;
 using MvvmCross.Core.ViewModels;
 
 namespace $rootnamespace$.ViewModels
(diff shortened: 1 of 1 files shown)


nuspec/CoreContent/MainViewModel.cs.pp
```diff
@@ -8,14 +9,11 @@
...
-        //TODO: For more information see: https://www.mvvmcross.com/documentation/fundamentals/navigation
-        public void Init()
+        public override Task Initialize()
         {
-        }
-        
-        public override void Start()
-        {
-            //TODO: Add starting logic here
+			//TODO: Add starting logic here
```
Cheesebaron: Spacing seems odd here

mvanbeusekom:
Those tabs get me everytime, apologies. They are corrected. Squashed and forced push changes.

Issue #1882, opened 2017-05-29
android.support.v7.widget.SearchView Query binding not working


StephenBeirlaen:
### Steps to reproduce

View:

    <android.support.v7.widget.SearchView
        android:id="@+id/searchView"
        android:layout_width="wrap_content"
        android:layout_height="wrap_content"
        local:MvxBind="Query SearchString" />

ViewModel:

     private string _searchString;
     public string SearchString
     {
         get { return _searchString; }
         set
         {
             _searchString = value;
             RaisePropertyChanged(() => SearchString);
         }
     }

### Expected behavior

Query should bind correctly.

### Actual behavior
Only the getter is working when the viewmodel loads. The setter doesn't.

    I/MvxBind(7360):  15,87 Failed to create target binding for binding Query for SearchString

### Configuration

**Version:** 5.0.1

**Platform:** 
- [ ] iOS
- [x] Android
- [ ] WPF
- [ ] UWP

Cheesebaron:
Query property does not have a Setter. The target binding needs SetImplementation, to invoke SetQuery().

Cheesebaron:
Just played with the Target Binding a bit. It should work From the View -> ViewModel. So if you edit the text inside of the SearchView, your ViewModel should be updated. However, as is now, if you edit the SearchString in the ViewModel, it won't be reflected in the SearchView. Will prepare a fix for that now.

Cheesebaron:
As for AppCompat SearchView, we have TargetBindings for that too. They are in MvvmCross.Droid.Support.V7.AppCompat. Make sure you replace MvxAndroidSetup with MvxAppCompatSetup instead.

StephenBeirlaen:
Confirmed, it is now working after replacing MvxAndroidSetup with MvxAppCompatSetup. Thanks a lot! :)

PR #1883, opened 2017-05-29
Add setvalueimpl for SearchView Query and change to TwoWay.

Cheesebaron:
Bug fix

ViewModel -> SearchView changes are not reflected in the SearchView

Fixes the above

No

Add EditText and SearchView to a layout, bind to same string property in ViewModel. Test that when typing in SearchView changes text in VM and EditText and vice versa.

#1882 

- [x] All projects build

### DIFF
--- a/MvvmCross/Binding/Droid/MvvmCross.Binding.Droid.csproj
+++ b/MvvmCross/Binding/Droid/MvvmCross.Binding.Droid.csproj
@@ -43,6 +43,7 @@
     <AndroidUseSharedRuntime>False</AndroidUseSharedRuntime>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Java.Interop" />
     <Reference Include="Mono.Android" />
     <Reference Include="Mono.Android.Export" />
     <Reference Include="mscorlib" />
--- a/MvvmCross/Binding/Droid/MvvmCross.Binding.Droid.csproj
+++ b/MvvmCross/Binding/Droid/MvvmCross.Binding.Droid.csproj
@@ -43,7 +43,6 @@
     <AndroidUseSharedRuntime>False</AndroidUseSharedRuntime>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Java.Interop" />
     <Reference Include="Mono.Android" />
     <Reference Include="Mono.Android.Export" />
     <Reference Include="mscorlib" />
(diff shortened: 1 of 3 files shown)


MvvmCross/Binding/Droid/MvvmCross.Binding.Droid.csproj
```diff
@@ -43,6 +43,7 @@
     <AndroidUseSharedRuntime>False</AndroidUseSharedRuntime>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Java.Interop" />
```
kjeremy: Is this needed?

MvvmCross/Binding/Droid/Target/MvxSearchViewQueryTextTargetBinding.cs
```diff
@@ -29,6 +29,8 @@ public override void SubscribeToEvents()
 
         protected override void SetValueImpl(object target, object value)
         {
+            if (target is SearchView sv)
```
kjeremy: Do we need this check? We shouldn't be getting here if target is null.

Cheesebaron: It is instead of doing:

```
var sv = target as SearchView;
if (sv != null)
{
}
```

Cheesebaron: I can change it to explicit cast instead, if you prefer.

MvvmCross/Binding/Droid/MvvmCross.Binding.Droid.csproj
Cheesebaron: Everything is red without

MvvmCross/Binding/Droid/Target/MvxSearchViewQueryTextTargetBinding.cs
kjeremy: It's already guaranteed to be a `SearchView` and it can't be null (the caller checks that it is not and that the Java handle is valid). Note that I'm on mobile if this sounds terse.

MvvmCross/Binding/Droid/MvvmCross.Binding.Droid.csproj
Cheesebaron: Might be a intellisense dud.

Cheesebaron:
Can you please review again @kjeremy?

Issue #1894, opened 2017-05-31
Appearing event called multiple times on Android platform


Bowman74:
### Steps to reproduce

1. Create a MvvmCross Android project

2. Have an Activity inherit from MvxAppCompatActivity<TypeofViewModel> and a view model that inherits from MvxViewModel

3. Register that view model as the app start view model in the MvxApplication class.

4. Add Debug.WriteLine statements to the view model in the lifecycle events

5. Run the app and view the application output.

### Expected behavior
Appeared event should only be called once when screen is loading (this is how it works in the iOS application)

### Actual behavior
Appeared event called multiple times

Relevant log output:
mvx:Diagnostic:  3.05 Attempting to load new ViewModel from Intent with Extras
CustomerListViewModel: instance: c2d7695d-212b-466e-bc48-6c792bb06d71: constructor
CustomerListViewModel: instance: c2d7695d-212b-466e-bc48-6c792bb06d71: start
CustomerListViewModel: instance: c2d7695d-212b-466e-bc48-6c792bb06d71: appearing
CustomerListViewModel: instance: c2d7695d-212b-466e-bc48-6c792bb06d71: appeared
CustomerListViewModel: instance: c2d7695d-212b-466e-bc48-6c792bb06d71: appeared
CustomerListViewModel: instance: c2d7695d-212b-466e-bc48-6c792bb06d71: appeared

One of the extra appeared calls seems to be tied to issue #1880. One call goes away if I implement the workaround I found, still leaving the appeared event getting called twice.

### Configuration
Debug

**Version:** 5.0.1

**Platform:** 
- [ ] iOS
- [x] Android
- [ ] WPF
- [ ] UWP

Solution that shows the problem:
https://github.com/Bowman74/VSLiveMvvmCross/tree/MvvmCrossError/src/Demo2/VsLiveMvvmCross

Cheesebaron:
This is the relevant piece of code where this event is fired: https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.AppCompat/MvxAppCompatActivity.cs#L72

kjeremy:
We probably need to check if the vto is alive before firing the event and
unsubscribing.

Bowman74:
I've done some investigation. What I suspect is happening is that when the Android screen is being measured three measurements happen in rapid succession and GlobalLayout events called before the event handler can be unsubscribed.

[...]

PR #1897, opened 2017-05-31
#1894 Fixed problem where C# version of event handlers do not unsubsc…

Bowman74:
…ribe from native events as expected. Instead switch to native style events to get expected behavior.

## :sparkles: What kind of change does this PR introduce? 
Bug fix

Android Appeared event fires multiple times.  

Android Appeared event only fires once

No

Check lifecycle events for activities.

#1894

- [x] All projects build

### DIFF
--- a/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.Fragment/MvxCachingFragmentActivity.cs
+++ b/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.Fragment/MvxCachingFragmentActivity.cs
@@ -59,16 +59,9 @@ namespace MvvmCross.Droid.Support.V4
 
 			base.OnCreate(bundle);
 
-            var rootView = Window.DecorView.RootView;
+            _view = Window.DecorView.RootView;
 
-            EventHandler onGlobalLayout = null;
-            onGlobalLayout = (sender, args) =>
-            {
-                rootView.ViewTreeObserver.GlobalLayout -= onGlobalLayout;
-                ViewModel?.Appeared();
-            };
-
-            rootView.ViewTreeObserver.GlobalLayout += onGlobalLayout;
+            _view.ViewTreeObserver.AddOnGlobalLayoutListener(this);
 
             if (bundle == null)
 				HandleIntent(Intent);
@@ -89,6 +82,13 @@ namespace MvvmCross.Droid.Support.V4
 				RestoreViewModelsFromBundle(serializer, bundle);
 			}
 		}
+
+        public override void SetContentView(int layoutResId)
+        {
+            var view = this.BindingInflate(layoutResId, null);
+
+            SetContentView(view);
+        }
 
 		protected override void OnNewIntent(Intent intent)
 		{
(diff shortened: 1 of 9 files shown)


kjeremy:
So... this worries me.  Is there a Xamarin bug this corresponds to?  If not then one needs to be filed.

Bowman74:
@kjeremy 
I have submitted a bug to Xamairn here:

https://bugzilla.xamarin.com/show_bug.cgi?id=57031

It also includes a link to a simple (non-MvvmCross) project that demonstrates the difference in behavior of these two methods of using events that handle the GlobalLayoutListener.

Bowman74:
@kjeremy I have updated the other Mvx components as you requested to use the native event style. Let me know if you need anything else done.

Cheesebaron:
I really don't like that we keep a reference to the inflated view. However, I guess that is the best way to be able to easily unsubscribe later on.

kjeremy:
Let's give it a few days and hear what xamarin has to say before merging.

Bowman74:
@Cheesebaron To be fair we were keeping a class level reverence to the inflated view before, just the way the code was written made it less obvious. We used the view reference scoped in the method in the event handler. What the compiler does so that reference doesn't go out of scope is elevate that to a class level reference. It's the source of a lot of inadvertent memory leaks.

kjeremy:
Out of curiosity does this fix the bug?

```csharp
EventHandler onGlobalLayout = null;
onGlobalLayout = (sender, args) =>
{
    var vto = (ViewTreeObserver)sender;
    if (vto.IsAlive)
    {
         vto.GlobalLayout -= onGlobalLayout;
         ViewModel?.Appeared();
    }
};
```

Stackoverflow suggests that you need to remove the listener from the `ViewTreeObserver` that is in use.

Bowman74:
@kjeremy Sorry, the code you proposed doesn't work. The version of the ViewTreeObserver sent in the sender parameter of the event is never alive and it isn't the same instance the event was added to so the event couldn't be removed from it. 

I could add the IsAlive code before removing the event handler in the new code if you want. Having said that, based on the lifecycle, there should never be a case where IsAlive isn't true.

thongdoan:
Should we change to below ?, We will add a "raised" flag, and on next version of Xamarin maybe they will fix it
```
public override void SetContentView(int layoutResId)
        {
            var view = this.BindingInflate(layoutResId, null);

            var weakThis = new WeakReference(this);
            var weakView = new WeakReference(view);
            var weakOnGlobalLayout = new WeakReference(null);

            bool raised = false;
            EventHandler onGlobalLayout = null;
            onGlobalLayout = (sender, args) => {
                if (raised)
                    return;
                raised = true;
                var strongThis = weakThis.Target as MvxAppCompatActivity;
                var strongView = weakView.Target as View;
                var strongOnGlobalLayout = weakOnGlobalLayout.Target as EventHandler;

                if (strongOnGlobalLayout != null && strongView != null) {
                     strongView.ViewTreeObserver.GlobalLayout -= strongOnGlobalLayout;
                }
                strongThis?.ViewModel?.Appeared();
            };

            weakOnGlobalLayout.Target = onGlobalLayout;
            view.ViewTreeObserver.GlobalLayout += onGlobalLayout;

            SetContentView(view);
        }
```

Bowman74:
@kjeremy I made the changes you requested to check the IsAlive variable. I also clear out the _view pointer when the even is unhooked as it is no longer needed.

PR #1899, opened 2017-06-01
Adds support for INotifyPropertyChanged to MvxWithEventPropertyInfoTargetBinding

DanielStolt:

New feature.

Two-way bindings only work if the target implements an event named `[PropertyName]Changed` with respect to the target property name.

This change adds support to `MvxWithEventPropertyInfoTargetBinding` for also falling back to the more generic `INotifyPropertyChanged` interface if a more specific event was not found.

Not as far as I can tell.

Two-way bindings in general, and two-way bindings to `INotifyPropertyChanged` targets in particular.

No issues found for this.

This documentation is relevant, but does not explicitly mention this aspect as far as I can tell:
https://github.com/MvvmCross/MvvmCross/blob/develop/docs/_documentation/fundamentals/data-binding.md

- [X] All projects build

### DIFF
--- a/MvvmCross/Core/Binding/Bindings/Target/MvxWithEventPropertyInfoTargetBinding.cs
+++ b/MvvmCross/Core/Binding/Bindings/Target/MvxWithEventPropertyInfoTargetBinding.cs
@@ -1,4 +1,4 @@
-// MvxWithEventPropertyInfoTargetBinding.cs
+﻿// MvxWithEventPropertyInfoTargetBinding.cs
 
 // MvvmCross is licensed using Microsoft Public License (Ms-PL)
 // Contributions and inspirations noted in readme.md and license.txt
@@ -8,6 +8,7 @@
 namespace MvvmCross.Binding.Bindings.Target
 {
     using System;
+    using System.ComponentModel;
     using System.Reflection;
 
     using MvvmCross.Platform;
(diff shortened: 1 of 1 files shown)


MvvmCross/Core/Binding/Bindings/Target/MvxWithEventPropertyInfoTargetBinding.cs
```diff
@@ -53,26 +71,64 @@ public override void SubscribeToEvents()
...
-            var eventName = this.TargetPropertyInfo.Name + "Changed";
+
+            // Try a specific [PropertyName]Changed event first.
+            var eventInfo = GetNamedPropertyChangedEvent(viewType, this.TargetPropertyInfo.Name);
+            if (eventInfo != null)
+                this._subscription = eventInfo.WeakSubscribe(target, OnValueChanged);
+            else if (target is INotifyPropertyChanged notifyPropertyChangedTarget)
+            {
+                // If target implements INPC then try a generic PropertyChanged event next.
+                eventInfo = GetPropertyChangedEvent(viewType);
```
DanielStolt: Checking for the actual event and constructing `EventInfo` is not strictly necessary in this case since the subscription only needs the target and we have already checked for the `INotifyPropertyChanged` interface, but it seemed more consistent to do it this way. Let me know if you prefer to have it changed.

Issue #1880, opened 2017-05-29
Navigation Bug introduced in 5.0.1: View.OnCreate reinstantiates existing target ViewModel 


daniatic:
I'm observing a bug in the navigation since 5.0.1: Find a sample project [here](https://github.com/daniatic/mvvmcross-navigation-bug) 

### Steps to reproduce

1. From a ViewModel navigate to another ViewModel and passing a parameter to it by: `navigationService.Navigate<TargetViewModel,string>("test")`

2. The TargetViewModel gets instantiated and afterwards the Initialize method is invoked passing the parameter. Save that parameter to variable. 

3. Now the TargetViewModel gets instantiated again by its view.OnCreate method. The old instance with the passed parameters gets overridden and the saved variable is gone.

### Expected behavior
The second instantiation shouldn't happen. Like it works in v5.0.0

### Configuration

**Version:** 5.0.1

**Platform:** 
- [ ] iOS
- [ ] Android
- [ ] WPF
- [ ] UWP

Bowman74:
I have also encountered this same issue.

Bowman74:
I traced this problem to line 48 in the MvxAndroidViewPresenter:

requestTranslator.RemoveSubViewModelWithKey(instanceRequest.Item2);

I created a custom view presenter that is an exact copy of the 5.0.1 view presenter and corrected the problem by commenting out this line.

@martijn00 I noticed that you added the line as part of commit #1871 . I assume simply removing it will cause a regression for the fragment code?

Bowman74:
@mvanbeusekom I hope someone has a chance to review my PR #1907 before the next release or the solution to this bug will be replaced with a memory leak. @martijn00 was right that the ViewModel needed to be removed from the cache, it was just happening in the wrong spot.

arogan:
Please make sure this is also fixed for iOS.  I was able to reproduce the double viewmodel instance issue both in 5.0.0 and 5.0.1.  Sample code is here:  https://github.com/arogan/HelloMvx5Forms

mvanbeusekom:
Looks like this is a problem when using MvvmCross with Xamarin.Forms. I have not encountered this issue with a Xamarin.iOS application (using the MvxIosViewPresenter)

JTOne123:
The issue is reproducing on 5.0.6 in Forms

[...]

PR #1871, opened 2017-05-26
Fix for view models not correctly restored in fragments

martijn00:
Fixes:
- https://github.com/MvvmCross/MvvmCross/issues/1817
- https://github.com/MvvmCross/MvvmCross/issues/1855

PR #1907, opened 2017-06-02
Kevinf/1880 memory leak

Bowman74:
Removes memory leak

Android now currently and correctly maps view models from the navigator into the activity but never cleans up the cache where the view models are stored for the activity to grab when it is done loading. This has caused a memory leak that I verified with the profiler as the ViewModels mapped this way now stay in the cache forever. 

The original person who introduced the bug was correct that the cache needed to be cleaned up, it was just done it in the wrong place.

ViewModel is removed from the cache, removing the memory leak. The fix was verified using profiler.

No 

Verify view models created using NavigationService passing parameters are being cleaned up correctly.

- [x] All projects build

### DIFF
--- a/MvvmCross/Droid/Droid/Views/MvxAndroidViewsContainer.cs
+++ b/MvvmCross/Droid/Droid/Views/MvxAndroidViewsContainer.cs
@@ -114,8 +114,11 @@ namespace MvvmCross.Droid.Views
             {
                 {
                     mvxViewModel = Mvx.Resolve<IMvxChildViewModelCache>().Get(embeddedViewModelKey);
-                    if(mvxViewModel != null)
+                    if (mvxViewModel != null)
+                    {
+                        RemoveSubViewModelWithKey(embeddedViewModelKey);
                         return true;
+                    }
                 }
             }
             mvxViewModel = null;


mvanbeusekom:
This is a bit hard for me to verify, but have you made sure the bug described in issue #1880 is not reintroduced by this PR?

Bowman74:
@mvanbeusekom Yes, I'm sure it is not been reintroduced. What caused #1880 was that the ViewModel was removed from the cache in the presenter and the activity was being loaded asynchronously. When the activity was ready it asks the views container for the ViewModel instance but it is already removed from the cache and creates a new one. The fix only removes it right after the Activity has requested to get it out out of the cache.

I do have my VSLive presentation that encountered this problem and has the solution implemented here:
https://github.com/Bowman74/VSLiveMvvmCross

It uses a custom presenter and a custom views container for Android that has the fixes in it (the current fix for #1880 in the presenter and my fix to the fix in the views container). If you remove both overrides to use the custom versions from the setup class the original #1880 problem will return when drilling down on list items. 

If you put back in just the custom presenter (with the current fix) you will see that the problem goes away but if you monitor with profiler you will also see the EditCustomerViewModel instances can never be garbage collected if you keep drilling down on list items. 

If you then add my fix you can verify the fix to #1880 still works but if you also watch with profiler you can verify that when garbage collection eventually happens the EditCustomerViewModel instances will be able to be cleaned up.

Luckily the MvvmCross ability to replace out most framework components with custom implementations make these kind of fixes easy to test. ;)

martijn00:
@Bowman74 I think that you are right with this fix, but just to verify, https://github.com/MvvmCross/MvvmCross/pull/1871/files#diff-359ac96e4d58f2919d9c9a441b49e3dfR219 i added something to remove as well for the fragments. Should we revert that, or is it still valid?

Bowman74:
@martijn00 We would need a sample that uses fragments to be 100% sure but I *think* we are OK, here's why. 

The code I modified is called from MvxActivityViewExtensions.OnViewCreate which applies to things of type IMvxAndroidView. That handler is only mapped if the MvxActivityViewExtensions. AddEventListeners method is called. The MvxActivity, MvxAppCompatActivity do call it in their constructors. 

The MvxCachingFragmentActivity that you modified inherits from MvxActivity but overrides that constructor, makes it protected, doesn't call into the base constructor nor calls AddEventListeners on its own. It does not appear to interact with the generic shared code I modified in any way and appears to completely have it's own management code for the ViewModels in the cache. In other words the change I made *should* not make better or worse the handling of the ViewModels in the cache used by MvxCachingFragmentActivity or be called by it in any way.

Whew, that just took me about two hours if digging through the code base to figure out what was happening and I'd say I'm only about 90% sure. I'm starting to know more about how this works under the hood that I expected. Lol.

martijn00:
@Bowman74 The RoutingExample(https://github.com/MvvmCross/MvvmCross/tree/develop/TestProjects/Navigation) contains fragments. You could use that to test.

I happy you take the time to help out. We all really appriciate it!

Bowman74:
@martijn00 OK, it still works. The SecondHostView does call into MvxAndroidViewsContainer.TryGetEmbeddedViewModel but for whatever reason in your code there is never an embedded ViewModel key and so my new line of code is never called. I'm not sure if that is due to something specific with the example. So right now the new code doesn't interact with the code you added.

PR #1930, opened 2017-06-09
Don't mask exceptions when completing android initialization

edwinvanderham:
Currently, if an exception occurs during the execution of the Start method of the class implementing IMvxAppStart on Android, this exception is masked and the Splash screen is never replaced with the first app screen. The expected behaviour would be that the app crashes on the unexpected exception. This makes it easier for developers to tackle bugs they may have introduced.

No exception is thrown when startup fails because it is wrapped.

The startup will throw and fail when an exception occurs.

No the parameter is optional and defaults to the old behaviour.

- [ ] All projects build


Cheesebaron:
Please use the Pull Request template https://github.com/MvvmCross/MvvmCross/blob/develop/.github/PULL_REQUEST_TEMPLATE.md to describe what this changes.

PR #1937, opened 2017-06-10
PreferredContentSize

g0rdan:
Feature

Now you can use 4 parameters in MvxModalPresentationAttribute

PR will add new, fifth paramater - PreferredContentSize. It can set content size of your modal view.

Nope

I'm not sure if it works on iPhone because minimum width of content size should be 320. It definitely works on iPad.

[Official doc](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621476-preferredcontentsize)

- [x] All projects build

### DIFF
--- a/MvvmCross/iOS/iOS/Views/Presenters/Attributes/MvxModalPresentationAttribute.cs
+++ b/MvvmCross/iOS/iOS/Views/Presenters/Attributes/MvxModalPresentationAttribute.cs
@@ -1,4 +1,5 @@
-﻿using UIKit;
+﻿using CoreGraphics;
+using UIKit;
 
 namespace MvvmCross.iOS.Views.Presenters.Attributes
 {
@@ -9,6 +10,8 @@ namespace MvvmCross.iOS.Views.Presenters.Attributes
         public UIModalPresentationStyle ModalPresentationStyle { get; set; } = UIModalPresentationStyle.FullScreen;
 
         public UIModalTransitionStyle ModalTransitionStyle { get; set; } = UIModalTransitionStyle.CoverVertical;
+
+        public CGSize PreferredContentSize { get; set; }
 
         public bool Animated { get; set; } = true;
     }
--- a/MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
+++ b/MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
@@ -1,12 +1,12 @@
-﻿﻿using System;
- using System.Collections.Generic;
- using System.Linq;
- using CoreGraphics;
- using MvvmCross.Core.ViewModels;
- using MvvmCross.iOS.Views.Presenters.Attributes;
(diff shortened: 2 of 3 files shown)


Cheesebaron:
Thank you for your Pull Request. Please use the Pull Request template, when creating a Pull Request. You can find it here: https://github.com/MvvmCross/MvvmCross/blob/develop/.github/PULL_REQUEST_TEMPLATE.md

g0rdan:
@Cheesebaron Sure, done.

Cheesebaron:
Looks good to me. Can someone else also review this please?

nmilcoff:
@Cheesebaron 👍

mvanbeusekom:
I like the feature and it has an obvious use case. Just to note that we also solved similar options through adding an interface. This might be a nice alternative since this option is only applicable for iPad and adding it to the attribute might confuse developers in case of iPhone solutions

Cheesebaron:
@mvanbeusekom it is not obvious how to do exactly that reading the documentation.

mvanbeusekom:
@cheesebaron can you elaborate which part is not obvious, maybe we can improve the documentation?

Cheesebaron:
Reading the docs about overriding the presentation attributes at runtime: https://www.mvvmcross.com/documentation/platform/ios-view-presenter?scroll=720#override-a-presentation-attribute-at-runtime It only allows you to pass along presentation attributes. So passing along your own presentation attributes here wont work either. Until, you actually override the presenter too, which then responds to your custom attributes.

It is a lot more work, compared to just have out of the box support for a PreferredContentSize property, which might only work on iPad.

The docs may be fine for this and readers smart enough would know how to combine both overriding the presenter and adding their own attributes. However, I would love to see a more in depth example, which is not just using the default attributes.

mvanbeusekom:
I see what you mean, unfortunately what I mean is not documented (so this will be a nice task). For example to pas additional settings to the Xamarin Sidebar implementation you can implement the [`IMvxSidebarMenu`](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/Views/IMvxSidebarMenu.cs) interface. This will allow you to add additional behaviour to the [`MvxSidebarPresentationAttribute`](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/Attributes/MvxSidebarPresentationAttribute.cs). In these case these are settings like:
- AnimateMenu
- DisablePanGesture
- DarkOverlayAlpha
- HasDarkOverlay
- HasShadowing
- MenuButtonImage
- etc...

These are all similar to the settings added by this pull-request, hence my suggestion to keep things in sync.

Cheesebaron:
Alright. We can undo this one. We just need to make it more clear in the docs how to do these things.

mvanbeusekom:
I'll start writing some additional documentation

Issue #1940, opened 2017-06-13
Navigation Issues UWP after Upgrade


smokingfish:
I tried upgrading a quite big (mainly) UWP project from 4.x to 5.0.2.
All ShowViewModel calls were replaced by IMvxNavigationService calls.

### Steps to reproduce
I attached a demo program.

### Expected behavior
**NavigationCacheMode = Disabled:**
As in 4.x, create new instance of ViewModel and View on each Navigate() and Close()
I manually managed resources when working like this because 4.x did not clear the cache or disposed the old ViewModels.

**NavigationCacheMode = Enabled**
Close() as it is now. Navigate() should create a new Viewmodel instance (currently does so) and set the DataContext to the new instance if the view is reused.

### Actual behavior
**NavigationCacheMode = Disabled:**
Using the last version Close() would create a new ViewModel instance and set the DataContext. After the Update the DataContext is not set. Callin Navigate with parameter successfully calls Initialize() and the parameter can be used for the newly created view.

**NavigationCacheMode = Enabled**
We wanted to use this to improve loading times and maybe as a workaround for the previous issue. Unexpected behavior: When calling Navigate() (with or without parameter) Initialize() gets called successfully and a new instance of the Viewmodel is created, but the DataContext of the cached view is still the previous copy of the Viewmodel.

### Configuration

Version: 5.0.2

Platform: 
- [ ] iOS
- [ ] Android
- [ ] WPF
- [x] UWP

[WebViewTestApp.zip](https://github.com/MvvmCross/MvvmCross/files/1071310/WebViewTestApp.zip)

smokingfish:
Maybe this helps a little bit: I updated to the latest unstable build (5.1.0unstable0086).

In this version it is even more broken :) 
With cache enabled it will create 2 Viewmodels on first Navigate(), call Initialize() on the version not set as datacontext. After that creates a new instance/calls Initialize() on Navigate(), but still uses cached version. No Action on back navigation using Close() from another view.

With no cache it always creates 2 instances, uses the version not bound to the view and calls Initialize(). On back navigation using Close() from another view creates new instance but does not call Initialize().

smokingfish:
I think it would be helpful if someone could take a look at the sample program and clarify if that is indeed the correct usage.

[...]

PR #1945, opened 2017-06-14
#1940 Made changes to UWP navigation so:

Bowman74:
Bug Fix

Navigating back through back stack to view displayed using the new navigation service looses the view model and view is displayed without a view model when NavigationCahceMode is disabled.

Navigating to a new view that has been shown before will use only the first view model instance when NavigationCacheMode is enabled. New calls to these views never replace the view model.

-Navigating through the back stack properly restores the existing instances of ViewModels to the new instances of the views that are created when NavigationCahceMode is disabled.

-Navigating to a new view that has been shown before will replace the current view model with the new one when NavigationCacheMode is enabled.

There is the possibility of the user manually removes a view from the back stack (something like Frame.BackStack.Remove()) that they will leave references to view models in the cache that will never be recovered until the application is unloaded. Calling things like Frame.GoBack() will work fine because all the normal navigation events are fired.

Implement in various navigation scenarios on UWP.

#1940

- [x] All projects build

### DIFF
--- a/MvvmCross/Windows/Uwp/Views/IMvxWindowsViewModelRequestTranslator.cs
+++ b/MvvmCross/Windows/Uwp/Views/IMvxWindowsViewModelRequestTranslator.cs
@@ -17,5 +17,7 @@ namespace MvvmCross.Uwp.Views
         string GetRequestTextWithKeyFor(IMvxViewModel existingViewModelToUse);
 
         void RemoveSubViewModelWithKey(int key);
+
+        int RequestTextGetKey(string requestText);
     }
 }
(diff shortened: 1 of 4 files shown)


ryssel:
Is there a nightly build somewhere where I can get this fix as a nuget package.

martijn00:
@ryssel Nightly builds are available on Myget: https://www.myget.org/F/mvvmcross/api/v3/index.json

ryssel:
Thnx, about what time does it build?

martijn00:
It should build after every merged PR but something seems to be off. I think we have to check this.

PR #1948, opened 2017-06-15
Add generic interfaces for IMvxCommand

### LINKED ISSUE #1946: Add generic interface for IMvxCommand
We should add generic interfaces for `IMvxCommand` and `IMvxAsynxCommand` to be able to use the interface as type for the command. This would probably also enables us to provide compiler errors when types don't match and make the code a bit simpler.

willsb:
This adds the `IMvxCommand<T>` and the `IMvxAsyncCommand<T>` interfaces.

One doesn't have type safety when using the command interfaces

Some type safety is added

Nope, but it could. The reason `MvxCommand<T>` implements both `IMvxCommand` and `IMvxCommand<T>` is because `IMvxCommandBuilder` would break horribly if it didn't. I didn't want to make this breaking, but I wasn't even aware of the existence of `IMvxCommandBuilder`. 

Is it used? Honest question, I do not know.

Please verify Windows projects are working fine, since I'm on a Mac 👍 

Closes #1946 

- [ ] All projects build

### DIFF
--- a/MvvmCross/Core/Core/ViewModels/IMvxCommand.cs
+++ b/MvvmCross/Core/Core/ViewModels/IMvxCommand.cs
@@ -5,17 +5,30 @@
 //
 // Project Lead - Stuart Lodge, @slodge, me@slodge.com
 
+using System;
 using System.Windows.Input;
 
 namespace MvvmCross.Core.ViewModels
 {
-    public interface IMvxCommand
-        : ICommand
+    public interface IMvxCommand : ICommand
     {
         void RaiseCanExecuteChanged();
 
         void Execute();
 
         bool CanExecute();
+    }
+
+    public interface IMvxCommand<T> : ICommand
+    {
+        [Obsolete("Use the strongly typed version of Execute instead", true)]
+        new void Execute(object parameter);
+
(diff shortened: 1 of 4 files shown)


MvvmCross/Core/Core/ViewModels/IMvxCommand.cs
```diff
@@ -5,17 +5,30 @@
...
+    public interface IMvxCommand : ICommand
     {
         void RaiseCanExecuteChanged();
 
         void Execute();
 
         bool CanExecute();
     }
+
+    public interface IMvxCommand<T> : ICommand
```
martijn00: Should this not inherit `IMvxCommand` instead of `ICommand`?

willsb: We don't want people to use the parameterless `Execute` and `CanExecute`

SlyNet: In v4 we had a custom binding that bound `IMvxCommand` to control events, and it worked for both - generic, and non generic commands. In v5 two bindings are needed?

PR #1950, opened 2017-06-15
[Feature] Add MvxIosFlexibleViewPresenter

kvandake:
#### Current behavior:
* We can have only one modal nesting;
* Many private view controllers: _MasterNavigationController_, _ModalNavigationController_, _TabBarViewController_, _SplitViewController_;
* Binding to Window.RootViewController.

#### New behavior:
* A lot of modal view controllers;
* One view controller for control navigation state;
* Not binding with Window.RootViewController.

#### Does this PR introduce a breaking change? 
- No, only feature.

#### Recommendations for testing:
- Build Configuration: **Playground.iOS**|**Debug-FlexiblePresenter**.


martijn00:
Why would we need a new presenter to add this? Can't we just add this functionality to the existing one?

nmilcoff:
It isn't clear to me:

- What the problem / issues are with the current ViewPresenter
- What are the improvements of this ViewPresenter

I also think we'd try to improve the existing one instead of managing two ViewPresenters (that was probably why we did a new one for Mvx 5.0 in the first place)

kvandake:
I agree that it is better to support only one Presenter.  If I add functionality to an existing presenter, then the derived classes from MvxIosViewPresenter will be affected.

kvandake:
What are the improvements of this ViewPresenter

New behavior:
* A lot of modal view controllers;
* One view controller for control navigation state;

For example, a [problem](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs#L212). Why problem? A developer want to show ViewModel, but the presenter dismiss current modal view controller.

nmilcoff:
This one:

- A lot of modal view controllers;

Can be achieved by holding a reference to the presented ViewControllers (maybe `List<UIViewController> presentedViewControllers`?). It can be quite unstable in my opinion though, since a presented ViewController can be dismissed by itself and then the history of views held by the ViewPresenter would be wrong. This is a potential problem that is also present in this FlexibleViewPresenter. 

- One view controller for control navigation state;

Why would you say this is a problem? We're cleaning all references when they're not used anymore.
Also iOS apps have at most one window, so I don't see why holding a reference to it would be an issue.

kvandake:
Can be achieved by holding a reference to the presented ViewControllers (maybe List<UIViewController> presentedViewControllers?). 

Why, if there is a property PresentingViewController from UIViewController. We can see PresentingViewController, ParentViewController...

nmilcoff:
You're right, but think about closing a presented ViewController with no more information than a [ViewModel instance](https://github.com/MvvmCross/MvvmCross/blob/f9f6e2643d2defc6ccc033f899a8ecf599441e67/MvvmCross/Core/Core/Views/IMvxViewPresenter.cs#L21). It's a real mess to explore the entire views hierarchy and cover all possibilities.

kvandake:
We can [control closing by using steps](https://github.com/kvandake/MvvmCross/blob/develop/MvvmCross/iOS/iOS/Views/Presenters/MvxIosFlexibleViewPresenter.cs#L436):
* Try close modal view controller;
* Try close child view controller;

Well, if you use List <UIViewController> presentedViewControllers, then how does this help control if we only have one visible view controller?

nmilcoff:
Also suppose you want to close a Tab from a TabBarController, while having a ViewController presented as modal. This version doesn't support that scenario.

nmilcoff:
Having a list of presented ViewControllers was just an idea, each item of the list could be explored to see what it is, if it has a stack inside of it, etcetera. If the ViewModel instance is found there -> close it.

kvandake:
We [check](https://github.com/kvandake/MvvmCross/blob/develop/MvvmCross/iOS/iOS/Views/Presenters/MvxIosFlexibleViewPresenter.cs#L466) the presence of the attribute MvxModalPressionAttribute before closing the modal view. We can try to reproduce errors in Playground.iOS. I also created a test project [MvvmCross.Test.zip](https://github.com/MvvmCross/MvvmCross/files/1078872/MvvmCross.Test.zip)

mvanbeusekom:
I am trying to review this pull-request but i am struggling a bit. Personally I cannot get a grip on the actual use-cases that this pull-request is trying to solve (on a functional level). Could you add some more detailed information in your description to answer the following questions:

- what is the benefit of using only one view controller (compared to the separated view controllers we have now)? At the moment I feel that the code gets difficult to read because of the use of the different lambda expressions need to be supplied to handle the different actions per type of view controller.
- why should we avoid binding to the `Window.RootViewController`?

It also feels that you are trying to fix multiple issues inside the current iOS view presenter all at once. I agree with @nmilcoff here that is would be better to improve / expand the existing view controller and take each issue one at a time. This will also make it easier for us to review the pull-request.

To add to the discussion on how to manage multiple modal view controllers, Greg Shackles has a good [blog post](https://gregshackles.com/presenters-in-mvvmcross-nested-modal-navigation-in-ios/) on how he solved this issue, maybe this could be a good solution to integrate as well.

kvandake:
Thanks for the detailed question.
> what is the benefit of using only one view controller (compared to the separated view controllers we have now)? At the moment I feel that the code gets difficult to read because of the use of the different lambda expressions need to be supplied to handle the different actions per type of view controller. 

The [current implementation](ss/MvvmCross/blob/develop/MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs#L152) at each step of the **Show** and control each ModalNavigation/TabBar/MasterNavigation/Split view controllers:
```csharp
            if (ModalNavigationController != null)
            {
                ModalNavigationController.PushViewController(viewController, attribute.Animated);
                return;
            }

            if (TabBarViewController != null)
            {
                TabBarViewController.ShowChildView(viewController);
                return;
            }

            if (MasterNavigationController != null)
            {
                MasterNavigationController.PushViewController(viewController, attribute.Animated);
                return;
            }
            ...
```
The [lambda expressions](https://github.com/kvandake/MvvmCross/blob/develop/MvvmCross/iOS/iOS/Views/Presenters/MvxIosFlexibleViewPresenter.cs#L223) and control one view controller:
```csharp
            var result = ExpandCurrentViewController(
                navVc =>
            {
                navVc.PushViewController(viewController, attribute.Animated);
                return true;
            },
                tabBarVc =>
            {
                return tabBarVc.ShowChildView(viewController);
            },
                splitVc =>
            {
                return false;
            });
            ...
```

> why should we avoid binding to the Window.RootViewController?

We allow multiple modal view controllers.Then when you call the present view controller from the root controller is receiving a warning:
```
Warning: Attempt to present <ViewController: 0x7fa323e107a0> on <ViewController: 0x7fa323e11850> whose view is not in the window hierarchy!
```
> To add to the discussion on how to manage multiple modal view controllers, Greg Shackles has a good blog post on how he solved this issue, maybe this could be a good solution to integrate as well.

Why keep the stack if we can only DismissViewController from the top modal view controller. We can not cause a dismiss view controller from the middle of the stack.

nmilcoff:
I don't see a problem with holding multiple ViewControllers in the presenter. I also think it's easier to read as it is now, and probably easier to maintain / extend. 

@kvandake would you be able to add the "multi modal" feature to the current presenter? I don't think it's necessary to refactor the entire mechanics to add it. 
And then if you can provide solid benefits of changing how it works, you can make a separate PR for the current logic :)

kvandake:
Please see my changes. I removed the lambda expressions and I removed MvxIosFlexibleViewPresenter. I added an example with "multi modal" feature.

Cheesebaron:
Can you please elaborate why this is needed and what it solves? 

We appreciate Pull Requests a lot. However, this one has been made without pinging any of us. There is no link to any issues. You don't use the Pull Request template. The only description of what was changed is this:

> - A lot of modal view controllers;
> - One view controller for control navigation state;
> - Not binding with Window.RootViewController.

How can that describe a substantial change to the code in the presenter? The behavior must be entirely different from what it was, no?

Please, don't be discouraged by this comment, but please understand that we just want to understand what you are thinking and why you are trying to solve in this Pull Request.

kvandake:
Problem:
 A modal view controller opens another modal view controller. When we open the modal window, then change managers view controllers(Tab/Split/Navigation). For example, open a modal window, and this tab controller. 

This feature I have implemented in my project. The normal flight. I send pull request the first time, so could be wrong with the workflow pull request. I propose to make this feature in the question to fit the rules of the project.

kvandake:
The problem description:
A modal view controller cannot open another modal view controller. When we open the modal view controller, we change managers of the view controllers(Tab/Split/Navigation). For example, open a modal window, and this tab controller.

My pool request solves this problem. When we open the modal view controller, we change the managers of the view  controllers. This feature I have implemented in my project. The problem is solved. I send pull request the first time, so could be wrong with pull request. I propose to make this feature in the question to fit the rules of the project.

nmilcoff:
I'm closing this PR, as the feature was added in a different PR (which is esentially a clean version of this one). We appreciate the feature request, insight on the functionality and hope to see you here around again @kvandake :)

Issue #1943, opened 2017-06-14
New navigation service creates two instances of VM and initialize the wrong one


mgochmuradov:
Link to sample github project: [MvxNavigationServiceIssue](https://github.com/mgochmuradov/MvxNavigationServiceIssue)

### Steps to reproduce

1. Create Core and UI (WPF) project add nuget package (MvvmCross.StarterPack) on both projects

2. Create [SecondViewModel](https://github.com/mgochmuradov/MvxNavigationServiceIssue/blob/master/MvxNewNavigation.Core/ViewModels/SecondViewModel.cs)
```cs
    public class SecondViewModel : MvxViewModel
    {
        private static int _times = 0;
        public SecondViewModel() // <- ViewModel is being constructed TWICE
        {
            _times++;
            Title = "From Constructor " + _times;
        }

        private string _title;
        public string Title
        {
            get { return _title; }
            set { SetProperty(ref _title, value); }
        }

        public override Task Initialize()
        {
            // Called on a different instance than the one that is bound to view
            Title = "From Initializer";
            return base.Initialize();
        }
    }
```

3. On [MainViewModel](https://github.com/mgochmuradov/MvxNavigationServiceIssue/blob/master/MvxNewNavigation.Core/ViewModels/MainViewModel.cs) inject IMvxNavigationService, and add command to navigate to SecondViewModel
```cs
    public class MainViewModel : MvxViewModel
    {
        private readonly IMvxNavigationService _navigationService;
        public MainViewModel(IMvxNavigationService navigationService)
        {
            _navigationService = navigationService;
        }
        
        public override Task Initialize() // <- Never being called
        {
            return base.Initialize();
        }

        public IMvxAsyncCommand GoToSecondCommand =>
            new MvxAsyncCommand(GoToSecondAsync);
        private async Task GoToSecondAsync()
        {
            await _navigationService.Navigate<SecondViewModel>();
        }
    }
```

### Expected behavior
Construct `SecondViewModel`, navigate to `SecondView`, and call `Initialize()` method on `SecondViewModel`

### Actual behavior
`SecondViewModel` is being constructed twice, and one of them is set as BindingContext on `SecondView`, and `Initialize()` method is being called on the other one.

### Configuration

**Version:** 5.0.2

**Platform:** 
- [ ] iOS
- [ ] Android
- [X] WPF
- [ ] UWP

mgochmuradov:
This are the places where VM is being constructed:
[...]

Issue #1921, opened 2017-06-08
WPF Navigation Service doesn't work with parameter


bregol:
### Steps to reproduce

1. Create a WPF project

2. Navigate to new viewmodel passing a parameter using the new Navigation Service 

3. Have that viewmodel you are navigating to try to use the parameter in the Initialize method 

### Expected behavior
The viewmodel should have a single instance created and the parameter passed to it in the Initialize method.

### Actual behavior
Two viewmodels are created.  The first one created runs Initialize with the parameter passed to it.  Then a second version of the viewmodel is created and this one is shown but does not have the parameter passed to it.

This is the same thing for WPF that was happening for UWP here: https://github.com/MvvmCross/MvvmCross/issues/1905

### Configuration

**Version:** 5.0.2

**Platform:** 
- [ ] iOS
- [ ] Android
- [x] WPF
- [ ] UWP

Bowman74:
Likely this is the same issue with UWP and the cache may need to come into play.

Bowman74:
I'll see if I can't grab this in the next few days. Probably similar solution to UWP.

Bowman74:
I'll see if I can't grab this in the next few days. Probably similar solution to UWP.

PR #1963, opened 2017-06-16
#1921 Changed wpf implementation to use view model parameter sent wit…

Bowman74:
…h request by new navigation service. Updated wfp event hooks sample application to be more complete.

Bug Fix

When using the new navigation service the view model created with the request is ignored, instead the original ShowViewModel behavior is used creating a second view model ignoring the one from the request that has the parameter passed to it.

Only one view model is created when the new navigation service is used and the parameter properly passed to it.

No

Test in a variety of scenarios for navigation. I tested about three of the many possible scenarios

#1921, #1943 

- [x] All projects build

### DIFF
--- a/MvvmCross/Windows/Wpf/Views/MvxWpfViewsContainer.cs
+++ b/MvvmCross/Windows/Wpf/Views/MvxWpfViewsContainer.cs
@@ -37,8 +37,15 @@ namespace MvvmCross.Wpf.Views
             if (viewControl == null)
                 throw new MvxException("Loaded View is not a FrameworkElement " + viewType);
 
-            var viewModelLoader = Mvx.Resolve<IMvxViewModelLoader>();
-            wpfView.ViewModel = viewModelLoader.LoadViewModel(request, null);
+            if (request is MvxViewModelInstanceRequest instanceRequest)
+            {
+                wpfView.ViewModel = instanceRequest.ViewModelInstance;
+            }
+            else
+            {
+                var viewModelLoader = Mvx.Resolve<IMvxViewModelLoader>();
+                wpfView.ViewModel = viewModelLoader.LoadViewModel(request, null);
+            }
 
             return viewControl;
         }
--- a/TestProjects/Eventhooks/Eventhooks.Wpf/App.xaml.cs
+++ b/TestProjects/Eventhooks/Eventhooks.Wpf/App.xaml.cs
@@ -1,4 +1,8 @@
-﻿using System.Windows;
+﻿using MvvmCross.Core.ViewModels;
+using MvvmCross.Platform;
+using MvvmCross.Wpf.Views;
+using System.Windows;
+using System;
 
 namespace Eventhooks.Wpf
 {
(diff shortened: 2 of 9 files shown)


nickrandolph:
We're seeing a similar problem with using Master-Details in a Forms app when navigating between details pages. An alternative to this implementation is to move the check for existing view model into the LoadViewModel method - this would fix this issue for all platforms.

martijn00:
@nickrandolph Hmm you might be right about that one! We have to look into it.

nickrandolph:
I just tried this, which seems to solve this issue (in MvxPresenterHelpers), but this is specific to Mvvmcross for Forms. Wonder if it could be pushed into the viewModelLoader?

public static IMvxViewModel LoadViewModel(MvxViewModelRequest request)
{
    if (request is MvxViewModelInstanceRequest instance) return instance.ViewModelInstance
    var viewModelLoader = Mvx.Resolve<IMvxViewModelLoader>();
    var viewModel = viewModelLoader.LoadViewModel(request, null);
    return viewModel;
}

Bowman74:
@nickrandolph That would not work for all platforms. Not all platforms can pass the request instance to the view and instead keep the view model in a cache location that needs to be handled completely differently with the position in the cache passed to the view instead (UWP and Android both have variations on this).

nickrandolph:
Yep fair enough. I didn't have time to do a full test on that suggestion (hence no pull request). 

This issue is actually pretty severe as it is completely screwing up navigation on Forms using Master Details. You can see this if you attempt to navigate to a different view model from one of the details view models in the forms master-detail sample project

Bowman74:
Unfortunately about to board a flight. I can look at that problem later. I suspect there may be a custom presenter in play that may need modification. Can you send me a link to the sample (just to be sure we are looking at the same one)?

Bowman74:
@nickrandolph, I do not see a master detail example under TestProjects\Forms\ directory for WPF? Is there some WPF example in another location? Nor do I even find any MvvmCross support for Forms in general for WPF?

If not this PR only impacts WPF and is intended to fix a WFP specific bug so any separate issues should likely be handled under another issue.

Cheesebaron:
There is no WPF support from Xamarin.Forms to begin with. So dunno how what @nickrandolph talks about relates to this issue.

kjeremy:
Not yet, but I believe they're working on it https://github.com/xamarin/Xamarin.Forms/pull/895

Cheesebaron:
@kjeremy well still not related to this issue though. @nickrandolph if you have an issue with Forms, make a new issue and just link to this.

nickrandolph:
I'll raise another issue later today but my observation stands. The issue which is fixed by this commit seems to be mirrored (and admittedly I don't know the code base well enough to know whether the root cause is the same) by breaking functionality when using Mvvmcross for forms with master detail.

Bowman74:
@nickrandolph Yes a new issue would be great. If you look at what files this commit changes they are all in WPF specific projects which are not used for the Forms projects in any way, shape or form.

PR #1984, opened 2017-06-22
added features to be able to provided the selected image on tab item

biozal:
feature

you can set a tab bar item's image, but not the selected image

you can provide the selected image icon

I don't think so because the string for selected image can be null and if it's null then it's not set on the tab bar item

I updated the code in the play ground for testing

I mentioned this in slack to the team

- [X] All projects build


biozal:
@nmilcoff would the recommendation be to leave current methods in place and just have them call new methods with the new signature?  Then this wouldn't break as the old method signatures would still be there and they would simply call the new ones.

PR #1997, opened 2017-06-27
Updated the XamarinSidebar sample to use the navigationservice

mvanbeusekom:
Updates the XamarinSidebarSample test project to use the new navigationservice. This change was triggered by a complaint from @martijn00 that the XamarinSidebar implementation doesn't seem to work with the navigation service. 

Current implementation uses the old `ShowViewModel<TViewModel>()` method to trigger a navigation request.

## :new: What is the new behaviour (if this is a feature change)?
Use the `IMvxNavigationService` implementation to trigger a navigation request

No

Run the "MvvmCross.iOS.Support.XamarinSidebarSample.iOS" iOS-Support test project (part of the MvvmCross_All.sln

none

- [x] All projects build

### DIFF
--- a/TestProjects/iOS-Support/XamarinSidebar/MvvmCross.iOS.Support.Core/App.cs
+++ b/TestProjects/iOS-Support/XamarinSidebar/MvvmCross.iOS.Support.Core/App.cs
@@ -1,4 +1,6 @@
 using MvvmCross.Core.ViewModels;
+using MvvmCross.Core.Navigation;
+using MvvmCross.Platform;
 using MvvmCross.Platform.IoC;
 
 namespace MvvmCross.iOS.Support.XamarinSidebarSample.Core
@@ -13,7 +15,7 @@ namespace MvvmCross.iOS.Support.XamarinSidebarSample.Core
                 .AsInterfaces()
                 .RegisterAsLazySingleton();
 
-            RegisterAppStart(new AppStart());
+            RegisterAppStart(new AppStart(Mvx.Resolve<IMvxNavigationService>()));
         }
     }
 }
--- a/TestProjects/iOS-Support/XamarinSidebar/MvvmCross.iOS.Support.Core/App.cs
+++ b/TestProjects/iOS-Support/XamarinSidebar/MvvmCross.iOS.Support.Core/App.cs
@@ -15,7 +15,11 @@ namespace MvvmCross.iOS.Support.XamarinSidebarSample.Core
                 .AsInterfaces()
                 .RegisterAsLazySingleton();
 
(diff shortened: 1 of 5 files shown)


TestProjects/iOS-Support/XamarinSidebar/MvvmCross.iOS.Support.Core/App.cs
```diff
@@ -13,7 +15,7 @@ public override void Initialize()
                 .AsInterfaces()
                 .RegisterAsLazySingleton();
 
-            RegisterAppStart(new AppStart());
+            RegisterAppStart(new AppStart(Mvx.Resolve<IMvxNavigationService>()));
```
martijn00: Why not use construct to create the appstart so it injects the navigation service?

mvanbeusekom: I thought it didn't work, will update and check. Thx

mvanbeusekom: Turns out you'll need to resolve it somewhere since the `App` class is constructed in the platform specific `Setup` class and `Setup` is constructed in the `AppDelegate` class (in case of iOS). 

Resolving is in `AppDelegate` is to early (since the service is not registered yet at this point). The other option is to resolve it in the `Setup` class but this would mean you would potentially duplicate code (of course not here since it is a iOS only example, but for other projects this might be the case). So I think the best place would be to resolve it where it currently is. Of course I am open for suggestions in case I am missing something.

martijn00: Why not just do:
```
Mvx.ConstructAndRegisterSingleton<IMvxAppStart, AppStart>();
            var appStart = Mvx.Resolve<IMvxAppStart>();

            // register the appstart object
            RegisterAppStart(appStart);
```

TestProjects/iOS-Support/XamarinSidebar/MvvmCross.iOS.Support.Core/AppStart.cs
```diff
@@ -1,17 +1,32 @@
...
     {
+        private readonly IMvxNavigationService _navigationService;
+        private Func<IMvxNavigationService> resolve;
+
+        public AppStart(IMvxNavigationService navigationService)
+        {
+            _navigationService = navigationService ?? throw new ArgumentNullException(nameof(navigationService));
+        }
+
+        public AppStart(Func<IMvxNavigationService> resolve)
```
martijn00: This can then be removed?

PR #2002, opened 2017-06-29
Move the sidebar attribute to the root to avoid annoyance with usings

martijn00:
When adding an attribute for the Sidemenu you need include 2 usings. This is a bit annoying, and because we have only 1 attribute, i think we should move this to the root.

2 usings

1 usings

A little bit since 1 using will become unnecessary, but since we already use the other one it will just continue to work.

- [x] All projects build


nmilcoff:
@martijn00 I think the solution doesn't build with this change, because of the missing namespace. Could you remove the usings for it?

martijn00:
@nmilcoff fixed that!

Issue #2018, opened 2017-07-06
Feature request: Lifecycle event for OnCreate and ViewDidLoad 


KennethKr:
MvvmCross 5.0 introduced new lifecycle events, Appearing/Appeared/Disappearing/Disappeared. 

A great addition would be events for the "view created" event, like Android's OnCreate and iOS ViewDidLoad.

Events for "view destroyed" could also be included, but iOS ViewDidUnload is deprecated, and Android's OnDestroy aren't really guaranteed to be called, so I'm not sure about this part.

PR #2020, opened 2017-07-06
Introduced new lifecycle event "Created" in MvxViewModel

KennethKr:

Feature (solves #2018)

New lifecycle event Created for MvxViewModel. On Android this maps to OnCreate, on iOS ViewDidLoad and UWP OnNavigatedTo.

Renamed existing View events with prefix View. ViewAppearing, ViewAppeared, ViewDisappearing, ViewDisappeared.

Event does not exist.

New lifecycle event Created for MvxViewModel.

Yes. Renamed existing View events with prefix View. ViewAppearing, ViewAppeared, ViewDisappearing, ViewDisappeared.

Feature (solves #2018)

- [x] All projects build

### DIFF
--- a/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.Fragment/MvxCachingFragmentActivity.cs
+++ b/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.Fragment/MvxCachingFragmentActivity.cs
@@ -29,7 +29,7 @@ using MvvmCross.Platform.Platform;
 namespace MvvmCross.Droid.Support.V4
 {
     [Register("mvvmcross.droid.support.v4.MvxCachingFragmentActivity")]
-    public class MvxCachingFragmentActivity 
+    public class MvxCachingFragmentActivity
         : MvxFragmentActivity, IFragmentCacheableActivity, IMvxFragmentHost
     {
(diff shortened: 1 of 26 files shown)


willsb:
Awesome, though I find the naming a bit confusing. Since the method runs on the ViewModel, it seems like it runs when the VM is created, which is not the case. Maybe renaming it to `ViewCreated` or something else might make the use case clearer.

KennethKr:
Yeah, agreed. Tho, it raises questions for the other view-coupled events. Should they be prefixed with View as well? Like Destroy, or the Appearing-events (altho one can argue those are actually valid for the ViewModel as well).

modplug:
What if we prefix the lifecycle events with ViewLifecycleEvent?

martijn00:
I agree that the naming might not be too clear. We can change this to prepend `View` to all the names, but since that will be breaking we should include it in 5.1.0.

KennethKr:
Agreed. I'll make a new commit including the View prefix for all events.

KennethKr:
There it is. I renamed all current events with a View prefix.

willsb:
Now someone just needs to label this with `do-not-merge` and `5.1.X` :)

PR #2033, opened 2017-07-12
Stop using unsupported applicationDidFinishLaunching on iOS.

DanielStolt:

Bug fix/hygiene

`MvxApplicationDelegate` uses an overload of the AppDelegare.FinishedLaunching that is no longer supported. Applications which instead override the newer supported overload will cause the LifetimeChanged event to not fire.

`MvxApplicationDelegate` uses the correct, newer supported overload.

No.

Apple states:

> Important
> For app initialization, it is highly recommended that you use this method and the application:willFinishLaunchingWithOptions: method and do not use the applicationDidFinishLaunching: method, which is intended only for apps that run on older versions of iOS.

https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622921-application?language=objc

- [X] All projects build


nmilcoff:
👍 Great

PR #2035, opened 2017-07-13
Fix Forms implementation for MvvmCross

martijn00:
Makes Forms easy to use with MvvmCross.

Hacks and stuff not working

Easy to get started with a Forms app

No, apps keep working, but are recommended to update.

- [ ] All projects build

### DIFF
--- a/MvvmCross-Forms/MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs
+++ b/MvvmCross-Forms/MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs
@@ -20,32 +20,20 @@ namespace MvvmCross.Forms.Presenters
     {
         public const string ModalPresentationParameter = "modal";
 
-        private Application _mvxFormsApp;
-
-        public Application MvxFormsApp
+        private MvxFormsApplication _formsApplication;
+        public MvxFormsApplication FormsApplication
         {
-            get
-            {
-                return _mvxFormsApp;
-            }
-            set
-            {
-                if (value == null)
-                {
-                    throw new ArgumentException("MvxFormsApp cannot be null");
-                }
-
-                _mvxFormsApp = value;
-            }
+            get { return _formsApplication; }
+            set { _formsApplication = value; }
         }
 
         protected MvxFormsPagePresenter()
         {
         }
 
-        protected MvxFormsPagePresenter(Application mvxFormsApp)
+        protected MvxFormsPagePresenter(MvxFormsApplication formsApplication)
         {
-            MvxFormsApp = mvxFormsApp;
+            FormsApplication = formsApplication ?? throw new ArgumentNullException(nameof(formsApplication), "MvxFormsApp cannot be null");
         }
 
         public override void ChangePresentation(MvxPresentationHint hint)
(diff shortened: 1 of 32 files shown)


MvvmCross/iOS/iOS/Platform/MvxIosSetup.cs
```diff
@@ -42,10 +42,11 @@ protected MvxIosSetup(IMvxApplicationDelegate applicationDelegate, UIWindow wind
             _applicationDelegate = applicationDelegate;
         }
 
-        protected MvxIosSetup(IMvxApplicationDelegate applicationDelegate, IMvxIosViewPresenter presenter)
+        protected MvxIosSetup(IMvxApplicationDelegate applicationDelegate, UIWindow window, IMvxIosViewPresenter presenter) : this (applicationDelegate, window)
         {
-            _presenter = presenter;
+            _window = window;
             _applicationDelegate = applicationDelegate;
```
nmilcoff: `_applicationDelegate` and  `_window` are already set in the call for `: this (applicationDelegate, window)`, no need to assigning them again here

MvvmCross-Forms/MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs
```diff
@@ -20,32 +20,20 @@ public abstract class MvxFormsPagePresenter
     {
         public const string ModalPresentationParameter = "modal";
 
-        private Application _mvxFormsApp;
-
-        public Application MvxFormsApp
+        private MvxFormsApplication _formsApplication;
+        public MvxFormsApplication FormsApplication
```
nmilcoff: This rename is actually a breaking change, isn't it?

TestProjects/Playground/Playground.iOS/Setup.cs
```diff
@@ -14,8 +14,8 @@ public Setup(MvxApplicationDelegate applicationDelegate, UIWindow window)
         {
         }
 
-        public Setup(MvxApplicationDelegate applicationDelegate, IMvxIosViewPresenter presenter)
-            : base(applicationDelegate, presenter)
+        public Setup(MvxApplicationDelegate applicationDelegate, UIWindow window, IMvxIosViewPresenter presenter)
```
nmilcoff: If I'm not missing anything, then this is a big breaking change. The constructor 
`public Setup(MvxApplicationDelegate applicationDelegate, IMvxIosViewPresenter presenter)` doesn't exist anymore on `MvxIosSetup`. Do we need this change?

MvvmCross-Forms/MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs
martijn00: Yes. But it is necessary because otherwise we can't extend this with MvvmCross methods.

TestProjects/Playground/Playground.iOS/Setup.cs
martijn00: The problem with this was that in this case Window was never set, so it was never working. In any case, i've never used this ctor, but always the other one.

Issue #1972, opened 2017-06-20
Initialize not called when manually instantiating an MvxViewModel


mrtnkrstn:
To help us fix your issue, please provide the information in the below template. 

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.
>
> Please only file bugs verified on latest version of MvvmCross and _stable_ Xamarin tools

### Steps to reproduce

1. when manually instantiating an MvxViewModel, I use the following code so that methods of the VM life cycle (`Init()`, `Start()`) are called:

```
var mvxViewModelLoader = Mvx.Resolve<IMvxViewModelLoader>();
var vmRequest = MvxViewModelRequest.GetDefaultRequest(typeof(MyViewModel));
_myViewModel = mvxViewModelLoader.LoadViewModel(vmRequest, null) as MyViewModel;
```

This is typically done in a Parent-Child VM setup (e.g. Tab Bars), where the Parent VM has to instantiate the Children VMs.

### Expected behavior
Using the above method, `Initialize()` should be called in the Child VM so that the vm can be inited (e.g. populated via web services).

### Actual behavior
Using the above method, `Initialize()` is not called on the Child VM, (although `Start()` is).

### Configuration

**Version:** 5.0.3

**Platform:** 
- [x] iOS
- [x] Android
- [ ] WPF
- [ ] UWP

mrtnkrstn:
Also, `Initialize()` is not called in the VM that is navigated to from `RegisterAppStart()`. See test project at https://github.com/mrtnkrstn/MvvmCross_5.0.0

Cheesebaron:
RegisterAppStart uses ShowViewModel internally if using the default.

Cheesebaron:
IMvxViewModelLoader is not used by the new NavigationService, so you are mimicking ShowViewModel by using it.

IMvxViewModelLoader is kept as is to not break all apps using ShowViewModel.

mrtnkrstn:
@Cheesebaron How Should `RegisterAppStart()` be called in MvvmCross 5? What is the recommended way to "manually" instantiate VMs in MvvmCross 5?

Cheesebaron:
For now use a custom AppStart:

https://github.com/MvvmCross/MvvmCross/blob/develop/TestProjects/Navigation/RoutingExample.Core/AppStart.cs#L7

The manual instantiation really depends on what behavior you want

mrtnkrstn:
[...]

Issue #1947, opened 2017-06-15
Appearing() and other ViewModel LifeCycle methods not called


kodeo:
### Steps to reproduce

1. In Visual Studio 2015, do New | Project | MvvmCross Application

2. To FirstViewModel add `public override void Appearing()`

3. Set Droid project as Startup and launch

### Expected behavior
`Appearing()` is called

### Actual behavior
`Appearing()` isn't called

### Configuration

**Version:** 5.0

**Platform:** 
- [ ] iOS
- [X] Android
- [ ] WPF
- [ ] UWP

Bowman74:
@kodeo 
I noticed this too. What is happening here is that the first view model in an application is using the equivalent of the old CIRS lifecycle ShowViewModel. In order for Initialize to be called, one would need to push the view model onto the stack using the navigation service which isn't what any of the platform implementations I've tried are doing for startup.

martijn00:
The First ViewModel is still called with `ShowViewModel` to not break behaviour for all older apps. It is easy to fix this with some custom code.

In App.cs
```c#
// Construct custom application start object	
Mvx.ConstructAndRegisterSingleton<IMvxAppStart, AppStart>();
var appStart = Mvx.Resolve<IMvxAppStart>();

// register the appstart object
RegisterAppStart(appStart);
```

In AppStart.cs

```c#
public class AppStart : IMvxAppStart     
{         
    private readonly IMvxNavigationService _navigationService;  		
    public AppStart(IMvxNavigationService navigationService)         
    {             
        _navigationService = navigationService;         
    }          

    public void Start(object hint = null)         
    {            
        try             
        {                 
            _navigationService.Navigate<UserOverviewViewModel>();             
        }            
        catch (Exception e) 
        {      
        }
    }
} 
```

belhunter:
I noticed that such methods not called if view model is used in view pager fragment (only outdated lifecycle executed). So it makes difficulties in writing view models due to I need to check where it is used to write appropriate lifecycle methods.

tougher:
Anybody made a workaround for this?

I will work on a pull request for this issue soon, even though it's not a direct MvvmCross issue.

PR #2042, opened 2017-07-14
Add appstart using navigationservice

martijn00:
Feature

Able to navigate with the NavigationService in a save way on AppStart.

No

#1947
#1972

- [x] All projects build


nmilcoff:
Maybe we can make this AppStart the new default? Otherwise it doesn't really fix any issue, since users of MvvmCross 5 will expect `Initialize` and `Appearing` to be called always

willsb:
Maybe add a deprecation notice to the method with the old behavior so that backwards compatibility still exists. Though I see no reason not to use the new NavigationService, so dunno if making it default is a problem to anyone (for me it certainly isn't)

PR #2050, opened 2017-07-19
make WrapInNavigationController attribute working for tabBarControlle…

TingtingAn:
make WrapInNavigationController attribute of MvxRootPresentationAttribute  working for tabBarController/ splitController

[MvxRootPresentation(WrapInNavigationController = true)], the WrapInNavigationController attribute is not working for MvxTabBarViewController and MvxSplitViewController

make the WrapInNavigationController  attrobute working

No

None

- [x] All projects build

### DIFF
--- a/MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
+++ b/MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
@@ -111,69 +111,38 @@ namespace MvvmCross.iOS.Views.Presenters
             MvxRootPresentationAttribute attribute,
             MvxViewModelRequest request)
         {
-            // check if viewController is a TabBarController
-            if (viewController is IMvxTabBarViewController tabBarController)
-            {
-                TabBarViewController = tabBarController;
-
-                if (!attribute.WrapInNavigationController)
-                {
-                    SetWindowRootViewController(viewController);
-
-                    CloseMasterNavigationController();
-                }
-                else
-                {
-                    viewController = CreateNavigationController(viewController);
-                    MasterNavigationController = viewController as MvxNavigationController;
-                    SetWindowRootViewController(viewController);
-                }
-
-                CleanupModalViewControllers();
-                CloseSplitViewController();
-
-                return;
-            }
-            // check if viewController is a SplitViewController
-            else if (viewController is IMvxSplitViewController splitController)
-            {
-                SplitViewController = splitController;
-
-                if (!attribute.WrapInNavigationController)
-                {
-                    SetWindowRootViewController(viewController);
-
-                    CloseMasterNavigationController();
-                }
-                else
-                {
-                    viewController = CreateNavigationController(viewController);
-                    MasterNavigationController = viewController as MvxNavigationController;
-                    SetWindowRootViewController(viewController);
-                }
-
-                CleanupModalViewControllers();
-                CloseTabBarViewController();
-
-                return;
-            }
-
             // check if viewController is trying to initialize a navigation stack
             if (attribute.WrapInNavigationController)
             {
                 viewController = CreateNavigationController(viewController);
                 MasterNavigationController = viewController as MvxNavigationController;
                 SetWindowRootViewController(viewController);
+            }
+            else
+            // display the plain viewController as root
+            {
+                SetWindowRootViewController(viewController);
 
-                CleanupModalViewControllers();
-                CloseTabBarViewController();
-                CloseSplitViewController();
-
-                return;
+                CloseMasterNavigationController();
             }
 
-            // last scenario: display the plain viewController as root
-            SetWindowRootViewController(viewController);
+            // check if viewController is a TabBarController
+            if (viewController is IMvxTabBarViewController tabBarController)
+            {
+                TabBarViewController = tabBarController;
+
+                CloseTabBarViewController();
+                CleanupModalViewControllers();
+            }
+            // check if viewController is a SplitViewController
+            else if (viewController is IMvxSplitViewController splitController)
+            {
+                SplitViewController = splitController;
+
+                CloseSplitViewController();
+                CleanupModalViewControllers();
+            }
+
         }
 
         protected virtual void ShowChildViewController(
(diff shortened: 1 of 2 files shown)


MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
```diff
@@ -111,48 +111,43 @@ public virtual void Show(IMvxIosView view, MvxViewModelRequest request)
...
+            }
+
             // check if viewController is a TabBarController
             if (viewController is IMvxTabBarViewController tabBarController)
             {
                 TabBarViewController = tabBarController;
-                SetWindowRootViewController(viewController);
 
-                CloseMasterNavigationController();
+                CloseTabBarViewController();
```
mvanbeusekom: This doesn't make any sense... First you assign the `TabBarViewController` and then you call `CloseTabBarViewController` which will destroy it

MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
```diff
@@ -111,48 +111,43 @@ public virtual void Show(IMvxIosView view, MvxViewModelRequest request)
...
-
             // check if viewController is a SplitViewController
-            if (viewController is IMvxSplitViewController splitController)
+            else if (viewController is IMvxSplitViewController splitController)
             {
                 SplitViewController = splitController;
-                SetWindowRootViewController(viewController);
 
-                CloseMasterNavigationController();
+                CloseSplitViewController();
```
mvanbeusekom: Same as with the `TabBarViewController`, why first assign the `SplitViewController ` and after that directly destroy it by calling `CloseSplitViewController`

MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
```diff
@@ -111,48 +111,43 @@ public virtual void Show(IMvxIosView view, MvxViewModelRequest request)
             MvxRootPresentationAttribute attribute,
             MvxViewModelRequest request)
         {
+            // check if viewController is trying to initialize a navigation stack
+            if (attribute.WrapInNavigationController)
+            {
+                var navigationController = CreateNavigationController(viewController);
+                MasterNavigationController = navigationController as MvxNavigationController;
```
mvanbeusekom: Shouldn't we call `CloseMasterNavigationController` before assigning a new value to the `MasterNavigationController` property? If this is the case we should make the property readonly and add a set method which ensure everything is called in the correct order.

@nmilcoff could you maybe double check if my suggestion makes sense?

TingtingAn: @mvanbeusekom  mmm, l am not clear this logic, but if you were right, the `TabBarViewController = tabBarController;` also need call `CloseTabBarViewController();` and  the `SplitViewController ` is same.  @nmilcoff  please give some suggstion.

nmilcoff: We're clearing the presenter's reference to the old NavigationController by re-assigning it. That should be enough here... It will be disposed by the OS when appropiate

MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
```diff
@@ -111,48 +111,43 @@ public virtual void Show(IMvxIosView view, MvxViewModelRequest request)
...
-                viewController = CreateNavigationController(viewController);
-                MasterNavigationController = viewController as MvxNavigationController;
-                SetWindowRootViewController(viewController);
-
-                CleanupModalViewControllers();
                 CloseTabBarViewController();
                 CloseSplitViewController();
-
-                return;
+                CleanupModalViewControllers();
```
nmilcoff: `CleanupModalViewControllers()` is called in all conditions, please call it just once

MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
```diff
@@ -177,15 +171,15 @@ public virtual void Show(IMvxIosView view, MvxViewModelRequest request)
...
-            if (TabBarViewController != null && TabBarViewController.ShowChildView(viewController))
+            if (MasterNavigationController != null)
             {
+                PushViewControllerIntoStack(MasterNavigationController, viewController, attribute.Animated);
+
                 return;
             }
 
-            if (MasterNavigationController != null)
+            if (TabBarViewController != null && TabBarViewController.ShowChildView(viewController))
```
nmilcoff: Mhmmm I'm not sure about this change... we need to add a way to decide whether to prioritize a TabBarController child or a MasterNavigationController child. In any case, we should try to avoid a breaking change here, doesn't seem necessary

nmilcoff:
Hey @TingtingAn, we appreciate your PR and also that you're making good progress with all the changes we're requesting. Just a remark, please use more descriptive messages for your commits. We'll need also a squash when you're done

MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
TingtingAn: @nmilcoff   mmm,  actually, i think we should remove the code `if (TabBarViewController != null && TabBarViewController.ShowChildView(viewController))
            {
                return;
            }` in `ShowChildViewController` method.  Because if user set `MvxTabPresentation` attribute for a viewcontroller, It will just call the `ShowTabViewController` method.

TingtingAn:
[https://gyazo.com/95083c7134073c9d7191049448d99630](https://gyazo.com/95083c7134073c9d7191049448d99630 )

[https://gyazo.com/a230a2d3e2b298f6df9e803a26a3b1b3](https://gyazo.com/a230a2d3e2b298f6df9e803a26a3b1b3)

please see the differences for the two `Show child` ways. The second one is usual, but current we are showing the first one.

And aslo I know maybe this is a breaking change. But i have no enought ability/ time to complete it.
Could you mark it and fix it later? Thanks!

nmilcoff:
@TingtingAn those videos show two very different behaviors: 
- The first one shows a MasterNavigationController, where TabBarController is a child
- The second one shows a NavigationController inside a Tab of a TabBarController. It is a __very common__ situation. Instagram, Facebook and many other apps use this pattern.

I've made a new PR containing your changes, squashing commits and updating the related docs :). It's a different PR because I've tried but couldn't push my changes to your branch.

PR #2058, opened 2017-07-20
Add support for one sidemenu at a time

martijn00:
Feature

If you use a left and right sidemenu both will always be shown.

With the new enums you can shown either a menu on the right or just one on the left.

No

- [x] All projects build

### DIFF
--- a/MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresenter.cs
+++ b/MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresenter.cs
@@ -118,11 +118,11 @@ namespace MvvmCross.iOS.Support.XamarinSidebar
                     break;
                 case MvxPanelEnum.CenterWithLeft:
                     navigationController.PushViewController(viewController, true);
-                    viewController.ShowMenuButton(SideBarViewController as MvxSidebarViewController, true, false);
+                    viewController.ShowMenuButton(SideBarViewController as MvxSidebarViewController, showLeft: true, showRight: false);
                     break;
                 case MvxPanelEnum.CenterWithRight:
                     navigationController.PushViewController(viewController, true);
-                    viewController.ShowMenuButton(SideBarViewController as MvxSidebarViewController, false, true);
+                    viewController.ShowMenuButton(SideBarViewController as MvxSidebarViewController, showLeft: false, showRight: true);
                     break;
             }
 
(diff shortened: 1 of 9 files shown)


mvanbeusekom:
The following is an improvement suggestion for a future release (since it will introduce a breaking change).

Change the `ShowPanel` property of the `MvxSidebarPresentationAttribute` to an flag-enum which can be used to indicate if no, only left, only right or both menus should be shown. The enum could look something like this:
```
[Flags]
public enum ShowPanel 
{
    None = 0,
    Left = 1,
    Right = 2
}
```

This enum will replace the `MvxPanelEnum.CenterAndLeft` and `MvxPanelEnum.CenterAndRight` options, simplifying the logic in the `MvxSidebarPresenter.ShowPanelAndResetToRoot` method (last two cases of the switch statement can be removed).

Another benefit would be that we can more easily offer additional options (e.g. center menu button).

MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresenter.cs
```diff
@@ -90,14 +104,25 @@ protected virtual bool ShowPanel(MvxPanelEnum panel, UIViewController viewContro
...
                 case MvxPanelEnum.Right:
                     break;
                 case MvxPanelEnum.Center:
                 default:
-                    navigationController?.PushViewController(viewController, true);
+                    navigationController.PushViewController(viewController, true);
+                    break;
+                case MvxPanelEnum.CenterWithLeft:
+                    navigationController.PushViewController(viewController, true);
+                    viewController.ShowMenuButton(SideBarViewController as MvxSidebarViewController, true, false);
```
nmilcoff: Could you use parameter names here for the optional ones? It makes the code more defensive and easier to read. I mean:
`viewController.ShowMenuButton(SideBarViewController as MvxSidebarViewController, showLeft: true, showRight: false);`

MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresenter.cs
```diff
@@ -90,14 +104,25 @@ protected virtual bool ShowPanel(MvxPanelEnum panel, UIViewController viewContro
...
+                    break;
+                case MvxPanelEnum.CenterWithRight:
+                    navigationController.PushViewController(viewController, true);
+                    viewController.ShowMenuButton(SideBarViewController as MvxSidebarViewController, false, true);
```
nmilcoff: Same as previous comment about parameter names

martijn00:
@nmilcoff If you include a menu on both left and right it will currently always show both. This PR enables to just have a menu on the right side, while still having back button on the left.

nmilcoff:
@martijn00 👍 Makes sense now, thanks for the explanation!

Issue #2071, opened 2017-07-26
New navigation service doesn't seem to be truly async as opposite to old ShowViewModel() navigation


indexless:
### Steps to reproduce

1. Download Playground sample from https://github.com/MvvmCross/MvvmCross/tree/develop/TestProjects/Playground

2. In ./Playground.Core\ViewModels\Tab1ViewModel.cs, override initializer and add a couple of dummy other overridings (just to be able to set breakpoints and follow the call sequence):

        public override async Task Initialize()
        {
            await Task.Delay(2000);
        }

        public override void ViewAppearing()
        {
            base.ViewAppearing();
        }

        public override void ViewAppeared()
        {
            base.ViewAppeared();
        }

### Expected behavior
I'm expecting the view to appear immediately (as it does for ShowViewModel()), then ViewAppearing() and ViewAppeared() should be called, and after that the Initialize() should complete when Task.Delay(2000) is done. Possibly this issue is pretty close to #2046 

### Actual behavior
The app is freezing and waiting for Task.Delay(2000)'s completion, and after that ViewAppearing() and ViewAppeared() are called.

### Configuration

**Version:** 5.1.0, tested sample is from 491b012a513ad42522c24e22a4ba2e6f0ef496d2 with updated Playground.

**Platform:** 
- [x] iOS
- [ ] Android
- [ ] WPF
- [ ] UWP

martijn00:
`ShowViewModel` was not async, but Fire and forget because it was called using reflection. We do experience the same problem you describe here so i think we should look into this.

indexless:
Ah okay, got it, thanks.

martijn00:
We have 2 option right now i think. We can wrap the `Initialize` in a Task.Run so it will not stop the showing of the view, or we do the `Initialize` after the showing of the view, but then the parameter might not be available to the view events.

Another option would be to create 2 methods. One to pass the value, this can just be a normal `void` which is set before showing the view, and then run the `Initialize` Task after the showing. this would be a breaking change.

indexless:
Martijn, thanks for your reply.

[...]

Issue #2046, opened 2017-07-17
Navigation to tab bar models seems to be broken for MvvmCross/iOS 5.0.4 onwards


indexless:
To help us fix your issue, please provide the information in the below template. 

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.
>
> Please only file bugs verified on latest version of MvvmCross and _stable_ Xamarin tools

### Steps to reproduce

1. Download Playground sample from https://github.com/MvvmCross/MvvmCross/tree/develop/TestProjects/Playground

2. In ./Playground.Core/ViewModels/TabsRootViewModel.cs, replace command's value with

            new MvxAsyncCommand(() =>
            {
                var nav = Mvx.Resolve<IMvxNavigationService>();
                return Task.WhenAll(nav.Navigate<Tab1ViewModel>(),
                                    nav.Navigate<Tab2ViewModel>(),
                                    nav.Navigate<Tab3ViewModel>());
            });

(The command property's type should also become IMvxAsyncCommand.)

3. In  ./Playground.iOS/Views/TabsRootView.cs, replace

                ViewModel.ShowInitialViewModelsCommand.Execute(null);

with

                ViewModel.ShowInitialViewModelsCommand.ExecuteAsync(null);

4. For each tab view model ./Playground.Core/ViewModels/Tabs#ViewModel.cs declare Initialize() as

        public override Task Initialize() => Task.Delay(9000);
        public override Task Initialize() => Task.Delay(6000);
        public override Task Initialize() => Task.Delay(3000);

for tab view model 1, 2, 3, respectively. Build, run, and select Tabs navigation.

### Expected behavior
All three tabs should appear instantly as they actually do when navigation is performed via deprecated ShowViewModel(). Also that was working for Mvvm from 5.0.0 to 5.0.3, but stopped in 5.0.4 onwards.

### Actual behavior
There is a black screen with no tabs for first three seconds, then third, second, and first (in that order) tabs appear with 3-second intervals.

### Configuration

**Version:** 5.0.4, 5.0.5, 5.0.6

**Platform:** 
- [x] iOS
- [ ] Android
- [ ] WPF
- [ ] UWP

nmilcoff:
Playground project is now updated and uses the new MvxNavigationService. Tabs do work as expected. Maybe the Initialize method with Task.Delays is what makes the issue appear?

indexless:
Nice, thank you for update!

[...]

Issue #2009, opened 2017-07-03
iOS - public override async Task Initialize() throws exception on launch


benhysell:
To help us fix your issue, please provide the information in the below template. 

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.
> https://github.com/benhysell/XamarinSidebarSample
> Please only file bugs verified on latest version of MvvmCross and _stable_ Xamarin tools

### Steps to reproduce

1. Launch App

2. Exception thrown - Foundation.MonoTouchException: Objective-C exception thrown.  Name: NSInternalInconsistencyException Reason: Application windows are expected to have a root view controller at the end of application launch

3.

### Expected behavior
Tell us what should happen - Application should launch

### Actual behavior
Tell us what happens instead - Exception Thrown

### Configuration

**Version:** 5.0.5

**Platform:** 
- [ X] iOS
- [ ] Android
- [ ] WPF
- [ ] UWP

Modified the XamarinSidebarSample sample to show error I was seeing in my own application.  Changed AppStart.cs to use IMvxNavigationService to launch CenterPanelViewModel.  In CenterPanelViewModel I added
```
public override async Task Initialize()
{
            //simulate database lookup
            await Task.Delay(5000);
}
```

Exception is thrown:
Foundation.MonoTouchException: Objective-C exception thrown.  Name: NSInternalInconsistencyException Reason: Application windows are expected to have a root view controller at the end of application launch

This is new behavior...I just upgraded from 5.0.3 and started seeing this.

Cheesebaron:
Isn't this expected? iOS doesn't allow you to spend too much time before a root VC needs to be shown or app is nuked.

martijn00:
The problem is in Appstart: https://github.com/benhysell/XamarinSidebarSample/blob/master/MvvmCross.iOS.Support.Core/AppStart.cs#L23

You don't block navigation here, so it will continue before the actual screen was shown.

benhysell:
Tried @martijn00's suggestion...worked as expected.

At first blush the solution, 'didn't feel right'...I had pictures of Stephen Cleary yelling at me for committing  an async/await sin, [https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html](https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html).

[...]

Issue #1851, opened 2017-05-23
Call Close immediately after receiving the return value by NavigationService, view(activity) cannot close itself.


jz5:
### Steps to reproduce

```c#
public class FooViewModel : MvxViewModelResult<string>
{
    private readonly IMvxNavigationService NavigationService;
    public FooViewModel(IMvxNavigationService navigation)
    {
        NavigationService = navigation;
    }
    public MvxAsyncCommand BarCommand => new MvxAsyncCommand(async () =>
    {
        var result = await NavigationService.Navigate<BarViewModel, string>();
        await Close(result); // iOS: can close this, Android: cannot close this
    });
}
```

### Expected behavior
Android also can close 
I want to close both views.

### Actual behavior
View of BarViewModel is closed.
View of FooViewModel isn't closed.

Activity is finished in MvxAndroidViewPresenter's Close method.
MvxAndroidViewPresenter.Activitiy(```Mvx.Resolve<IMvxAndroidCurrentTopActivity>().Activity```) still refers to activity of BarViewModel when Close is called.
I have no solutions. Are there ideas or workarounds?

### Configuration

**Version:** 5.0

**Platform:** Android

MrXemiu:
I've had a similar problem.  On Android with 'public class MainActivity : MvxCachingFragmentCompatActivity<MainViewModel>, INavigationActivity' as a hosting Activity and navigating between Fragments, what I've seen is that, using your example, `FooViewModel.Close(result)' navigates **back** to `BarViewModel`.

I, too, am searching for a solution which would allow me to chain `Close(result)` to pass data backwards.

martijn00:
When using commands with the new navigation MvxAsyncCommand should be used.

jz5:
It still doesn't work using MvxAsyncCommand on v5.0.1.
Minimum project: https://github.com/jz5/MvxApp1851

nmilcoff:
The reproduction project doesn't use a custom presenter, therefore [this](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Droid/Droid/Views/MvxAndroidViewPresenter.cs) is the one being used. It looks like an issue with `IMvxAndroidCurrentTopActivity`, may be related to #783.

rrispoli:
Hi, I'm using MvvmCross 5.2.1 and this problem is occurring with me. I have a ViewModelA that go to ViewModelB and then ViewModelC.

[...]

PR #2072, opened 2017-07-26
Move result close to the navigation service

martijn00:
Because it can only be used with that.

This fixes unwanted behaviour

Close with result will just close if the NavigationService was used. But since it is in the base class you could use it anywhere.

You need to use it through the NavigationService. This also makes it easier to implement the interface for it.

Yes, but it would not have worked before if you used ShowViewModel.

Fixes https://github.com/MvvmCross/MvvmCross/issues/2071
Fixes https://github.com/MvvmCross/MvvmCross/issues/2046
Fixes https://github.com/MvvmCross/MvvmCross/issues/2009

- [x] All projects build

### DIFF
--- a/MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
+++ b/MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
@@ -169,9 +169,9 @@ namespace MvvmCross.Core.Navigation
             var args = new NavigateEventArgs(viewModel);
             OnBeforeNavigate(this, args);
 
+            ViewDispatcher.ShowViewModel(request);
             await viewModel.Initialize().ConfigureAwait(false);
 
-            ViewDispatcher.ShowViewModel(request);
             OnAfterNavigate(this, args);
         }
 
@@ -180,10 +180,10 @@ namespace MvvmCross.Core.Navigation
             var args = new NavigateEventArgs(viewModel);
             OnBeforeNavigate(this, args);
 
-            await viewModel.Initialize(param).ConfigureAwait(false);
+            viewModel.Declare(param);
+            ViewDispatcher.ShowViewModel(request);
             await viewModel.Initialize().ConfigureAwait(false);
 
-            ViewDispatcher.ShowViewModel(request);
             OnAfterNavigate(this, args);
         }
 
@@ -200,12 +200,12 @@ namespace MvvmCross.Core.Navigation
             }
 
             var tcs = new TaskCompletionSource<object>();
-            viewModel.SetClose(tcs);
+            viewModel.CloseCompletionSource = tcs;
             _tcsResults.Add(viewModel, tcs);
 
+            ViewDispatcher.ShowViewModel(request);
             await viewModel.Initialize().ConfigureAwait(false);
 
-            ViewDispatcher.ShowViewModel(request);
             OnAfterNavigate(this, args);
 
             try
@@ -231,13 +231,14 @@ namespace MvvmCross.Core.Navigation
             }
 
             var tcs = new TaskCompletionSource<object>();
-            viewModel.SetClose(tcs);
+            viewModel.CloseCompletionSource = tcs;
             _tcsResults.Add(viewModel, tcs);
 
-            await viewModel.Initialize(param).ConfigureAwait(false);
+            viewModel.Declare(param);
+            ViewDispatcher.ShowViewModel(request);
             await viewModel.Initialize().ConfigureAwait(false);
 
-            ViewDispatcher.ShowViewModel(request);
+
             OnAfterNavigate(this, args);
 
             try
@@ -371,7 +372,7 @@ namespace MvvmCross.Core.Navigation
             _tcsResults.TryGetValue(viewModel, out TaskCompletionSource<object> _tcs);
 
             //Disable cancelation of the Task when closing ViewModel through the service
-            viewModel.SetClose(null);
+            viewModel.CloseCompletionSource = null;
 
             try
             {
@@ -379,7 +380,7 @@ namespace MvvmCross.Core.Navigation
                 if (closeResult)
                     _tcs?.TrySetResult(result);
                 else
-                    viewModel.SetClose(_tcs);
+                    viewModel.CloseCompletionSource = _tcs;
                 return closeResult;
             }
             catch (Exception ex)
--- a/MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
+++ b/MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
@@ -378,7 +378,10 @@ namespace MvvmCross.Core.Navigation
             {
                 var closeResult = await Close(viewModel);
                 if (closeResult)
+                {
                     _tcs?.TrySetResult(result);
+                    _tcsResults.Remove(viewModel);
+                }
                 else
                     viewModel.CloseCompletionSource = _tcs;
                 return closeResult;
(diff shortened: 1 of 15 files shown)


martijn00:
It also seems that https://github.com/MvvmCross/MvvmCross/issues/1851 is fixed with this PR.

MvvmCross/Core/Core/ViewModels/IMvxViewModel.cs
```diff
@@ -37,14 +37,13 @@ public interface IMvxViewModel
 
     public interface IMvxViewModel<TParameter> : IMvxViewModel where TParameter : class
     {
-        Task Initialize(TParameter parameter);
+        void Declare(TParameter parameter);
```
nmilcoff: @martijn00 so this PR introduces a new ViewModel lifecyle method? It isn't in the PR description/any new docs 🤔

MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
```diff
@@ -164,74 +164,115 @@ public virtual Task<bool> CanNavigate(string path)
...
+
+            OnAfterNavigate(this, args);
+        }
+
+        protected virtual async Task Navigate<TParameter>(MvxViewModelRequest request, IMvxViewModel<TParameter> viewModel, TParameter param, IMvxBundle presentationBundle = null, CancellationToken cancellationToken = default(CancellationToken)) where TParameter : class
+        {
+            var args = new NavigateEventArgs(viewModel);
+            OnBeforeNavigate(this, args);
+
+            viewModel.Declare(param);
```
nmilcoff: So `Declare` is run before the navigation itself is done, right?

MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
```diff
@@ -164,74 +164,115 @@ public virtual Task<bool> CanNavigate(string path)
...
+                });
             }
-            else if (viewModel is IMvxViewModelResult<TResult> vmResult)
-                vmResult.SetClose(tcs, cancellationToken);
-            else if (viewModel is IMvxViewModel<TParameter> vmParam)
-                await vmParam.Initialize(param);
-              
-            await viewModel.Initialize();
+
+            var tcs = new TaskCompletionSource<object>();
```
nmilcoff: `Declare` is not called here. Is it missing?

MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
```diff
@@ -25,10 +26,9 @@ public IMvxViewDispatcher ViewDispatcher
         }
 
         protected static readonly Dictionary<Regex, Type> Routes = new Dictionary<Regex, Type>();
-
         protected virtual IMvxNavigationCache NavigationCache { get; private set; }
-
-        public virtual IMvxViewModelLoader ViewModelLoader { get; set; }
+        protected IMvxViewModelLoader ViewModelLoader { get; set; }
+        protected ConditionalWeakTable<IMvxViewModel, TaskCompletionSource<object>> _tcsResults = new ConditionalWeakTable<IMvxViewModel, TaskCompletionSource<object>>();
```
nmilcoff: You only use `.Add` and `.TryGetValue` on `_tcsResults`. Does it mean you never remove items from it? Is that right? (never used ConditionalWeakTable before)

MvvmCross/Core/Core/ViewModels/IMvxViewModel.cs
```diff
@@ -37,14 +37,13 @@ public interface IMvxViewModel
 
     public interface IMvxViewModel<TParameter> : IMvxViewModel where TParameter : class
```
nmilcoff: Shouldn't `Declare` be defined in `IMvxViewModel` as well?

martijn00: No, `Declare` only used when using a parameter.

MvvmCross/Core/Core/ViewModels/IMvxViewModel.cs
martijn00: Not really, it is just used to set parameters before the view is shown.

MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
martijn00: That is right. This is because there is no `TParameter` used here.

MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
martijn00: This is because there is no `TParameter` used here.

MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
martijn00: Fixed

nmilcoff:
__Opinionated Comment__: May I suggest a different name for `Declare`? Maybe `Prepare` or `Instantiate` makes more sense in the context where it will be used

MvvmCross/Core/Core/Navigation/MvxNavigationService.cs
softlion: Should Initialize be really awaited ?

MvvmCross/Core/Core/ViewModels/IMvxViewModel.cs
softlion: I would have called it "Create"

softlion: Or maybe Created as it is a kind of internal lifecycle.

MvvmCross/Core/Core/ViewModels/MvxNavigationServiceAppStart.cs
```diff
@@ -18,9 +18,9 @@ public class MvxNavigationServiceAppStart<TViewModel>
     {
         protected readonly IMvxNavigationService NavigationService;
 
-        public MvxNavigationServiceAppStart()
+        public MvxNavigationServiceAppStart(IMvxNavigationService navigationService)
         {
-            NavigationService = Mvx.Resolve<IMvxNavigationService>();
+            NavigationService = navigationService;
```
softlion: Mvx.Resolve should be banned :)

softlion:
> ! As this PR introduces a new ViewModel lifecycle (Declare) and breaking changes to Initialize, maybe it makes sense to release as 5.2?

mvvmcross 5 has introduced a kind of breaking change from 4.x: Initialize is replacing Init(), but Init was called after the view was attached to the viewmodel (or seems to), and Initialize is currently called before creating the view. So moving it after the view is attached and adding a new customizable method before it is attached is better.

indexless:
>> So moving it after the view is attached and adding a new customizable method before it is attached is better.

Why don't use it just for parameter's passing (before all view events), and place the awaitable initialization code right into `ViewCreated()` (making view event callbacks aware of passed parameters)?

martijn00:
`ViewCreated` is not async.

indexless:
But it's possible to prepend callbacks by `async` keyword:

    public override async void ViewCreated() => await Task.Delay(3000);
    public override async void ViewAppearing() => await Task.Delay(2000);
    public override async void ViewAppeared() => await Task.Delay(1000);

And they will complete in the reverse order (I checked that). Isn't this desired behavior?

martijn00:
async void is very bad  practice: http://theburningmonk.com/2012/10/c-beware-of-async-void-in-your-code/

PR #2022, opened 2017-07-06
[WIP] New default Android presenter

martijn00:
## :sparkles: What kind of change does this PR introduce? (Bug fix, feature, docs update...)
It introduces a new Android presenter that is able to handle Fragments in any sort of Activity, support for Dialogs, and optional attributes for Activities.

## :arrow_heading_down: What is the current behavior?
No support for any of that.

## :new: What is the new behavior (if this is a feature change)?
Support for the new features:

- [x] Attributes for: Activtiy, Fragment, Tab, Dialog
- [x] Showing of those views even without Attribute
- [x] CustomAnimations per Attribute
- [x] SharedElement transitions
- [x] No dependency on MvxCachingFragmentActivity. All Activities can show Fragments now
- [x] Support for Fragments in default non AppCompat presenter
- [x] Closing of Dialog, Fragments, and Fragments when Activity closes
- [x] Override behaviour on runtime with `IMvxOverridePresentationAttribute`
- [x] Navigation to Tabs in ViewPager
- [x] Nested Fragments
- [ ] Caching of fragments from presenter
- [ ] Hook into Fragment lifecycle manager

## :boom: Does this PR introduce a breaking change?
Yes, but for apps most things should be easy to change.

## :bug: Recommendations for testing
Fragments and navigation between them.

## :memo: Links to relevant issues/docs
https://github.com/MvvmCross/MvvmCross/issues/1934

## :thinking: Checklist before submitting

- [x] All projects build
- [x] Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))
- [ ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contribute/mvvmcross-docs-style-guide))
- [x] Nuspec files were updated (when applicable)
- [x] Rebased onto current develop

PR #2083, opened 2017-07-29
Add ability to set fragment content by name rather than id

andyci:
Feature

To use the MvxFragment attribute, you have to supply a Resource ID.  Due to the way Android UI shared projects work, you cannot use resource IDs in an attribute because they are defined as `static int` rather than `const int`.

See https://bugzilla.xamarin.com/show_bug.cgi?id=39185

This pull request adds the ability to define the content ID using the resource name, i.e.
`[MvxFragment(typeof(ViewModel), "content_id")]`

No.

- [x] All projects build


andyci:
If you think this use case is quite rare then I would also be happy with just making FragmentContentId virtual so I can override it in my own class.

martijn00:
I think this is valuable to have, but also a bit conflicting with https://github.com/MvvmCross/MvvmCross/pull/2022 It would be great to work together to get that PR finished with the additional requirements like this one.

andyci:
I'll close this pull request then and help with #2022.

Issue #2084, opened 2017-07-29
MvxTabBarViewController not working anymore without WrapInNavigationController parameter


DelcoigneYves:
### Steps to reproduce

1. Update Star Wars sample to latest version of MvvmCross (5.1.1)

2. Run on iOS.

### Expected behavior
App shows pages correctly.

### Actual behavior
Crash: MvvmCross.Platform.Exceptions.MvxException: Trying to show a tab without a TabBarViewController, this is not possible!

When adding the the WrapInNavigationController parameter set to true, the app runs (but then we get 2 navigation bars in the sample).

### Configuration

**Version:** 5.1.1

**Platform:** 
- [ x ] iOS
- [ ] Android
- [ ] WPF
- [ ] UWP

nmilcoff:
Thanks for reporting this issue! It is fixed by #2085

PR #2085, opened 2017-07-29
MvxIosViewPresenter: Fix ShowRootViewController

nmilcoff:
Bug fix & improvements

#2076 introduced a refactor of ShowRootViewController method that broke the logic in some scenarios. Reason for this is that some properties of the ViewPresenter were being assigned too late (like TabBarViewController property, which caused #2084).

The logic was reverted without loosing the feature introduced in #2076. Also improved some other parts:
- MvxTabBarController.CloseChildViewModel now closes Tabs that are NavigationControllers (previously only worked for Tabs that were plain ViewControllers)
- `MvxTabBarController.ShowChildView` now uses `as` operator instead of a direct cast (app could potentially crash because of that)
- Also improved doc for MvxIosViewPresenter

No.

Run Playground.iOS

Related PR: #2076.
Related issue: #2084 

- [X] All projects build

### DIFF
--- a/MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
+++ b/MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
@@ -111,46 +111,56 @@ namespace MvvmCross.iOS.Views.Presenters
             MvxRootPresentationAttribute attribute,
             MvxViewModelRequest request)
         {
-            // first set the new RootViewController
-            if (attribute.WrapInNavigationController)
-            { 
-				// viewController is initializing a navigation stack
-
-                var navigationController = CreateNavigationController(viewController);
-                MasterNavigationController = navigationController as MvxNavigationController;
-                SetWindowRootViewController(navigationController);
-            }
-            else
-            {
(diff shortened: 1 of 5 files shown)


MvvmCross/iOS/iOS/Views/Presenters/MvxIosViewPresenter.cs
```diff
@@ -424,7 +445,7 @@ protected virtual void SetWindowRootViewController(UIViewController controller)
             foreach (var v in _window.Subviews)
                 v.RemoveFromSuperview();
 
-            _window.AddSubview(controller.View);
+            //_window.AddSubview(controller.View);
             _window.RootViewController = controller;
```
mvanbeusekom: I think it is good practise not to leave commented out code in the code base. Just remove it, or if there is a good reason to leave it mention it with a comment in the code so other people also know why it is commented out.

nmilcoff: Definitely right @mvanbeusekom! Good spot, forgot to remove it before commiting

PR #2088, opened 2017-08-01
Allowing MvxNavigationFacades to set parameters

b099l3:
feature

When changing the parameters in a navigation facade they are ignored.

This PR allows for parameter to be set in the navigation facade when deeplinking.

No

Open the Navigation Test Project and look at how the PrePopulateRoutingFacade adds the parameter to TestCViewModel

https://stackoverflow.com/questions/45436119/sending-new-parameters-in-mvxviewmodelrequest-from-a-imvxnavigationfacade-when-d

- [X] All projects build

### DIFF
--- a/TestProjects/Navigation/RoutingExample.Core/ViewModels/TestCViewModel.cs
+++ b/TestProjects/Navigation/RoutingExample.Core/ViewModels/TestCViewModel.cs
@@ -1,0 +1,41 @@
+﻿using System.Threading.Tasks;
+using MvvmCross.Core.Navigation;
+using MvvmCross.Core.ViewModels;
+using MvvmCross.Platform;
+using RoutingExample.Core.ViewModels;
+
+namespace RoutingExample.Core.ViewModels
+{
+    public class TestCViewModel
+        : MvxViewModel<User, User>
+    {
+        private readonly IMvxNavigationService _navigationService;
+        public TestCViewModel(IMvxNavigationService navigationService)
+        {
+            _navigationService = navigationService;
+        }
+
+        private string _parameter;
+        public string Parameter
+        {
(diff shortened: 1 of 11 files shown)


TestProjects/Navigation/RoutingExample.Core/ViewModels/TestCViewModel.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        private string _parameter;
+        public string Parameter
+        {
+            get { return _parameter; }
+            set { SetProperty(ref _parameter, value); }
+        }
+
+        public void Init(string viewmodelcparameter)
```
martijn00: The only problem with this is that it is using the old `Init` with reflection. Can we improve this?

b099l3: So do you mean to I should try to send the parameters through to the Prepare method?

I was just following [this](https://www.mvvmcross.com/documentation/fundamentals/navigation#uri-navigation) but I can try to change it to use the Prepare method.

martijn00: Prepare is implemented because the interface is there. It would be great to have a way to make this typed as well. So maybe and a new interface which will receive the parameters?

b099l3: I will take a look into creating an interface that can have typed parameters too and create another PR

martijn00: Yeah i was thinking too that could be done in the next PR.

PR #2106, opened 2017-08-08
Allow for non-reference types to be passed as parameter

mvanbeusekom:
Feature

Currently it is not possible to pass a value type as a parameter when navigating with the `MvxNavigationService`.

After applying this PR, the constraint `where TParameter : class` has been removed. Which will allow value types to be passed in as a parameter when navigating using the `MvxNavigationService`.

Yes, the signatures of some of the methods in the `IMvxViewModel` and `IMvxNavigationService` interfaces have been updated (the constraint `where TParameter : class` has been removed) which will cause a breaking change in case people used custom implementations for these interfaces.

The iOS test project `Navigation/RoutingExample.iOS` contains example usage of passing in a value type as parameter. Start the application and click the "mvx://test/c" button.

- [x] All projects build


MvvmCross/Core/Core/Navigation/IMvxNavigationService.cs
```diff
@@ -21,24 +21,24 @@ public interface IMvxNavigationService
...
 
         Task Navigate(IMvxViewModel viewModel, IMvxBundle presentationBundle = null);
-        Task Navigate<TParameter>(IMvxViewModel<TParameter> viewModel, TParameter param, IMvxBundle presentationBundle = null) where TParameter : class;
+        Task Navigate<TParameter>(IMvxViewModel<TParameter> viewModel, TParameter param, IMvxBundle presentationBundle = null);
         Task<TResult> Navigate<TResult>(IMvxViewModelResult<TResult> viewModel, IMvxBundle presentationBundle = null, CancellationToken cancellationToken = default(CancellationToken)) where TResult : class;
-        Task<TResult> Navigate<TParameter, TResult>(IMvxViewModel<TParameter, TResult> viewModel, TParameter param, IMvxBundle presentationBundle = null, CancellationToken cancellationToken = default(CancellationToken)) where TParameter : class where TResult : class;
+        Task<TResult> Navigate<TParameter, TResult>(IMvxViewModel<TParameter, TResult> viewModel, TParameter param, IMvxBundle presentationBundle = null, CancellationToken cancellationToken = default(CancellationToken)) where TResult : class;
```
martijn00: Can you do the same for Result?

PR #2121, opened 2017-08-17
Now Forms MasterDetail can launch on iOS

mellson:
Now the Forms MasterDetail example can launch on iOS.
Still not working on Android though...

Bug fix

The Forms MasterDetail Example can't launch because of a missing call to Forms.Init()

Now the example can launch on iOS - but still not Android though...

No

- [ ] All projects build

### DIFF
--- a/MvvmCross-Forms/MvvmCross.Forms.Droid/MvxFormsAndroidSetup.cs
+++ b/MvvmCross-Forms/MvvmCross.Forms.Droid/MvxFormsAndroidSetup.cs
@@ -42,6 +42,8 @@ namespace MvvmCross.Forms.Droid
         {
             get
             {
+                if (!Xamarin.Forms.Forms.IsInitialized)
+                    Xamarin.Forms.Forms.Init(Mvx.Resolve<IMvxAndroidCurrentTopActivity>().Activity, null);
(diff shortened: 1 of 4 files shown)


martijn00:
In the FormsSetup this was done in https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross-Forms/MvvmCross.Forms.iOS/MvxFormsIosSetup.cs#L58 but indeed not called on MasterDetail. I'm wondering if we need to change this a bit so it gets called no matter what presenter you use.

mellson:
I have tried to solve this in a more general way, what do you think?

mellson:
I only tested iOS (which should be ok for this change I guess), with normal and masterdetail presenters.

martijn00:
Aah right, yes, but i think we should probably do something similar in Android and UWP.

mellson:
I have added similar code for Android and UWP. I have tested some of the Android test projects, it's not all of them that can even launch - with or without this change.
I am not able to test the UWP projects. But I don't think this will break anything 🤞

martijn00:
Some of the test projects are a bit broken indeed. Looks good i think, lets see how it goes with this.

PR #2132, opened 2017-08-22
Fix #2096 PictureChooser Android Incorrect Rotation

### LINKED ISSUE #2096: PictureChooser Android Incorrect Rotation
I have two different development devices for testing with, one is an LG Stylus2 running Android 6.0.1 and the other is a Samsung Galaxy Note 4 also on 6.0.1

On the Samsung, pictures selected from the gallery using the PictureChooser plugin are always rendered in the correct orientation. On the LG, the are always rendered in Landscape orientation.

I believe this was also reported in issue [#2017](https://github.com/MvvmCross/MvvmCross/issues/2017) by @mac138, albeit in the comments and it was not the initial problem that the issue was opened for.

### Steps to reproduce
1. Use the PictureChooser plugin to select a picture from the gallery on the device

### Expected behavior
The picture is selected and rotated according to the Orientation attribute in the ExifInterface

### Actual behavior
Samsung Galaxy Note 4: Picture is selected with correct rotation

LG Stylus2: The picture is selected and is always displayed in Landscape mode

### Configuration

**Version:** 4.4, 5.1, 5.1.1

**Platform:** 
- [ ] iOS
- [x] Android
- [ ] WPF
- [ ] UWP

I have added the PictureChooser code to my project and stepped through the issue. The problem lies in MvxPictureChooserTask

On the LG, an exception is caught in the following method, which returns the original sampled bitmap with no orientation changes:

       private Bitmap LoadScaledBitmap(Uri uri)
        {
            ContentResolver contentResolver = Mvx.Resolve<IMvxAndroidGlobals>().ApplicationContext.ContentResolver;
            var maxDimensionSize = GetMaximumDimension(contentResolver, uri);
            var sampleSize = (int) Math.Ceiling(maxDimensionSize / (double) _currentRequestParameters.MaxPixelDimension);
            if (sampleSize < 1)
            {
                // this shouldn't happen, but if it does... then trace the error and set sampleSize to 1
                MvxTrace.Trace(
                    "Warning - sampleSize of {0} was requested - how did this happen - based on requested {1} and returned image size {2}",
                    sampleSize,
                    _currentRequestParameters.MaxPixelDimension,
                    maxDimensionSize);
                // following from https://github.com/MvvmCross/MvvmCross/issues/565 we return null in this case
                // - it suggests that Android has returned a corrupt image uri
                return null;
            }
            var sampled = LoadResampledBitmap(contentResolver, uri, sampleSize);
            try
            {
                var rotated = ExifRotateBitmap(contentResolver, uri, sampled);
                return rotated;
            }
            catch (Exception pokemon)
            {
                Mvx.Trace("Problem seem in Exit Rotate {0}", pokemon.ToLongString());
                return sampled;
            }
        }

The exception results from the following method (called from ExifRotateBitmap(contentResolver, uri, sampled)), as the return statement is returning null:

    private string GetRealPathFromUri(ContentResolver contentResolver, Uri uri)
        {
            var proj = new string[] { MediaStore.Images.ImageColumns.Data };
            using (var cursor = contentResolver.Query(uri, proj, null, null, null))
            {
                var columnIndex = cursor.GetColumnIndexOrThrow(MediaStore.Images.ImageColumns.Data);
                cursor.MoveToFirst();
                return cursor.GetString(columnIndex);
            }
        }

The ExifRotateBitmap method does not use the ExifInterface from the Android Support Library, it uses the one from Android.Media. 
According to [this](https://stackoverflow.com/a/44246248/3592627) StackOverflow post, the Android Support Library should be used due to a security flaw for older devices. What it also highlights is that android.support.media.ExifInterface includes the ability to pass an InputStream into the Constructor for ExifInterface, rather than a URI.
When changing the ExifRotateBitmap method to use the Android Support Library and to use the InputStream to get the Orientation attribute, the image is returned in the correct orientation.

The following is the updated ExifRotateBitmap method that now works correctly on both devices:

    private Bitmap ExifRotateBitmap(ContentResolver contentResolver, Uri uri, Bitmap bitmap)
        {
            if (bitmap == null)
                return null;
            
            using (var inputStream = contentResolver.OpenInputStream(uri))
            {
                var ei = new Android.Support.Media.ExifInterface(inputStream);
                var orientation = ei.GetAttributeInt(Android.Support.Media.ExifInterface.TagOrientation, (int)Orientation.Normal);

                var rotationInDegrees = ExifToDegrees(orientation);
                if (rotationInDegrees == 0)
                    return bitmap;

                using (var matrix = new Matrix())
                {
                    matrix.PreRotate(rotationInDegrees);
                    return Bitmap.CreateBitmap(bitmap, 0, 0, bitmap.Width, bitmap.Height, matrix, true);
                }
            }
    }

jz5:
Bug fix

Exception occurs when converting from URL to file path. So picture rotation is failed.

Use stream from URL, and Android Support Library. As the result, no exception, and picture rotation is successful.
The code is from #2096

No (But needs new packages)

#2096 

- [ ] All projects build

### DIFF
--- a/MvvmCross-Plugins/PictureChooser/MvvmCross.Plugins.PictureChooser.Droid/MvxPictureChooserTask.cs
+++ b/MvvmCross-Plugins/PictureChooser/MvvmCross.Plugins.PictureChooser.Droid/MvxPictureChooserTask.cs
@@ -1,4 +1,4 @@
-﻿// MvxPictureChooserTask.cs
+// MvxPictureChooserTask.cs
 // (c) Copyright Cirrious Ltd. http://www.cirrious.com
 // MvvmCross is licensed using Microsoft Public License (Ms-PL)
 // Contributions and inspirations noted in readme.md and license.txt
@@ -22,11 +22,12 @@ using MvvmCross.Platform.Platform;
 using Path = System.IO.Path;
 using Stream = System.IO.Stream;
 using Uri = Android.Net.Uri;
+using ExifInterface = Android.Support.Media.ExifInterface;
 
 namespace MvvmCross.Plugins.PictureChooser.Droid
 {
     [Preserve(AllMembers = true)]
-	public class MvxPictureChooserTask
+    public class MvxPictureChooserTask
         : MvxAndroidTask, IMvxPictureChooserTask
     {
         private Uri _cachedUriLocation;
@@ -103,7 +104,7 @@ namespace MvvmCross.Plugins.PictureChooser.Droid
 
             _currentRequestParameters = new RequestParameters(maxPixelDimension, percentQuality, pictureAvailable,
                                                               assumeCancelled);
-            StartActivityForResult((int) pickId, intent);
+            StartActivityForResult((int)pickId, intent);
         }
 
         protected override void ProcessMvxIntentResult(MvxIntentResultEventArgs result)
@@ -112,7 +113,7 @@ namespace MvvmCross.Plugins.PictureChooser.Droid
 
             Uri uri;
 
-            switch ((MvxIntentRequestCode) result.RequestCode)
+            switch ((MvxIntentRequestCode)result.RequestCode)
             {
                 case MvxIntentRequestCode.PickFromFile:
                     uri = result.Data?.Data;
@@ -196,7 +197,7 @@ namespace MvvmCross.Plugins.PictureChooser.Droid
         {
             ContentResolver contentResolver = Mvx.Resolve<IMvxAndroidGlobals>().ApplicationContext.ContentResolver;
             var maxDimensionSize = GetMaximumDimension(contentResolver, uri);
-            var sampleSize = (int) Math.Ceiling(maxDimensionSize / (double) _currentRequestParameters.MaxPixelDimension);
+            var sampleSize = (int)Math.Ceiling(maxDimensionSize / (double)_currentRequestParameters.MaxPixelDimension);
             if (sampleSize < 1)
             {
                 // this shouldn't happen, but if it does... then trace the error and set sampleSize to 1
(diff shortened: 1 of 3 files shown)


martijn00:
Looks fine to me except you use v25.4.0.1 for the packages, but those are a bit broken in their bindings. Would it be possible to downgrade for now to v25.3.1?

MvvmCross-Plugins/PictureChooser/MvvmCross.Plugins.PictureChooser.Droid/MvxPictureChooserTask.cs
```diff
@@ -251,27 +251,20 @@ private Bitmap ExifRotateBitmap(ContentResolver contentResolver, Uri uri, Bitmap
...
-                return bitmap;
-
-            using (var matrix = new Matrix())
+            using (var inputStream = contentResolver.OpenInputStream(uri))
             {
-                matrix.PreRotate(rotationInDegrees);
-                return Bitmap.CreateBitmap(bitmap, 0, 0, bitmap.Width, bitmap.Height, matrix, true);
-            }
-        }
+                var ei = new Android.Support.Media.ExifInterface(inputStream);
```
Cheesebaron: use `using Android.Support.Media`

Cheesebaron: Use a more describing name for `ei` it is not a german egg...

jz5: I modified.  There are also Android.Media.ExifInterface class. So I changed to ```using ExifInterface = Android.Support.Media.ExifInterface;```

Cheesebaron: Is there another class called `ExifInterface` since you are making an alias? I think it should be enough to just write `using Android.Support.Media;`

jz5: Yes. There are
```Android.Media.ExifInterface``` and
```Android.Support.Media.ExifInterface```.
Already use ```using Android.Media```
If add```using Android.Support.Media```, can not decide either.

Cheesebaron: It is fine then 👍

PR #2139, opened 2017-08-24
[WIP]Apply async suffix on IMvxNavigationService

CSkoubo:
Still WIP, haven't checked on UWP, documentation etc.

Applies the Async suffix to methods which return Task in the IMvxNavigationService. It has also been done to methods not using the async modifier as the MvxNavigationService is usually used as a baseclass and its good practice to suffix it Async then.

If implementing the IMvxNavigationService, you would need to update your code.

- [ ] All projects build

### DIFF
--- a/MvvmCross/Core/Core/Navigation/IMvxNavigationService.cs
+++ b/MvvmCross/Core/Core/Navigation/IMvxNavigationService.cs
@@ -20,15 +20,15 @@ namespace MvvmCross.Core.Navigation
         event BeforeCloseEventHandler BeforeClose;
         event AfterCloseEventHandler AfterClose;
 
-        Task Navigate<TViewModel>(IMvxBundle presentationBundle = null) where TViewModel : IMvxViewModel;
-        Task Navigate<TViewModel, TParameter>(TParameter param, IMvxBundle presentationBundle = null) where TViewModel : IMvxViewModel<TParameter>;
-        Task<TResult> Navigate<TViewModel, TResult>(IMvxBundle presentationBundle = null, CancellationToken cancellationToken = default(CancellationToken)) where TViewModel : IMvxViewModelResult<TResult>;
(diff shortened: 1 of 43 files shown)


martijn00:
As this is quite a breaking change, i'm really not sure we should do it. This isn't even done in most of .NET's code itself.

Cheesebaron:
> This isn't even done in most of .NET's code itself.
That is a bold claim 😄 

I think we should do this for a breaking release, such as 6.0.

mvanbeusekom:
Here is what Microsoft suggests: "The name of an async method, by convention, ends with an "Async" suffix" (see documentation here: https://msdn.microsoft.com/library/hh191443(vs.110).aspx )

willsb:
Methinks that appending `Async` to methods that do not have a sync counterpart just adds clutter 🤷‍♂️ 

A good read on the topic can be found [here](https://docs.particular.net/nservicebus/upgrades/5to6/async-suffix), in particular:

> **No requirement for conflicting overloads**
> The Async suffix convention was adopted by necessity in .NET CLR since async APIs were added in a non-breaking version. Since C# cannot have overloads that differ only by return type, the new async APIs needed to have a different name, hence the Async suffix was used.

CSkoubo:
Yes its a "major" breaking change, but i think that since the IMvxNavigationService is an essential part of Mvx and people often use the MvxNavigationService as a baseclass it is good practice to make them suffix with Async.

You could argue if you should make the methods without a async modifier "Async", but again i think it makes things clear to the developer that it could be async and most likely is.

modplug:
I agree with @willsb

martijn00:
I've thought a bit on this one, and think we should not do this. We don't use this convention in the rest of our code base as far as i know. Also i agree with the point @willsb makes. Let's close this if we can agree on that.

PR #2143, opened 2017-08-27
Add methods for using combiners with Fluent Bindings

willsb:

This adds custom methods that make using Combiners nicer when using Fluent Bindings

Right now all you can do is use free text

You can now use Expressions, pass each property individually and even pass an instance of your combiner instead of relying on a string that will retrieve a registered combiner.

Nope

Fiddle around with the test projects or try using it to replace existing free text bindings that use combiners to see if behavior is correct.

- [X] All projects build
- [ ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contribute/mvvmcross-docs-style-guide))*

* There's a great lack of docs regarding combiners, so I could not even find a place that describes how to use them in first place. If anyone knows where such docs should be added, please link me so I can provide a full write up on both the new methods and the current way of using combiners :)

### DIFF
--- a/MvvmCross/Core/Binding/BindingContext/MvxBaseFluentBindingDescription.cs
+++ b/MvvmCross/Core/Binding/BindingContext/MvxBaseFluentBindingDescription.cs
@@ -117,26 +117,30 @@ namespace MvvmCross.Binding.BindingContext
         public class CombinerSourceSpec
             : ISourceSpec
         {
+            private readonly bool _useParser;
             private readonly string[] _properties;
             private readonly IMvxValueCombiner _combiner;
 
-            public CombinerSourceSpec(IMvxValueCombiner combiner, string[] properties)
+            public CombinerSourceSpec(IMvxValueCombiner combiner, string[] properties, bool useParser)
             {
                 _combiner = combiner;
+                _useParser = useParser;
                 _properties = properties;
             }
 
             public MvxSourceStepDescription CreateSourceStep(MvxSourceStepDescription inputs)
             {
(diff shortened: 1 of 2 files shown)


martijn00:
This is a great addition. I've been missing this sometimes! There are some docs about combiners here: https://www.mvvmcross.com/documentation/fundamentals/data-binding#tibet

I think we should move it into it's own section and add this to the samples.

MvvmCross/Core/Binding/BindingContext/MvxBaseFluentBindingDescription.cs
```diff
@@ -113,6 +114,37 @@ public MvxSourceStepDescription CreateSourceStep(MvxSourceStepDescription inputs
...
+            {
+                _combiner = combiner;
+                _useParser = useParser;
+                _properties = properties;
+            }
+
+            public MvxSourceStepDescription CreateSourceStep(MvxSourceStepDescription inputs)
+            {
+                var innerSteps = _useParser ?
+                    _properties.Select(p => Mvx.Resolve<IMvxBindingDescriptionParser>().ParseSingle(p).Source) :
```
nmilcoff: Seems like this sentence can be optimized a bit by moving the `Mvx.Resolve<IMvxBindingDescriptionParser>()` outside of the Select call

Cheesebaron: Make a PR or log an issue. This PR is merged and closed 😄

Issue #2061, opened 2017-07-20
Wrong behavior on Move in MvxCollectionViewSourceAnimated


kenkosmowski:
### Steps to reproduce

1. Git Repo: https://github.com/kenkosmowski/MvvmCross-MvxCollectionViewSourceAnimated

2. Run the iOS App.

4. Tap button "Move image!", which moves an item in the ObservableCollection from Index 2 to Index 0.

5. See what happens in the UI.

### Expected behavior
The following item moves are expected:
- Index 2 -> Index 0
- Index 0-> Index 1
- Index 1 -> Index 2

### Actual behavior
- Index 1 -> Index 2
- Index 2 -> gets lost(!)
- Index 0 stays the same

The error occurs in `MvxCollectionViewSourceAnimated.CollectionChangedOnCollectionChangedAsync()` with `NotifyCollectionChangedAction.Move`

I think we should not do this:

```
var oldCount = args.OldItems.Count;
var newCount = args.NewItems.Count;
var indexes = new NSIndexPath[oldCount + newCount];

var startIndex = args.OldStartingIndex;
for (var i = 0; i < oldCount; i++)
    indexes[i] = NSIndexPath.FromRowSection(startIndex + i, 0);
startIndex = args.NewStartingIndex;
for (var i = oldCount; i < oldCount + newCount; i++)
    indexes[i] = NSIndexPath.FromRowSection(startIndex + i, 0);

CollectionView.ReloadItems(indexes);
```
 Because it reloads the items and Index 1 and Index 2, rather than all items.

I think it should be fixed with this code:

```
if (args.NewItems.Count != 1 && args.OldItems.Count != 1)
{
    return;
}

var oldIndexPath = NSIndexPath.FromRowSection(args.OldStartingIndex, 0);
var newIndexPath = NSIndexPath.FromRowSection(args.NewStartingIndex, 0);
this.CollectionView.MoveItem(oldIndexPath, newIndexPath);
```

### Configuration

**Version:** 5.1.0

**Platform:** 
- [x] iOS
- [ ] Android
- [ ] WPF
- [ ] UWP

PR #2159, opened 2017-09-02
Fix CollectionViewSourceAnimated Move Action

nmilcoff:
Bug fix.

Move action doesn't work on `MvxCollectionViewSourceAnimated`

Move action works for `MvxCollectionViewSourceAnimated`

No.

Run the sample project provided by @kenkosmowski: https://github.com/kenkosmowski/MvvmCross-MvxCollectionViewSourceAnimated. It will fail to move items with the current implementation, but it will work when the changes he suggested are applied.

Closes issue #2061

- [X] All projects build

### DIFF
--- a/MvvmCross/Binding/iOS/Views/MvxCollectionViewSourceAnimated.cs
+++ b/MvvmCross/Binding/iOS/Views/MvxCollectionViewSourceAnimated.cs
@@ -80,18 +80,16 @@ namespace MvvmCross.Binding.iOS.Views
             {
                 await CollectionView.PerformBatchUpdatesAsync(() =>
                 {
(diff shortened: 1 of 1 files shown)


softlion:
This introduces breaking change as it won't animate more than one move at a time, and instead reload the whole list. Why not use CollectionViewSource (not animated) instead then ?

I don't agree.

nmilcoff:
A move operation involves only one item at a time. If you move more than
one item at a time and send a single notification to the view then it makes
sense to reload all the data, because you don't have enough information to
reorder all items.
Also there was a bug, did you run the test project that was attached in the
issue?
If you have a better solution in mind please make a PR so we can review it.

PR #2184, opened 2017-09-13
Add a constructor with a binding context

wh1t3cAt1k:

I require an MvxFilteringAdapter that accepts an explicit Android binding context.

There is no such constructor. In my scenario, the adapter is created while there is still no context registration on the stack, and because the base constructor is invoked with a `Current` from the helper (which is `null`), it fails with exception.

The new behavior is simply another constructor with the same common initialization.

No breaking changes expected.

- [X] All projects build

### DIFF
--- a/MvvmCross/Binding/Droid/Views/MvxFilteringAdapter.cs
+++ b/MvvmCross/Binding/Droid/Views/MvxFilteringAdapter.cs
@@ -12,6 +12,7 @@ using Android.Content;
 using Android.Runtime;
 using Android.Widget;
 using Java.Lang;
+using MvvmCross.Binding.Droid.BindingContext;
 using MvvmCross.Platform.Droid;
 using MvvmCross.Platform.Platform;
 using Object = Java.Lang.Object;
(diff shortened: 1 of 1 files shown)


MvvmCross/Binding/Droid/Views/MvxFilteringAdapter.cs
```diff
@@ -117,6 +118,16 @@ public override void NotifyDataSetChanged()
 
         public MvxFilteringAdapter(Context context) : base(context)
```
martijn00: Can you call the base class here and put the `CommonInitialize` code in there? Use `MvxAndroidBindingContextHelpers.Current()` to prevent have to add the parameter.

`public MvxFilteringAdapter(Context context) : base(context, MvxAndroidBindingContextHelpers.Current())`

wh1t3cAt1k: That is precisely what I would like to avoid: the `Current` goes to the registration stack, and there is none at the moment of the adapter creation (I explicitly pass a binding context in my `MvxAdapter` descendant class, but `MvxFilteringAdapter` does not have the appropriate constructor.

martijn00:
I've made the change to the file to what i mean. Can you check if this would work for you?

wh1t3cAt1k:
@martijn00 , this would certainly work for me, thank you!
Shouldn't the first constructor call `this` instead of `base` though?

martijn00:
Good catch! It should call this. Thanks for your help! We would love to see more contributions like this!

Issue #2180, opened 2017-09-12
Support Toolbar and Unified Toolbar bindings by view for Mac


tofutim:
I really like the new MacPresenter - however, since it is View-based, certain WindowController based items like Toolbar and Unified Toolbar are a bit more complicated to tie in. Per @nmilcoff it possible to override MvxMacViewPresenter with a attribute that allows specifying the WindowViewController. I imagine something like 

```
    [MvxFromStoryboard("Main")]
    [MvxWindowPresentation("RootWindowWithToolbarWindowController")]
    public partial class RootView : MvxViewController<RootViewModel>, IMvxOverridePresentationAttribute
	{
        public RootView(IntPtr handle) : base(handle)
        {
            Title = "Root view";
        }
```
and later
```
        public override void ViewDidLoad()
        {
            base.ViewDidLoad();

            var set = this.CreateBindingSet<RootView, RootViewModel>();
            set.Bind(labelTitle).For(v => v.StringValue).To(vm => vm.Title);
            set.Apply();

            var toolbarSet = this.CreateBindingSet<...the toolbar...>, RootViewModel>();
            toolbarSet.Bind(...).To(vm => vm.....);
            toolbarSet.Apply();
        }
```
I'm not sure how I would do the binding, but it would be great to be able to build out the toolbar in Xcode and somehow tie it together.

I should say, I see how I could override 
```
        protected virtual MvxWindowController CreateWindowController(NSWindow window)
        {
            return new MvxWindowController(window);
        }
```
and point to the WindowController with the Toolbar, but that means every WindowController, including my modal dialogs, would have that Toolbar... so I think it has to happen in the attributes.

tofutim:
I'd like to be able to make a Window Controller and Window in Xcode, and then 
```
        protected virtual void ShowToolbarWindowViewController(
            NSViewController viewController,
            ToolbarWindowPresentationAttribute attribute,
            MvxViewModelRequest request)
        {
            var windowController = new ToolbarWindow();     // ToolbarWindow is MvxWindowController
            windowController.ShouldCascadeWindows = true;
            windowController.ShowWindow(null);

            var window = windowController.Window;
            window.Identifier = attribute.Identifier ?? viewController.GetType().Name;
            if (!string.IsNullOrEmpty(viewController.Title))
                window.Title = viewController.Title;

            Windows.Add(window);
            window.ContentView = viewController.View;
            window.ContentViewController = viewController;
        }
```
with
```
[...]

Issue #2178, opened 2017-09-11
Auto-creation of window in MvvmCross Mac now default?


tofutim:
Hi, I just upgraded from MvvmCross 4.4 to 5.1 and notice that my subview is now created in a new window. This seems like new behavior which makes it no longer necessary to create a window in DidFinishLaunching in MvxApplicationDelegate. (For example, in the Mac example here - https://github.com/MvvmCross/MvvmCross-Samples/blob/master/FirstDemo/FirstDemo.Mac/AppDelegate.cs). Can someone who worked on this confirm? If this is expected behavior now, where is it recommended to make changes to the window frame and size?

nmilcoff:
Hey @tofutim, there is a new default ViewPresenter for macOS on MvvmCross 5.1. You can read about it [here](https://www.mvvmcross.com/documentation/platform/mac-view-presenter).

You are right, you don't need to create a window by yourself anymore in DidFinishedLaunching. You can customize the frame/size of your window by using the properties of MvxWindowPresentationAttribute. Also if no attribute is found in your View, then [it might be presented as a window](https://github.com/MvvmCross/MvvmCross/blob/8dd5707ebd8d2158cf0e5197060295bad4ba7f36/MvvmCross/Mac/Mac/Views/Presenters/MvxMacViewPresenter.cs#L239). 

You can see the new presenter in action in [this](https://github.com/MvvmCross/MvvmCross/tree/master/TestProjects/Playground/Playground.Mac) sample.

tofutim:
Thank you!

tofutim:
Do you happen to know if I can use MvvmCross to control the MainMenu options as well? Or is that still separate for now?

nmilcoff:
I'm afraid that isn't supported yet... If you think it's something possible to add, please open a new issue with more details for it (and of course a PR would be more than welcome)

tofutim:
It might be good to add some examples of changing the default window size, I'm still a bit perplexed from the examples how one could do this. Looking into it.

tofutim:
Never mind I figured it out from the samples! Thank you nmilcoff!

nmilcoff:
That is great! :+1: you're very welcome.

tofutim:
@nmilcoff  any thoughts on how to get Toolbar to work with MvvmCross? It typically gets attached to a window instead of a view....

nmilcoff:
[...]

PR #2185, opened 2017-09-13
update MvxWindowPresentation to support creation from Storyboard for Mac

tofutim:
Allows developer to visually create a custom window inside IB (including toolbars, etc.) which will be used to preset the view.

Currently a window is created by code and modified using attribute parameters.

A new attribute for Mac views is introduced. It used the following way:
```
    [MvxFromStoryboard("Main")]
    [StoryboardWindowPresentation("CustomWindow", "Main")]
    public partial class RootView : MvxViewController<RootViewModel>, IMvxOverridePresentationAttribute
	{
        public CustomWindow WindowController {
            get {
                return View.Window != null ? ((CustomWindow)View.Window.WindowController) : null;
            }
        }
```
bindings can be made to outlets in the WindowController using something like
```
            set.Bind(WindowController.labelTitle).For(v => v.StringValue).To(vm => vm.Title);
```

Nope.

Create a new Mac application and a window with custom features in IB. The new window should hold your base view.

https://github.com/MvvmCross/MvvmCross/issues/2180
https://github.com/MvvmCross/MvvmCross/issues/2178

- [x] All projects build
- [x] Nuspec files were updated (when applicable) *n/a*


tofutim:
Modified playground sample to use storyboard ToolbarWindow but you may want a separate button. Wasn't sure if I should change the .Core project to add a new view, so I will leave to you....

tofutim:
I think I might fiddle with the sample a bit more and use the WindowViewModel for this.

tofutim:
I made some changes to the sample so that the StoryboardWindow is used when clicking on Show Window. Please let me know if there are any questions, but it should be pretty clear now. The only catch is that binding to the WindowController has to happen at ViewDidAppear because only then is the Window available.

tofutim:
I thought about this. But there is a problem on what settings should take priority. For example the width. Should it take the width from the attribute or the storyboard? I recommend two separate presenters

martijn00:
What we have done on Android is for example putting a default value of int.MinVal in and then check in the presenter if that is used then use the value of the view(storyboard), but if it is set than take that value.

tofutim:
I still think it is a bit odd, what if the storyboard width is 500, the default width is 200, and the user writes in 200 in the code? you would end up with the storyboard width since the default width is 200. The only way I can think of logically doing this is to have a starting width of something like -1. Then if there is no storyboard, it will set the width to 200.

tofutim:
Ironically I have now created a complex method which got flagged

image

tofutim:
Please let me know if the result is what you are looking for, remarkably it works.

MvvmCross/Mac/Mac/Views/Presenters/Attributes/MvxWindowPresentationAttribute.cs
```diff
@@ -6,18 +6,50 @@ namespace MvvmCross.Mac.Views.Presenters.Attributes
 {
     public class MvxWindowPresentationAttribute : MvxBasePresentationAttribute
     {
-        public float PositionX { get; set; } = 200;
+        public static float DefaultPositionX = 200;
```
martijn00: Did you make those properties static so you can set default values from your own code?

martijn00:
Looks great to me! Just one question about the static values.

MvvmCross/Mac/Mac/Views/Presenters/Attributes/MvxWindowPresentationAttribute.cs
tofutim: No I wasn't that smart. Just didn't think that we should keep making the same DefaultPositionX for each class. But now that you mention it, good idea!

tofutim: Probably initialpositionx should be const static though

tofutim: Not possible?

image

docs/_documentation/presenters/mac-view-presenter.md
```diff
@@ -59,6 +59,19 @@ The presentation can be customized through this properties:
 - TabTitle: Defines the title of the tab that will be displayed in the segmented control for tabs.
 - WindowIdentifier: identifier for the window where the view should be displayed.
 
+### MvxStoryboardWindowPresentationAttribute
```
martijn00: I think this documentation needs to be merged into the old attribute now we will keep that.

MvvmCross/Mac/Mac/Views/Presenters/Attributes/MvxWindowPresentationAttribute.cs
martijn00: Just leave it static. After the docs are updated i'm all good to merge.

docs/_documentation/presenters/mac-view-presenter.md
tofutim: ah yes good catch

MvvmCross/Mac/Mac/Views/Presenters/Attributes/MvxWindowPresentationAttribute.cs
tofutim: I changed the initials to public float {get;}

docs/_documentation/presenters/mac-view-presenter.md
tofutim: i changed the docs

PR #2187, opened 2017-09-13
[WIP] New Xamarin.Forms presenters

martijn00:
This introduces new presenters for Xamarin.Forms. With this you should be able to mix and match Forms with Native views. It's also based on the attribute system.

Yes

- [x] All projects build


MvvmCross-Forms/MvvmCross.Forms.Droid/Views/MvxFormsAndroidViewPresenter.cs
```diff
@@ -0,0 +1,272 @@
...
+        protected virtual bool CloseNavigationPage(IMvxViewModel viewModel, MvxNavigationPagePresentationAttribute attribute)
+        {
+            return true;
+        }
+
+        protected virtual void ShowTabbedPage(
+            Type view,
+            MvxTabbedPagePresentationAttribute attribute,
+            MvxViewModelRequest request)
+        {
```
CSkoubo: Just trying to wrap my head around the presenter implementations.... So for the sake of my mind, you would need to make a MvxTabbedPage which implements the IMvxPage, in order for CreatePage to actually set the VM?

martijn00: `MvxTabbedPage` already implements `IMvxPage` so it will set the VM.

Cheesebaron:
```
Bindings\MvxFormsWindowsBindingBuilder.cs(10,25): error CS0234: The type or namespace name 'Uwp' does not exist in the namespace 'MvvmCross.Binding' (are you missing an assembly reference?) [C:\projects\mvvmcross\MvvmCross-Forms\MvvmCross.Forms.Uwp\MvvmCross.Forms.Uwp.csproj]
Bindings\MvxFormsWindowsBindingBuilder.cs(16,50): error CS0246: The type or namespace name 'MvxWindowsBindingBuilder' could not be found (are you missing a using directive or an assembly reference?) [C:\projects\mvvmcross\MvvmCross-Forms\MvvmCross.Forms.Uwp\MvvmCross.Forms.Uwp.csproj]
Bindings\MvxFormsWindowsBindingBuilder.cs(22,30): error CS0115: 'MvxFormsWindowsBindingBuilder.DoRegistration()': no suitable method found to override [C:\projects\mvvmcross\MvvmCross-Forms\MvvmCross.Forms.Uwp\MvvmCross.Forms.Uwp.csproj]
Bindings\MvxFormsWindowsBindingBuilder.cs(28,61): error CS0115: 'MvxFormsWindowsBindingBuilder.CreateTargetBindingRegistry()': no suitable method found to override [C:\projects\mvvmcross\MvvmCross-Forms\MvvmCross.Forms.Uwp\MvvmCross.Forms.Uwp.csproj]
```

Grrbrr404:
Tryed out the new code with Test Project 'Playground.Forms.Droid'. When I added a new navigation scenario I found out there is a probably a unwanted behavior change. I added a new page 'MixedNavFirstPage' and a new view model 'MixedNavFirstViewModel' and was wondering why my navigation was not working correctly. The Page is a regular MvxContentPage, ViewModels base is MvxViewModel. Not used any of the new attributes.

Inside of class MvxViewModelViewTypeFinder, the Method LookupNamedViewModelType trys to find the ViewModel Named "MixedNavFirstPageViewModel" but the actual name is "MixedNavFirstViewModel". From my perspecvice a lookup for the Page called "MixedNavFirstPage" should result in "MixedNavFirstViewModel"

MvvmCross/Windows/Wpf/Views/Presenters/MvxWpfViewPresenter.cs
```diff
@@ -19,35 +19,56 @@
 namespace MvvmCross.Wpf.Views.Presenters
 {
     public class MvxWpfViewPresenter
-        : MvxBaseWpfViewPresenter
+        : MvxBaseWpfViewPresenter, IMvxAttributeViewPresenter
```
martijn00: @jz5 can you check the changes here?

jz5: WPF playground doesn't work expectedly. (WindowView and WindowChildVIew)
Because, override PresentationAttribute is called so early. The view can't refer ViewModel.
WindowView(Model) and WindowChildVIew(Model) are for multiple parent windows and multiple child contents. (Please run the previous playground)
I think that it is good if there is the workaround for View/ViewModel.

martijn00: The problem is that it was never intended to work with ViewModel because ViewModel shouldn't handle view state. What workaround are you thinking of?

jz5: In this case, I think that View handle state using ViewModel state.
However, I will consider the workaround.

PR #2197, opened 2017-09-16
Add ability to set a default value for an attribute

martijn00:
Add ability to set a default value for an attribute

Have to set a value you want on every attribute every time.

One time.

No

- [ ] All projects build


tofutim:
This looks really good.

martijn00:
I see what you mean with subclassing being cleaner. I just think this makes it a lot easier to change behavior out of the box. Marking the properties as virtual in addition to this would probably be a good idea as well.

Issue #1934, opened 2017-06-09
New presenter for Android


martijn00:
The current Android presenter is very limited. Out of the box there is no support for Fragments, Dialogs, or other navigation patterns. To fix this we need to develop a new presenter which will replace the current one as default. The new presenter should support:

- [x] Fragments (Nested)
- [x] Activity's
- [x] Dialogs
- [x] Tabs / ViewPager

Navigation patterns that should be easy to implement with this are:

- [x] NavigationDrawer
- [ ] BottomNavigationBar
- [x] BottomSheetDialog
- [x] Master/Detail Flows
- [x] Nested navigation

Other requirements are:

- [ ] Unit Tests
- [x] Support for Native and AppCompat views
- [ ] Handle tombstoning
- [x] Easy usage of animations
- [x] Closing of views as well as showing
- [ ] Handling the backstack
- [ ] Caching fragments
- [ ] Comply with the [Android Architecture](https://developer.android.com/topic/libraries/architecture/index.html) components

The plan is to implement it similar to the new iOS presenter, were we have multiple View attributes to indicate how the view will be shown. Also a interface can be implemented to customize view behaviour at runtime.

Cheesebaron:
We should really concentrate on keeping this super simple. The MvxCachingFragmentActivity stuff has grown into a complex Fragments monster.

Hopefully those Android Architecture bits can help us out here a bit.

PelleRavn:
Also it would be nice if the same core-interface (e.g. the tab presenter) on both iOS and Android shared the same interface. Then it would be a lot easier to implement it on both platforms, without having to do something specific for each platform in core.

nmilcoff:
#1917 is a feature request for the presenter

nmilcoff:
Based on #1264 and #1986, it would be great if the new presenter could improve the caching mechanism for fragments, current caching strategy sometimes leads to issues.

The problem is that Android automatically caches the state of all controls that have an `android:id` set. This means that when a fragment is displayed for a second time, it contains old data. The current workaround for that is to avoid caching on such fragments (mainly fragments that contain form inputs)

nmilcoff:
#1373 is also related (fragment caching issue)

nmilcoff:
[...]

PR #2200, opened 2017-09-17
Android ViewPresenter: Simple caching strategy for fragments

### LINKED ISSUE #1264: Binding in an MvxFragment is incorrectly resetting modified values
I have run into an issue with binding in a fragment that I believe is incorrect, and possibly a bug.  The first time a fragment with bound EditText views is shown, everything works as expected.  The second time the same fragment is shown, the form builds as expected, but then calls to the ViewModel property setters reset the values to those from when it was shown previously.

Oddly enough, we discovered that removing the android:id attribute from the element in the layout axml file removes the unwanted behavior for the field in question.

I have isolated the issue in a solution that can be found @ https://github.com/jrmcnair/MvvmCross_BindingIssue.git
### Steps to reproduce
1. Run BindingIssue.Droid Project 
2. Select "Show Second View" button
3. Modify text in both fields
4. Select "Show First View" button
5. Select "Show Second View" button
### Expected behavior

Since the SecondViewModel.Init method resets the values for the two bound properties in the View Model, I would expect the initialized values "First" and "Second" to appear in their respective fields.
### Actual behavior

The Second field is correctly showing the View Model property value, but the first field is incorrectly showing the modified value from the previous display of the view.

The first field has the android:id attribute in the layout axml, and the second field does not.
### Configuration

**Version:** 4.0.0

**Platform:** Android

**Plugins in addition to MvvmCross.StarterPack:**
MvvmCross.Droid.Support.V7.AppCompat v4.0.2
MvvmCross.Droid.Support.V7.Fragging v4.0.1
Xamarin.Android.Support.v7.AppCompat v23.1.1
Xamarin.Android.Support.v4 v23.1.1
MvvmCross.Plugin.Json v4.0.0
Newtonsoft.Json v8.0.2

### LINKED ISSUE #1373: Fragment viewmodel life-cycle events are not called if viewmodel has saved state and calling ShowViewModel<FragmentViewModel>(data) with data
I have three activities and one fragment.
1. MainActivity
2. FragmentHostActivity with a Fragment
3. DetailActivity
### Steps to reproduce
1. Navigate from Main to Host with `ShowViewModel<FragmentViewModel>(data)`.
2. Navigate to Detail from Host (will cause 2 calls to SaveState on FragmentViewModel to occur).
3. Navigate back to Host from Detail, then call Finish on Host.
4. Navigate from Main to host with `ShowViewModel<FragmentViewModel>(data)`.
### Expected behavior

FragmentViewModel Init, ReloadState and Start is called with the **data**.
### Actual behavior

Init, ReloadState and Start is NOT called, the fragment gets the previous view model that was in memory.
### Configuration

**Version:** 4.2.0 with Shared, not FullFragging (what's the difference?) Your sample uses Shared and not FullFragging.

**Platform:** Android

### LINKED ISSUE #1986: Wrong ViewModel data in cached Fragment
I notice an issue while using the new ```NavigationService```. When I navigate multiple times to a ```ViewModel``` with different parameters, the ```Fragment``` shows every time the values of the first call.

### Steps to reproduce

1. First call: 
```c#
await NavigationService.Navigate<ViewModel, Parameter, Result>(new Parameter{ Text = "one" })
```
2. In ```ViewModel``` the ```Initialize``` method will save the parameter in ```ViewModel``` property.

3. In ```Fragement``` the ```OnCreateView``` method will use the value of the Property from the ```ViewModel```.

4. Next call from other location:
```c#
await NavigationService.Navigate<ViewModel, Parameter, Result>(new Parameter{ Text = "two" })
```

### Expected behavior
The ```Fragment``` shows the text "two".

### Actual behavior
The ```Fragment``` shows the text "one".

### Workaround
While debugging I can get the correct value ("two") of the ```ViewModel``` in the ```OnCreateView``` of the ```Fragment``` on second call. But on the Screen I see "one" from the first call. I think it has to do with retrieving thinks from cache, because when I change the ```ShowFragment``` method of the ```Activity``` to replace every ```Fragment```:
```c#
protected override void ShowFragment(string tag, int contentId, Bundle bundle, bool forceAddToBackStack = false, bool forceReplaceFragment = false)
{
	base.ShowFragment(tag, contentId, bundle, forceAddToBackStack, true);
}
```
Then "two" is shown after the second call.

### Configuration

**Version:** 5.0.3

**Platform:** 
- [ ] iOS
- [x] Android
- [ ] WPF
- [ ] UWP

nmilcoff:
Bug fix (we previously had a caching mechanism that is not working now).

When setting `IsCacheableFragment = true` nothing really happens other than bugs.

This PR introduces a simple caching strategy for fragments:
- `IsCacheableFragment` is now false by default.
- Previously OnSaveInstanceState was only run when fragments were cacheable. Now all of them are saved / restored (as expected).
- If `IsCacheableFragment = true` then the Presenter looks for an already instantiated fragment in the FragmentManager of the host activity. The fragment is reused in that case. This is all! Super simple.
- This caching mechanism will be useful when having heavy fragments. An example could also be a fragment containing a WebView with an already loaded URL. 
- Old caching code has been removed as part of this PR.

Probably, if you are using fragments. Also some code has been removed.

Run Playground.Droid and play with it. Especially with RootViewModel, ChildViewModel and SecondChildViewModel.

Android Presenter issue: #1934.
Android Presenter initial PR: #2022 

These issues will be fixed / no longer relevant, since the caching mechamism will change:
Closes #1264
Closes #1986
Closes #1373

- [X] All projects build


nmilcoff:
The PR is essentially ready to be reviewed, once it is approved I'll update the docs.

martijn00:
Can you make sure Appveyor builds?

Cheesebaron:
Btw. the WIP bot will automatically remove the red X when the WIP has been removed from the title.

PR #2213, opened 2017-09-19
Make sure the binding context is created when loading from nib

mvanbeusekom:
Bug fix

The binding context is currently not created when you create a custom implementation of the `MvxView`, load the design from a nib file and construct the view programmatically (as demonstrated here: https://developer.xamarin.com/recipes/ios/general/templates/using_the_ios_view_xib_template/). 

Binding context will be created when the `AwakeFromNib` method is called.

No

Steps to reproduce:
1. Create a custom view and derive from `MvxView` (instead of `UIView`)
2. Load the design from a Nib
3. Programmatically create an instance of the custom view and notice the binding context is not created by the base class.

- [x] All projects build
- [n/a] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contribute/mvvmcross-docs-style-guide))
- [n/a] Nuspec files were updated (when applicable)

### DIFF
--- a/MvvmCross/Binding/iOS/Views/MvxView.cs
+++ b/MvvmCross/Binding/iOS/Views/MvxView.cs
@@ -18,8 +18,6 @@ namespace MvvmCross.Binding.iOS.Views
         : UIView
         , IMvxBindable
     {
-        private bool _isInitialized;
-
         public IMvxBindingContext BindingContext { get; set; }
 
         // Constructor that will bind managed object to its unmanaged counterpart. This constructor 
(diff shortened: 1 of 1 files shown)


KennethKr:
Good job! I'd had to make a hack in the current stable version to be able to use MvxView.

As for the solution I might solve it otherwise to get rid of the bool state:

Alt 1: Add a method "CreateBindingContextOnce", embedding a null check, and calling that in AwakeFromNib/and ctors.

Alt 2: Just do a null check in AwakeFromNib and run CreateBindingContext if it is null

PR #2218, opened 2017-09-21
[WIP] MvxFormsListview with "ItemClick command"

MarcBruins:
feature

None existing

Adding a "MvxFormListView" which has a "ItemClick command". By binding to that command you can get the object that the row is binded to.

No

Run sample

none, but this is going to be documented.

- [x] All projects build


martijn00:
I vote to not prepend every class in the Forms project with `MvxForms` but just `Mvx` like we do everywhere else. The only case where we should add it is where there is a class or method that we override from Core with the same name.

MarcBruins:
This will give you "MvxListView" droid and "MvxListView" Forms which might be confusing

MvvmCross-Forms/MvvmCross.Forms.Droid/MvxFormsAndroidSetup.cs
```diff
@@ -77,6 +80,15 @@ protected override void InitializeBindingBuilder()
             bindingBuilder.DoRegistration();
         }
 
+        protected override void FillTargetFactories(IMvxTargetBindingFactoryRegistry registry)
+        {
+            registry.RegisterPropertyInfoBindingFactory(
```
martijn00: Since this will be the same on all Forms platforms let's make a static helper class and include it in Android, iOS and UWP.

MvvmCross-Forms/MvvmCross.Forms.iOS/MvxFormsIosSetup.cs
```diff
@@ -72,6 +76,15 @@ protected override IEnumerable<Assembly> ValueConverterAssemblies
             }
         }
 
+        protected override void FillTargetFactories(IMvxTargetBindingFactoryRegistry registry)
+        {
+            registry.RegisterPropertyInfoBindingFactory(
+                  typeof(MvxFormsListViewItemClickPropertyTargetBinding),
```
martijn00: Same here.

MvvmCross-Forms/MvvmCross.Forms/Views/MvxFormsListView.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace MvvmCross.Forms.Views
+{
+    public class MvxFormsListView : ListView
+    {
+        public ICommand ItemClick
+        {
+            get;
+            set;
+        } = new Command(() => { });
```
martijn00: Why is there a default value?

martijn00:
No i don't think it will be that confusing. This is for when you target Forms on a Forms Page, you could not include a Android View on this.

MvvmCross-Forms/MvvmCross.Forms/Views/MvxFormsListView.cs
MarcBruins: Testing. There is also no "ItemClick" in the Xaml right now, so this is never going to work. I'm testing why it doesn't work. It's WIP!

MvvmCross-Forms/MvvmCross.Forms.Droid/MvxFormsAndroidSetup.cs
MarcBruins: Sounds good

Cheesebaron:
You can add native views to a Forms page, can't you? In such cases it could be confusing to be allowed to add a MvxListView from Android

PR #2250, opened 2017-10-05
Separate the transition logic

KoenDeleij:
So you can override it without having to implement the stack implementation

Being able to override Android transactions at the MvxAppCompatViewPresenter without having to

### DIFF
--- a/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.AppCompat/MvxAppCompatViewPresenter.cs
+++ b/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.AppCompat/MvxAppCompatViewPresenter.cs
@@ -280,7 +280,7 @@ namespace MvvmCross.Droid.Support.V7.AppCompat
             ft.CommitAllowingStateLoss();
         }
 
-        protected virtual void AddTransitionToFragmentManager(FragmentTransaction ft,MvxFragmentPresentationAttribute attribute){
+        protected virtual void AddTransitionToFragmentManager(FragmentTransaction ft, MvxFragmentPresentationAttribute attribute){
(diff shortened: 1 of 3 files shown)


martijn00:
The View Attributes have a static property to set default values, but i think this is a good idea!

Can you make the same change in the non-appcompat presenter, and possible also do something similar to the Activity transitions.

KoenDeleij:
It saves some duplication. 
I actually got stuck with with the Dictionary of shared elements in the MvxFragmentPresentation attribute. I'm not sure if it works on attributes.

martijn00:
Starting to look good! Only thing i noticed is `AddTransitionToFragmentManager` name might not describe what it is doing because it adds the ability to change anything on the Transaction. Looking at https://github.com/MvvmCross/MvvmCross/blob/5.1.1/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.AppCompat/MvxCachingFragmentCompatActivity.cs#L410 which we had before it might be an idea to add some of those methods back.

KoenDeleij:
It is indeed similar. I'm migrating from 5.1 to 5.2 and used similar functionality in a MvxCachingFragmentActivity.
I'm not sure the naming would make sense. I believe it was the 'OnBeforeFragmentChanging'.

martijn00:
Maybe just add the virtual methods for `OnFragmentChanging` and `OnFragmentChanged` now you're at it. They can off course be empty for now.

KoenDeleij:
this is a bit of a noob question, where can I find the docs for the presenters?

martijn00:
@KoenDeleij https://github.com/MvvmCross/MvvmCross/tree/develop/docs/_documentation/presenters

KoenDeleij:
thanks! Ill update the docs as well

Issue #2246, opened 2017-10-03
[Android] Dialog doesn't survive screen orientation change


tbalcom:
Here is a sample demonstrating the issue:
https://github.com/tbalcom/MvxDialogBackButton

### Steps to reproduce

1. Click the "Show Dialog" button.

2. Rotate the device to change the screen orientation from landscape to portrait, or from portrait to landscape.

3. Finish the dialog using the Yes or No button. Look at the result output to the debug window from the following code: `Mvx.Trace($"ConfirmationViewModel navigation returned {confirmationResult}.");`

### Expected behavior
The dialog should remain on the screen after the rotation has finished, and the correct navigation result should be returned to the caller.

### Actual behavior
There are two scenarios:
#### Using `RetainInstance`
If I use `RetainInstance` then the dialog is closed permanently when the device is rotated and the `Task` is left incomplete.

#### Without using `RetainInstance`
If I don't use `RetainInstance` then the dialog is not closed permanently when the device is rotated. However, the navigation result is **always** `null` if the screen orientation changes at least once.

### Configuration

**Version:** 5.x

**Platform:** 
- [ ] iOS
- [ x ] Android
- [ ] WPF
- [ ] UWP

nmilcoff:
Accordingly to [this answer](https://stackoverflow.com/a/44229759/5000213), the rotation problem seems to be fixed by adding this code (I've personally tested it):

```c#
public override void OnDestroyView()
        {
            if (Dialog != null && RetainInstance)
            {
                Dialog.SetDismissMessage(null);
            }
            base.OnDestroyView();
        }
```

That said, I don't think we should include this workaround in the source code...

PR #2256, opened 2017-10-08
Fix close for DialogFragments when using MvxNavigationService

nmilcoff:
Bug fix.

When using `MvxNavigationService` and a Cancelable dialog, if the user cancels the dialog pressing the back button, the `CloseCompletionSource` hangs and the task never returns to the caller.
Also `MvxAppCompatDialogFragment` doesn't have any ViewModel lifecycle method implemented.

Cancel is correctly managed by dialogs when using `MvxNavigationService`.
Also ViewModel lifecycle methods are implemented on `MvxAppCompatDialogFragment`.

No.

Run Android-Support example and play with the dialog on Login screen.

Issue: #2246 (should be fixed).

- [X] All projects build

### DIFF
--- a/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.Fragment/MvxDialogFragment.cs
+++ b/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.Fragment/MvxDialogFragment.cs
@@ -6,6 +6,7 @@
 // Project Lead - Stuart Lodge, @slodge, me@slodge.com
 
 using System;
+using Android.Content;
 using Android.OS;
 using Android.Runtime;
 using MvvmCross.Binding.BindingContext;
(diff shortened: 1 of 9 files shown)


martijn00:
Looks good to me mostly. Just wondering if `ViewDestroy` will now be called multiple times?

nmilcoff:
Well, I don't think those methods interfere between each other. 
-`Dismiss` and `DismissAllowingStateLoss` are fired only when the user calls them.
-`OnCancel` is called automatically when the back button is pressed. 

-In case of `OnDestroy`, it's true that it _could_ make `ViewModel.ViewDestroy` to be called twice... But the problem is the TaskCompletionSource is set as canceled (only) inside of ViewModel.ViewDestroy. 

The only way to change that would be to cast `ViewModel` to `IMvxViewModelResult<TResult>` (which is impossible because you don't know which generic parameter to use there)

PR #2260, opened 2017-10-10
Update the docs for Android presenters for Fragment lifecycle

KoenDeleij:
Updated the documentation for Android viewpresenters (Fragment lifecycle override)


KoenDeleij:
thanks!

Issue #2117, opened 2017-08-15
Null reference error in MvxFormsAppCompatActivity on GetAccentColor


selaromdotnet:
I am crossposting this from SO (https://stackoverflow.com/questions/45686743/mvvmcross-and-xamarin-forms-android-app-crashes-with-null-reference-error-when) as I'm not sure if the problem I have is coming from XF, Mmx, Android, or just something wrong with my project.

I created a blank Xamarin Forms app using the Visual Studio template and added a toolbar item with an image icon. It runs fine. I updated all nuget to latest, and added MvvmCross (but didn't register the app with Mvx yet) and it still runs fine.

I then changed the app to use the Mvx Forms setup and initialization, and suddenly i'm getting a null reference error when the Xamarin.Forms library tries to call GetAccentColor. 

Like I said I don't know which change caused the error, but I did create and upload a sample project that shows the issue here: https://github.com/selaromdotnet/XamFormsAppTest

The commit history shows the changes I described above, first a blank project, then updating, then finally changing it to use Mvx.

I also found this post (https://blog.xamarin.com/material-design-for-your-xamarin-forms-android-apps/) on how to setup Material design with Xamarin Forms, which I thought might fix the problem, but I still get the same error. To be clear, I got this error immediately upon changing the app to use the Mvx stuff, before I added all the material design changes, but the error is the same.

However, the same app works fine on Windows UWP, so I'm guessing I just have something wrong setup on android, which I don't know enough about to figure out on myown.

### Steps to reproduce

1. Create a blank Xamarin Forms project using visual studio template.

2. Update nuget packages to latest, add MvvmCross

3. Update project to use MvxFormsApplication, Setup and MvxFormsAppCompatActivity

### Expected behavior
App should launch and show the same app from step 1

### Actual behavior
App crashes with null reference error:

> 08-15 01:58:19.234 I/MonoDroid( 2245): System.NullReferenceException: Object reference not set to an instance of an object.
> 08-15 01:58:19.234 I/MonoDroid( 2245):   at Xamarin.Forms.Forms.GetAccentColor () [0x00006] in C:\BuildAgent3\work\ca3766cfc22354a1\Xamarin.Forms.Platform.Android\Forms.cs:159 
[...]

PR #2270, opened 2017-10-14
Fix droid resource issue

johnnywebb:
Bug fix.

Drawables are not appearing specifically for Android leveraging MvvmCross + Forms.

Drawables are displaying as expected.

No.

1. Add an image to a Xaml page
1. Add a drawable to the Forms.Droid project
1. Run Droid

https://github.com/MvvmCross/MvvmCross/issues/2117

- [x] All projects build

### DIFF
--- a/MvvmCross-Forms/MvvmCross.Forms.Droid/Views/MvxFormsAppCompatActivity.cs
+++ b/MvvmCross-Forms/MvvmCross.Forms.Droid/Views/MvxFormsAppCompatActivity.cs
@@ -1,4 +1,4 @@
-﻿using Android.Content;
+using Android.Content;
 using Android.OS;
 using MvvmCross.Binding.BindingContext;
 using MvvmCross.Core.ViewModels;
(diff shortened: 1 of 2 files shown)


Cheesebaron:
Wouldn't it be smarter to just get the Assembly with:

```csharp
var assembly = Assembly.GetCallingAssembly();
```

It seems a bit excessive to involve the IoC in this case...

johnnywebb:
@Cheesebaron good point. However, in the scenario that the MvxFormsApplicationActivity or MvxFormsAppCompatActivity derive from a base activity class in another assembly, then GetCallingAssembly would return the base assembly, not the top level assembly (which I think should contain the resources).

Let me know what you think and I will update accordingly.

Cheesebaron:
Right. I still think it would be a more sane default. Then provide a way to override this default behavior.

martijn00:
By default Forms already calls Assembly.GetCallingAssembly() so we could leave that out of the constructor. The problem is as pointed out that this is in another assembly then where the actually activity is. We could provide this fix, or remove the call to Init and say users should call it themselves

Cheesebaron:
Or just provide a way to pass in their own Assembly, just like everywhere else in MvvmCross, where you can override default behavior.

Cheesebaron:
Lets continue this in #2275

PR #2271, opened 2017-10-14
Add animated option to MvxSidebarPresentationAttribute

Plac3hold3r:
Feature
MvvmCross iOS Support `MvxSidebarPresenter` currently forces new view controllers added with `MvxPanelHintType.PushPanel` to be added with animated of `True`.
The `MvxSidebarPresentationAttribute` now include a bool `Animated` with a default of `True`. Setting it to `False` will prevent the view controller being added with animations.
Yes, the signature for Show methods (`ShowPanel`, `ShowPanelAndResetToRoot` and `ShowPanelAndPopToRoot`) has been modified to make use the attribute instead of just the panel. This allows for easier overriding of the Show methods and having access to the attributes values.
[MvvmCross/TestProjects/iOS-Support/XamarinSidebar/MvvmCross.iOS.Support.Sidebar/](https://github.com/MvvmCross/MvvmCross/tree/develop/TestProjects/iOS-Support/XamarinSidebar/MvvmCross.iOS.Support.Sidebar)
[Docs](https://www.mvvmcross.com/documentation/platform/ios-support-library#mvxsidebarpresenter)

### _Side Note_
The documentation around the `MvvmCross.iOS.Support` will need some updating `MvxPanelPresentation` no longer exists. 

- [X] All projects build

### DIFF
--- a/MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresentationAttribute.cs
+++ b/MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresentationAttribute.cs
@@ -1,5 +1,4 @@
-﻿using MvvmCross.Core.Views;
-using MvvmCross.iOS.Views.Presenters.Attributes;
+using MvvmCross.Core.Views;
 
 namespace MvvmCross.iOS.Support.XamarinSidebar
 {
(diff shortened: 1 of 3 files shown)


MvvmCross-iOSSupport/MvvmCross.iOS.Support.XamarinSidebar/MvxSidebarPresentationAttribute.cs
```diff
@@ -26,18 +25,25 @@ public class MvxSidebarPresentationAttribute : MvxBasePresentationAttribute
...
+        /// <summary>
         /// Initializes a new instance of the <see cref="MvxSidebarPresentationAttribute"/> class.
         /// </summary>
         /// <param name="panel">The panel.</param>
         /// <param name="hintType">Type of the hint.</param>
         /// <param name="showPanel">if set to <c>true</c> [show panel].</param>
         /// <param name="behaviour">The splitview behaviour value</param>
-        public MvxSidebarPresentationAttribute(MvxPanelEnum panel, MvxPanelHintType hintType, bool showPanel, MvxSplitViewBehaviour behaviour = MvxSplitViewBehaviour.None)
+        /// <param name="animated">if set to <c>true</c> animate the View Controller.</param>
+        public MvxSidebarPresentationAttribute(MvxPanelEnum panel, MvxPanelHintType hintType, bool showPanel, MvxSplitViewBehaviour behaviour = MvxSplitViewBehaviour.None, bool animated = true)
```
Plac3hold3r: Related to the Codacy issues:

Should I rather use an overloading constructor pattern or make `Animated` a property? I had chosen to follow the pattern already in place with the `MvxSplitViewBehaviour` for this attribute. Looking at other presentation attributes used in Mvx the patterns used seems to vary between constructor with optional parameters and properties. 

Additionally, some Mvx presenter attributes defaults are referenced via a public static field, I do not understand why it is done this way. My suggestion would be to use constructor parameters for the minimum required values and properties for everything else.

From the [docs.microsoft](https://docs.microsoft.com/en-us/dotnet/standard/attributes/writing-custom-attributes#declaring-constructors)

> You can overload the constructor to accommodate different combinations of values. If you also define a property for your custom attribute class, you can use a combination of named and positional parameters when initializing the attribute. Typically, you define all required parameters as positional and all optional parameters as named. In this case, the attribute cannot be initialized without the required parameter. All other parameters are optional

martijn00: I agree that this gives some confusion and we should clean it up. Can you make an issue for this so we can take this up in the 6.0 release? The static fields are to set default values for all the attributes of that type. That way you don't need to inherit a attribute to be able to set a default value.

Issue #2307, opened 2017-10-29
Xamarin.Forms / MasterDetailPage: The master menu stays open after navigation


Sebastian1989101:
While the default Xamarin.Forms master menu page is closed after the navigation, I cannot reproduce this with MVX. The menu stays always open.

martijn00:
You need to trigger it from your navigate method:

```c#
if(Xamarin.Forms.Application.Current.MainPage is MasterDetailPage masterDetailPage)
{
    masterDetailPage.IsPresented = false;
}
else if(Xamarin.Forms.Application.Current.MainPage is NavigationPage navigationPage && navigationPage.CurrentPage is MasterDetailPage nestedMasterDetail)
{
    nestedMasterDetail.IsPresented = false;
}
```

djmpnt:
> 
> 
> While the default Xamarin.Forms master menu page is closed after the navigation, I cannot reproduce this with MVX. The menu stays always open.

Just put this IsPresented = false;
Inside the method.

Example: 

private void GoToPerfil1(object sender ,EventArgs args)
        {
            Detail = new NavigationPage(new Pages.Perfil1());
            
            **IsPresented = false;**
        }

Issue #2306, opened 2017-10-29
Xamarin.Forms / MasterDetailPage breaks my app because of the slide to open menu gesture


Sebastian1989101:
While I can disable the swipe to open of the master menu gesture, I cannot find any option for that in MVX. But it breaks my whole application. Here is, how it can be done in classic Xamarin.Forms: https://stackoverflow.com/questions/40515952/xamarin-form-master-detail-page-how-to-disable-swipe-gesture-to-load-the-men

martijn00:
You could do something like this:

```c#
#if __IOS__
if(Xamarin.Forms.Application.Current.MainPage is MasterDetailPage masterDetailPage)
{
    masterDetailPage.IsGestureEnabled = false;
}
else if(Xamarin.Forms.Application.Current.MainPage is NavigationPage navigationPage && navigationPage.CurrentPage is MasterDetailPage nestedMasterDetail)
{
    nestedMasterDetail.IsGestureEnabled = false;
}
#endif
```

Sebastian1989101:
Is this still required for ignoring the swipe gesture? And if so: Will this work with the MVX MasterDetail presenter? Because my page is only a "MvxContentPage" just as in the Playground project.

martijn00:
I don't know, you have to try it yourself if it works.

mvcaction:
simply you can put IsGestureEnabled="False" in RootPage.xaml

Issue #2292, opened 2017-10-24
Navigation between native and Forms is not correct


YvesDelcoigne:
To help us fix your issue, please provide the information in the below template. 

> Note: There is often little we can do without a minimal reproducible sample of the issue, so please provide that in a **standalone git repository** and link it here.
>
> Please only file bugs verified on latest version of MvvmCross and _stable_ Xamarin tools

### Steps to reproduce

1. Open and run the Playground.Forms.Droid or iOS sample app

2. Press the "Show Override" button

3. Press the "Show second child" button

### Expected behavior
Second child is shown.

### Actual behavior
Override screen is still there, second child screen is added in the backstack (it's visible when you press the back button)

### Configuration

**Version:** 5.3.2

**Platform:** 
- [x] iOS
- [x] Android
- [?] WPF
- [?] UWP

martijn00:
Currently there is no automatic closing or moving forward from a native view to a Forms view. If you manually call close first and then navigate it should work.

YvesDelcoigne:
Okay, so it's actually an improvement and not a bug. Manually closing it does indeed work, but it doesn't feel right.

PR #2321, opened 2017-10-30
Fixes for Forms presenter

martijn00:
Bug fixes

Fixes https://github.com/MvvmCross/MvvmCross/issues/2292
Fixes https://github.com/MvvmCross/MvvmCross/issues/2307
Fixes https://github.com/MvvmCross/MvvmCross/issues/2306

- [x] All projects build


MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.AppCompat/MvxAppCompatViewPresenter.cs
```diff
@@ -487,6 +487,20 @@ protected virtual bool CloseViewPagerFragment(IMvxViewModel viewModel, MvxViewPa
             return false;
         }
 
+        protected override bool CloseFragments()
+        {
+            try
+            {
+                CurrentFragmentManager.PopBackStackImmediate();
```
nmilcoff: Can we run this code only if there are entries on the backstack instead of running it inside of a try/catch block?

MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.AppCompat/MvxAppCompatViewPresenter.cs
```diff
@@ -487,6 +487,20 @@ protected virtual bool CloseViewPagerFragment(IMvxViewModel viewModel, MvxViewPa
...
+            }
+            catch (Exception ex)
+            {
+                return base.CloseFragments();
+                MvxTrace.Trace("Cannot close any fragments", ex);
```
nmilcoff: MvxTrace is obsolete now, should we use the new logging system here?

MvvmCross-Forms/MvvmCross.Forms.Uwp/Views/MvxFormsUwpViewPresenter.cs
```diff
@@ -129,5 +130,10 @@ private bool TryShowPage(MvxViewModelRequest request)
         public override void Close(IMvxViewModel toClose)
         {
         }
+
+        public virtual bool ClosePlatformViews()
+        {
+            return false;
```
nmilcoff: If the feature isn't implemented here can we leave a trace to make it easier to discover?

MvvmCross-Forms/MvvmCross.Forms/ViewModels/IMvxMasterDetailViewModel.cs
```diff
@@ -1,26 +0,0 @@
...
-using MvvmCross.Core.ViewModels;
-
-namespace MvvmCross.Forms.ViewModels
-{
-    /// <summary>
-    /// Extends IMvxViewModel interface to have track of the ViewModel used in Detail as root.
-    /// Also, it exposes a method to be  launched when this root page is activated due to navigation popping in Detail
-    /// 
-    /// </summary>
-    internal interface IMvxMasterDetailViewModel : IMvxViewModel
```
nmilcoff: Isn't this a breaking change? Same with `MvxMasterDetailViewModel`

MvvmCross/Droid/Droid/Views/MvxAndroidViewPresenter.cs
```diff
@@ -594,6 +594,19 @@ protected virtual bool CloseFragmentDialog(IMvxViewModel viewModel, MvxDialogFra
             return false;
         }
 
+        protected virtual bool CloseFragments()
+        {
+            try
+            {
+                CurrentFragmentManager.PopBackStackImmediate();
```
nmilcoff: Same here about running the code without try/catch

MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.AppCompat/MvxAppCompatViewPresenter.cs
martijn00: This is too prevent AppCompat and normal fragment type exceptions.

MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.AppCompat/MvxAppCompatViewPresenter.cs
martijn00: It is used all over the code base. I think we need to see how we can replace it all.

MvvmCross-Forms/MvvmCross.Forms/ViewModels/IMvxMasterDetailViewModel.cs
martijn00: This already was removed in 5.3 but i forgot to remove the viewmodels.

MvvmCross-Forms/MvvmCross.Forms.Uwp/Views/MvxFormsUwpViewPresenter.cs
martijn00: Sure

MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.AppCompat/MvxAppCompatViewPresenter.cs
nmilcoff: Alright, I think that makes sense as long as we print the exception. Does it make sense to call base in this context, when base uses a different fragment manager @martijn00? 
Sounds like in all cases two exceptions will be written in the output because of that.

martijn00: That is exactly why i call base, because base has a different manager which might be the correct type.

nmilcoff: Sorry for insisting but... When using the AppCompatViewPresenter there's no way to show a non-compat Fragment, [ShowFragment](https://github.com/martijn00/MvvmCross/blob/025dfabe03308c270661a7b6affe3b108334a5d3/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.AppCompat/MvxAppCompatViewPresenter.cs#L201) does not call base in any case. 

I think there are no scenarios where the base FragmentManager could resolve the request

martijn00: Hmm yeah you are right with that. I would still leave the try catch but remove the base call.

nmilcoff: Yes, good with me 👍

PR #2348, opened 2017-11-03
Improve UWP view presenter to use presentation attributes

Grrbrr404:
### :sparkles: What kind of change does this PR introduce? (Bug fix, feature, docs update...)
Changing the UWP view presenter to make use of new attribute system.

### :arrow_heading_down: What is the current behavior?
Presenter is not able to make us of attributes.

### :boom: Does this PR introduce a breaking change?
Yes, UWP Pages need annotations now.

### :thinking: Checklist before submitting

- [ ] All projects build
- [ ] Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))
- [ ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contribute/mvvmcross-docs-style-guide))
- [ ] Nuspec files were updated (when applicable)
- [ ] Rebased onto current develop

PR #2366, opened 2017-11-08
Improve UWP view presenter to use presentation attributes v2

nickrandolph:
Changing the UWP view presenter to make use of new attribute system.

Presenter is not able to make us of attributes.

Yes, UWP Pages need annotations now 

This supersedes the existing PR: https://github.com/MvvmCross/MvvmCross/pull/2348

- [x] All projects build

Known issues: 
XForms for UWP doesn't work and is not fixed in the PR
There is still some clean up to be done around multiple presenters
Playground.UWP still references XForms - removing this causes an exception. Need to separate UI/ViewModels into separate projects and fix this


Cheesebaron:
@nickrandolph can you take a look at the issues CodeFactor is pointing out?

nickrandolph:
@Cheesebaron  I've addressed the obvious layout issues raised by CodeFactor. The remaining issues are due to the duplication that was already present with the viewpresenters. This PR already reduces this with the MvxAttributeViewPresenter. I'm not familiar enough with the other platforms to know how to refactor the code from those platforms where there is still code duplicity

MvvmCross/Core/Core/Views/MvxAttributeViewPresenter.cs
```diff
@@ -0,0 +1,97 @@
...
+                }
+                return _attributeTypesActionsDictionary;
+            }
+        }
+
+        public abstract void RegisterAttributeTypes();
+
+        public abstract MvxBasePresentationAttribute CreatePresentationAttribute(Type viewModelType, Type viewType);
+
+        public virtual MvxBasePresentationAttribute GetOverridePresentationAttribute(Type viewModelType, Type viewType)
```
martijn00: Should this not be abstract as well?

MvvmCross/Core/Core/Views/MvxAttributeViewPresenter.cs
```diff
@@ -0,0 +1,97 @@
...
+using MvvmCross.Core.ViewModels;
+using MvvmCross.Platform;
+using System;
+using System.Reflection;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace MvvmCross.Core.Views
+{
+    public abstract class MvxAttributeViewPresenter : MvxViewPresenter, IMvxAttributeViewPresenter
```
martijn00: Good idea to do this!

MvvmCross/Core/Core/Platform/LogProviders/ConsoleLogProvider.cs
```diff
@@ -178,7 +178,8 @@ private static class ConsoleColorValues
 
             static ConsoleColorValues()
             {
-                Type = Type.GetType("System.ConsoleColor");
+                Type = Type.GetType("System.ConsoleColor") ?? Type.GetType("System.ConsoleColor, System.Console");
```
martijn00: @willsb what do you think of this?

MvvmCross/Windows/Uwp/Views/MvxWindowsViewPresenter.cs
```diff
@@ -28,34 +33,111 @@ public MvxWindowsViewPresenter(IMvxWindowsFrame rootFrame)
...
 
-        private IMvxViewModelTypeFinder _viewModelTypeFinder;
-        public IMvxViewModelTypeFinder ViewModelTypeFinder
+        public override void RegisterAttributeTypes()
+        {
+            AttributeTypesToActionsDictionary.Add(
+                typeof(MvxPagePresentationAttribute),
+                new MvxPresentationAttributeAction
+                {
+                    ShowAction = (view, attribute, request) => ShowPage(request),
```
martijn00: It's probably best to send the view, attribute and request parameters already to this method. You might not need them now, but maybe later, or someone overriding this needs it.

MvvmCross/Windows/Uwp/Views/MvxWindowsViewPresenter.cs
```diff
@@ -28,34 +33,111 @@ public MvxWindowsViewPresenter(IMvxWindowsFrame rootFrame)
...
+                {
+                    ShowAction = (view, attribute, request) => ShowSplitView((MvxSplitViewPresentationAttribute)attribute, request),
+                    CloseAction = (viewModel, attribute) => CloseSplitView(viewModel)
+                });
+        }
+
+        public override MvxBasePresentationAttribute CreatePresentationAttribute(Type viewModelType, Type viewType)
+        {
+            MvxTrace.Trace($"PresentationAttribute not found for {viewType.Name}. Assuming new page presentation");
+            return new MvxPagePresentationAttribute();
```
martijn00: Is there no way to detect a splitview?

MvvmCross/Windows/Uwp/Views/MvxWindowsViewPresenter.cs
```diff
@@ -109,49 +173,86 @@ protected virtual string GetRequestText(MvxViewModelRequest request)
...
 
             _rootFrame.GoBack();
 
             HandleBackButtonVisibility();
+
+            return true;
         }
 
-        protected virtual void HandleBackButtonVisibility()
+        private void ShowPage(MvxViewModelRequest request)
```
martijn00: Should be `protected virtual`

MvvmCross/Windows/Uwp/Views/MvxWindowsViewPresenter.cs
```diff
@@ -109,49 +173,86 @@ protected virtual string GetRequestText(MvxViewModelRequest request)
...
+        }
 
-            MvxTrace.Warning("Hint ignored {0}", hint.GetType().Name);
+        private bool CloseSplitView(IMvxViewModel viewModel)
+        {
+            return ClosePage(viewModel);
         }
 
-        public override void Close(IMvxViewModel viewModel)
+        private bool ClosePage(IMvxViewModel viewModel)
```
martijn00: Should be `protected virtual`

MvvmCross/Windows/Uwp/Views/MvxWindowsViewPresenter.cs
```diff
@@ -109,49 +173,86 @@ protected virtual string GetRequestText(MvxViewModelRequest request)
...
+                }
+                else if (attribute.Position == SplitPanePosition.Pane)
+                {
+                    splitView.Pane = (UIElement)Activator.CreateInstance(viewType);
+                }
             }
+        }
 
-            MvxTrace.Warning("Hint ignored {0}", hint.GetType().Name);
+        private bool CloseSplitView(IMvxViewModel viewModel)
```
martijn00: Should be `protected virtual`

TestProjects/Playground/Playground.Forms.Uwp/Setup.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        protected override MvxFormsApplication CreateFormsApplication()
+        {
+            return new Core.FormsApp();
+        }
+
+        protected override IMvxWindowsViewPresenter CreateViewPresenter(IMvxWindowsFrame rootFrame)
+        {
+            var presenter = new MvxFormsUwpViewPresenter(rootFrame, FormsApplication);
```
martijn00: This should not be overriden here.

TestProjects/Playground/Playground.Uwp/project.json
```diff
@@ -1,9 +1,10 @@
 ﻿{
   "dependencies": {
-    "Microsoft.NETCore.UniversalWindowsPlatform": "5.2.2"
+    "Microsoft.NETCore.UniversalWindowsPlatform": "5.2.2",
+    "NETStandard.Library": "2.0.1"
   },
   "frameworks": {
-    "uap10.0.10586": {}
+    "uap10.0.10240": {}
```
martijn00: Please consolidate all UWP projects to "uap10.0.10586"

MvvmCross/Core/Core/Platform/LogProviders/ConsoleLogProvider.cs
nickrandolph: I was actually thinking we can drop the initial GetType and just use the assembly qualified call, but will that work across all platforms?

TestProjects/Playground/Playground.Uwp/project.json
nickrandolph: @martijn00 is there a reason we want 10586, rather than 10240?

martijn00: After this we want to update to fall creators update to support .NET Standard 2.0. Other then that we have all Uwp projects target uap10.0.10586 but set minimum to uap10.0.10240

MvvmCross/Core/Core/Views/MvxAttributeViewPresenter.cs
nickrandolph: @martijn00 it looked like this was a common implementation for this method, so think prob reasonable to have as base implementation. Can make abstract but then we'll end up with code duplication across the platform implementations

MvvmCross/Windows/Uwp/Views/MvxWindowsViewPresenter.cs
nickrandolph: @martijn00 not without creating an instance of the viewType - not sure we should be encouraging that behaviour from developers as it'll only be negative for perf

TestProjects/Playground/Playground.Uwp/project.json
nickrandolph: @martijn00 did you mean for target to be 10586 or the creators update.
Will set min to 10240

nickrandolph:
@Cheesebaron @martijn00  I've merged the second presenter and added MvxRegionPresentationAttribute. Have also adjusted UWP min and target version. Let me know what other changes are required for this PR

MvvmCross/Core/Core/Platform/LogProviders/ConsoleLogProvider.cs
Grrbrr404: When I started working on this I had issues with  Type.GetType("System.ConsoleColor") on runtime, always returning null. So the workarround with Type.GetType("System.ConsoleColor, System.Console"); worked for me.

PR #2375, opened 2017-11-10
Improvements to UIDatePicker target bindings

DaRosenberg:

Features and bug fixes.

MvvmCross does not directly deal with dates in many places, but one such place is the bindings that target a `UIDatePicker` in iOS:

- MvxUIDatePickerDateTargetBinding
- MvxUIDatePickerTimeTargetBinding
- MvxUIDatePickerCountdownDurationTargetBinding
- MvxBaseUIDatePickerTargetBinding (abstract base class for the above)

The current implementation has several issues:

1. It uses the .NET `DateTime` class (implemented by Mono) to do time zone conversion, and it uses the explicit cast conversions between `DateTime` and `NSDate` (provided by Xamarin) to convert between the two. These implementations do not respond to time zone changes **while the app is running**. They initialize their notion of current system time zone when the app is started, and if time zone of the device changes while the app is running, they no longer produce the correct results.

2. When doing a two-way binding to the `UIDatePicker.Date` property, the binding converts a local `DateTime` to an `NSDate` in universal time (which is correct) but it does not perform the same conversion back to the view model when the `UIDatePicker` value is changed. Instead it writes back a UTC `DateTime` value, leading to strange behavior. To achieve correct results, an app must work around this by performing the write-back conversion using a converter.

3. There is no first-class support for binding to the `MinimumDate` and `MaximumDate` properties on `UIDatePicker`. If an app wants to do so, it must remember to do the conversion to `NSDate` and universal time using a converter.

This PR addresses the above issues in the following ways:

1. It changes all `UIDatePicker` target bindings listed above to use iOS-specific time zone conversion using the `NSTimeZone` class instead of relying on Mono implementation of `DateTime` methods. This makes the app responsive to time zone changes while the app is running.

2. It corrects `MvxUIDatePickerDateTargetBinding` to yield a proper local `DateTime` instead of a UTC `DateTime` with an unspecified Kind.

3. It adds a `MvxUIDatePickerMinMaxTargetBinding` for binding to `MinimumDate` and `MaximumDate` (using the same improved time zone conversion logic).

It also adds a useful extension method to `MvxIosDateTimeExtensionMethods` used from some of the above classes.

Very unlikely.

Apps which have previously worked around issue number 2 above using `DateTime.ToLocalTime()` will **not** be broken, because this method performs no conversion if the source `DateTime` already has its `Kind` property set to `Local`, which the improved binding code does. This should constitute the vast majority of those apps that even use this binding.

However, apps which have worked around that issue in some other strange ways (potentially involving homegrown timezone conversion logic) might be broken by this change, but honestly it's even hard to imagine that many such cases exist.

Any test cases that relate to binding between `DateTime` or `TimeSpan` values on the view model, and various properties on a `UIDatePicker` in an iOS app. In particular when running on a time zone different from UTC. I also encourage to test changing the device time zone while the app is running, which now works.

N/A

- [X] All projects build

### DIFF
--- a/MvvmCross/Binding/iOS/Target/MvxBaseUIDatePickerTargetBinding.cs
+++ b/MvvmCross/Binding/iOS/Target/MvxBaseUIDatePickerTargetBinding.cs
@@ -1,4 +1,4 @@
-﻿// MvxBaseUIDatePickerTargetBinding.cs
+// MvxBaseUIDatePickerTargetBinding.cs
 
 // MvvmCross is licensed using Microsoft Public License (Ms-PL)
 // Contributions and inspirations noted in readme.md and license.txt
@@ -7,7 +7,9 @@
 
 using System;
 using System.Reflection;
+using Foundation;
 using MvvmCross.Binding.Bindings.Target;
+using MvvmCross.Platform.iOS;
 using MvvmCross.Platform.Platform;
 using UIKit;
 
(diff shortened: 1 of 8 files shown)


MvvmCross/Binding/iOS/Target/MvxBaseUIDatePickerTargetBinding.cs
```diff
@@ -21,15 +23,18 @@ protected MvxBaseUIDatePickerTargetBinding(object target, PropertyInfo targetPro
             var datePicker = View;
             if (datePicker == null)
             {
-                MvxBindingTrace.Trace(MvxTraceLevel.Error,
-                                      "Error - UIDatePicker is null in MvxBaseUIDatePickerTargetBinding");
+                MvxBindingTrace.Trace(MvxTraceLevel.Error, "Error - UIDatePicker is null in MvxBaseUIDatePickerTargetBinding");
             }
-            else
+            else if (targetPropertyInfo.Name == nameof(UIDatePicker.Date)) // Only listen for changes if we are binding against the Date property.
```
sfmskywalker: @DaRosenberg This line is breaking the target binding for `CountDownDuration` changes.

DaRosenberg: Yep, I can see why. I'll fix.

PR #2379, opened 2017-11-13
Separation of concerns for Playground sample

nickrandolph:
Update Samples

Both view and view models are included in the same project (Playground.Core) which means that all head projects need to reference Xamarin Forms. 

Views have been moved into their own project (Playground.UI).
This demonstrates good separation of concerns for the Playground sample

No, as it does change any behaviour of the libraries themselves

Make sure Playground sample runs on all target platforms

N/A

- [ ] All projects build


MvvmCross/Platform/Platform/ReflectionExtensions.cs
```diff
@@ -44,7 +44,7 @@ public static IEnumerable<ConstructorInfo> GetConstructors(this Type type)
 
         public static bool IsInstanceOfType(this Type type, object obj)
         {
-            return type.IsAssignableFrom(obj.GetType()) || obj.IsMarshalByRefObject();
+            return obj != null && (type.IsAssignableFrom(obj.GetType()) || obj.IsMarshalByRefObject());
```
nmilcoff: This change isn't mentioned in the PR description, can you elaborate on it?

nmilcoff: This change isn't mentioned in the PR description, can you elaborate on it?

TestProjects/Playground/Playground.Core/ViewModels/MixedNavMasterDetailViewModel.cs
```diff
@@ -63,14 +57,15 @@ public MixedNavMasterDetailViewModel(IMvxNavigationService navigationService)
...
                     var vmType = item.ViewModelType;
-                    if(Xamarin.Forms.Application.Current.MainPage is MasterDetailPage masterDetailPage)
-                    {
-                        masterDetailPage.IsPresented = false;
-                    }
-                    else if(Xamarin.Forms.Application.Current.MainPage is NavigationPage navigationPage && navigationPage.CurrentPage is MasterDetailPage nestedMasterDetail)
-                    {
-                        nestedMasterDetail.IsPresented = false;
-                    }
+                    // TODO: This logic doesn't belong here
```
nmilcoff: Does the sample works without this code? Do you know where to put it if it doesn't belong there anymore?

MvvmCross/Platform/Platform/ReflectionExtensions.cs
nickrandolph: This was just a minor change - if this method is accidentally called with obj == null, it'll throw an exception. 
The fix, allows this method to be called passing in null

TestProjects/Playground/Playground.Core/ViewModels/MixedNavMasterDetailViewModel.cs
nickrandolph: Ah good question. The sample definitely runs - setting the IsPresented to false controls whether the master page is visible. I've left the TODO in there as the samples could probably do with further refining, particularly once we've fixed the Forms implementation for MasterDetails (which doesn't really work at the moment)

MvvmCross/Platform/Platform/ReflectionExtensions.cs
Cheesebaron: I think this should be in a separate PR.

TestProjects/Playground/Playground.Core/Playground.Core.csproj
```diff
@@ -1,304 +1,25 @@
...
-    <EmbeddedResource Include="Pages\MixedNavTab2Page.xaml">
-      <Generator>MSBuild:UpdateDesignTimeXaml</Generator>
-      <SubType>Designer</SubType>
-    </EmbeddedResource>
+    <EmbeddedResource Include="Resources\Text.json" />
   </ItemGroup>
+
   <ItemGroup>
-    <Folder Include="Services\" />
+    <ProjectReference Include="..\..\..\MvvmCross-Plugins\JsonLocalization\MvvmCross.Plugins.JsonLocalization\MvvmCross.Plugins.JsonLocalization.csproj" />
```
martijn00: We should not use project reference until we go over on all projects. For a sample it would work right now, but if you do it for one of the framework projects it would crash.

TestProjects/Playground/Playground.Core/ViewModels/MixedNavMasterDetailViewModel.cs
martijn00: I think we should fix this together with the implementation in a new PR.

TestProjects/Playground/Playground.Droid/Playground.Droid.csproj
```diff
@@ -16,7 +16,7 @@
     <AndroidManifest>Properties\AndroidManifest.xml</AndroidManifest>
     <MonoAndroidResourcePrefix>Resources</MonoAndroidResourcePrefix>
     <MonoAndroidAssetsPrefix>Assets</MonoAndroidAssetsPrefix>
-    <AndroidUseLatestPlatformSdk>false</AndroidUseLatestPlatformSdk>
+    <AndroidUseLatestPlatformSdk>true</AndroidUseLatestPlatformSdk>
```
martijn00: Please set not to latest. We want to control which we use.

TestProjects/Playground/Playground.Forms.Droid/Playground.Forms.Droid.csproj
```diff
@@ -10,14 +10,14 @@
     <OutputType>Library</OutputType>
     <RootNamespace>Playground.Forms.Droid</RootNamespace>
     <AssemblyName>Playground.Forms.Droid</AssemblyName>
-    <TargetFrameworkVersion>v8.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v7.1</TargetFrameworkVersion>
```
martijn00: Please revert to 8.0 and set not to use latest.

TestProjects/Playground/Playground.Forms.UI/Playground.Forms.UI.csproj
```diff
@@ -0,0 +1,104 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard1.4</TargetFramework>
```
martijn00: Can you use the PCL style projects until we switch?

TestProjects/Playground/Playground.Forms.Uwp/Setup.cs
```diff
@@ -21,9 +24,18 @@ protected override IMvxApplication CreateApp()
...
+        {
+            return base.GetViewAssemblies().Union(new[] { typeof(FormsApp).GetTypeInfo().Assembly });
+        }
+
         protected override MvxFormsApplication CreateFormsApplication()
         {
-            return new Core.FormsApp();
+            return new FormsApp();
+        }
+        protected override IMvxLogProvider CreateLogProvider()
```
martijn00: This should be removed and type set to none.

PR #2385, opened 2017-11-13
[WIP] Update to NetStandard 2.0

martijn00:

Updates all projects to NetStandard 2.0 using SDK style project files. Also using Multi-target where possible.

TODO:
- [ ] Make everything compile
- [x] Add Android-support
- [ ] Add Code analyzers
- [ ] Rework samples
- [ ] Rework plugins to new style and bait & switch
- [x] Remove old pcl code
- [ ] DownloadCache plugin doesn't work for Mac

PCL projects.

.NET Standard

Yes

Everything

- [ ] All projects build


Cheesebaron:
Not sure we want to do bait & switch for plugins in this one. Just to keep this PR quick and less complicated I think we should just leave the plugins for now, still convert them to NET Standard, but leave the bootstrap etc. Then make a separate PR for converting the plugins.


One current issue we have in the new sln:
- [ ] MvxReflectionExtensions contain extensions with the same name as other stuff from System.Reflection, we need to rename or figure out whether they are still needed

Another one is that (at least for me) some target frameworks are not recognized:
image

Cheesebaron:
This still fails due to MvxLog, MvxTrace, Log, MvxTraceLevel missing all over the place.

martijn00:
Yeah and the code is not up to date with develop anymore. I tried to merge the newer code in, but i was a nightmare.

NetStandard/MvvmCross/MvvmCross.csproj
```diff
@@ -0,0 +1,83 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net461;uap10.0;Xamarin.iOS10;Xamarin.Mac20;Xamarin.TVOS10;Xamarin.WatchOS10;MonoAndroid80;netcoreapp2.0</TargetFrameworks>
```
nickrandolph: Why are we targeting netstandard2.0 - this is way to prohibitive for a general purpose library as it restricts which target platforms it can be used by. For example UWP can only use netstandard2.0 if it has a min version of the Fall Creators Update. 
Recommendation is to select a lower netstandard version number (eg 1.0 which is the equivalent of the PCL that mvvmcross was) as this will result in the lowest number of breaking changes for developers wanting to upgrade

Cheesebaron: General recommendation by who, for what purpose? Why should we care about dead targets?

nickrandolph: Lol sorry by recommendation I meant that I was suggesting it would be better to target a lower net standard version.
Targeting net standard 2 will be break nearly every existing mvvmcross project, requiring them to immediately retarget their libraries to net standard 2.0 (a major pita if they’re still using PCLs). Targetting netstandard 1.0, existing projects, even those using PCL, will mean that they are able to upgrade to latest mvvmcross without having to switch their projects to net standard.
With regards to UWP it is still too early to disregard versions earlier than the Fall Creators Update. I don’t have exact numbers but setting min version to FCU (required for net standard 2.0 support) would immediately cut out a significant percentage of users which isnt acceptable imho.

Plac3hold3r: Just to add some updated figures for UWP as the figures @nickrandolph quoted where from the [November report](http://reports.adduplex.com/reports/2017-11/) and Windows seems to be moving there user base to newer versions quite quickly. For the [December report](http://reports.adduplex.com/reports/2017-12/) from AdDuplex estimate

OS version | % Market Share
:--------:|:------:
Fall Creators Update | 53.6%
Creators Update | 33.7%
Anniversary Update | 10.5%
November Update | 1.3%
RTM | 0.5%
Redstone 4 (Preview) | 0.4%

--------------------
If I can add my 2 cents, I see three potential options

___Option 1 - .NET Standard 2.0___

Advantage | Disadvantage
:--------:|:------:
Largest .NET APIs available for use in Mvx | Will not work with currently used PCL profiles
_ | Only support UWP 16299 (Fall Creators Update)

___Option 2 - .NET Standard 1.4___

Advantage | Disadvantage
:--------:|:------:
Larger than current .NET APIs available for use in Mvx | Fewer potential .NET APIs available for use in Mvx (Compared to .NET Standard 2.0)
Supported by all versions of UWP | Will not work with currently used  PCL profiles

___Option 3 - .NET Standard 1.0 - 1.1___

Advantage | Disadvantage
:--------:|:------:
Slightly larger than current .NET APIs available for use in Mvx*<sup>1</sup> | Fewer potential .NET APIs available for use in Mvx (Compared to .NET Standard 2.0)
Supported by all versions of UWP |
Will work with currently used  PCL profiles

*<sup>1</sup> When using .NET Standard 1.1

Personally, I like the idea of targeting .NET Standard 2.0 and giving more potential to the framework (However, I'm not sure want MvvmCross would need to made use of inside the .NET Standard 2.0 API spec, perhaps ValueTuples?). From a biased view I have also already updated most of the projects I work on to use .NET Standard and consuming a 2.0 library will be effortless. 

However, I agree with @nickrandolph point that targeting a higher version of the Standard like 2.0 will increase the work required to update to MvvmCross 6 and potential limit usage. Although, I'm not sure if there is a scenario where a user of MvvmCross could not update there PCL to a .NET Standard library. Any long term project would more than likely need move over at some point to take advantage of other newer libraries updates. My experience with moving over to a .NET Standard class library was fairly painless.

For the UWP scenario I can not really comment as I do not have any apps with dependencies on it. The Windows user base as a whole does seem to be updating fairly quickly, but I'm sure this will slow down and there will be plenty of Windows user running versions of Windows that would not support a MvvmCross 6 app targeting .NET Standard 2.0. I would therefore suggest perhaps looking at dropping the target to .NET Standard 1.4 as this is the highest version to guarantee older versions of UWP support. At least until mid 2018 or so to give more time for Windows users to update. This of cause would depend on when MvvmCross 6 would get released. Perhaps by the time MvvmCross 6 gets released UWP usage on newer versions would be high enough that this is a small issue and we may as well continue targeting .NET Standard 2.0 for MvvmCross 6. 

That being said, if we do drop down to support .NET Standard 1.4 for UWP users, would there be any reason why we could not drop down further to .NET Standard 1.1 and then support PCL users as well?

nickrandolph: I'd reiterate that I would start by targetting 1.0/1.1 (whichever is compatible with the PCL profile currently used by Mvvmcross). Over time this can be increased as more features are required and based on support for net standard versions by different target platforms.

nickrandolph:
**I want to lodge a severe concern about the size of this PR. My suggestion is that it is disbanded.**
Currently there is way to much instability with Mvvmcross, a PR like this will result in less stability in the short term which this framework cannot afford right now.
Instead of attempting to do too much, let's start by just switching from PCL to .NET standard - this would be a simple PR and shouldn't break any consumers of mvvmcross.

martijn00:
@nickrandolph you are right, this one is too big. I'm already working on a new one to start off fresh. Other than that i think we should probably go for net standard 2.0. It seems lots of other libs start targeting that one, and in the longer term this makes it easier for us the maintain Mvx.

Another option is to target both 1.1/1.4 and 2.0. This would be possible with multi-target.

nickrandolph:
@martijn00 I think multi-targetting 1.1 and 2.0 makes sense. 
I understand the desire to move to net standard but as I suggested earlier I would go to 1.x as a first pass as it won't break anything - get this PR done first so we can move onto other things (separate PRs):
We can add multi-targetting to simplify the project system for mvx
Do upgrades to various libraries 
Fix the async void issue that's a plague in the heart of  the navigation framework - this is **urgent** and shouldn't be pushed back any more
net standard 2.0 is a low priority and I'm still not sure what value it's going to offer for the core of mvx over and above multi-targetting

PR #2387, opened 2017-11-14
Fixing Master-Detail Forms implementation

nickrandolph:
Bug fix

Master Details didn't really work using attributes

Master Details works - 2 scenarios
Define only Master and Detail pages/view models - automatically generates the MasterDetailsPage
Define Root page/view model where the page inherits from MasterDetailsPage

No

Playground sample now works for Master-Detail

- [ ] All projects build


MvvmCross-Forms/MvvmCross.Forms/Views/MvxFormsPagePresenter.cs
```diff
@@ -199,6 +200,42 @@ public virtual bool CloseCarouselPage(IMvxViewModel viewModel, MvxCarouselPagePr
             }
         }
 
+        public virtual MvxBasePresentationAttribute CreatePresentationAttribute(Type viewModelType, Type viewType)
+        {
+            if (viewType.GetTypeInfo().IsSubclassOf(typeof(ContentPage)))
```
martijn00: We might be able to use this directly again on .NET Standard.

MvvmCross-Forms/MvvmCross.Forms/Views/MvxFormsPagePresenter.cs
```diff
@@ -552,6 +607,31 @@ public virtual bool ClosePage(Page rootPage, Page page, MvxPagePresentationAttri
                 return rootPage as TPage;
         }
 
+        public virtual void ReplaceMasterDetailRoot(Page existingMasterDetailPage, Page newRootPage, MvxMasterDetailPagePresentationAttribute masterDetailAttribute)
```
martijn00: I don't understand what this tries to solve. Can you explain the thought behind it?

nickrandolph:
@martijn00, @Cheesebaron hey let me know thoughts on this PR and what needs to be changed/done to get this approved.

MvvmCross-Forms/MvvmCross.Forms.Droid/Platform/MvxFormsAndroidSetup.cs
```diff
@@ -1,28 +1,28 @@
...
-using MvvmCross.Forms.Views;
-using MvvmCross.Forms.Droid.Bindings;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
 
 namespace MvvmCross.Forms.Droid.Platform
 {
-    public abstract class MvxFormsAndroidSetup : MvxAndroidSetup
+    public abstract class MvxFormsAndroidSetup<TForms> : MvxAndroidSetup
```
martijn00: Why does this need to add a generic? I don't see any value?

MvvmCross-Forms/MvvmCross.Forms.Droid/MvvmCross.Forms.Droid.csproj
```diff
@@ -96,16 +96,16 @@
...
-      <HintPath>..\..\packages\Xamarin.Forms.2.4.0.74863\lib\MonoAndroid10\Xamarin.Forms.Core.dll</HintPath>
+      <HintPath>..\..\packages\Xamarin.Forms.2.5.0.91635\lib\MonoAndroid10\Xamarin.Forms.Core.dll</HintPath>
     </Reference>
     <Reference Include="Xamarin.Forms.Platform, Version=1.0.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Xamarin.Forms.2.4.0.74863\lib\MonoAndroid10\Xamarin.Forms.Platform.dll</HintPath>
+      <HintPath>..\..\packages\Xamarin.Forms.2.5.0.91635\lib\MonoAndroid10\Xamarin.Forms.Platform.dll</HintPath>
     </Reference>
     <Reference Include="Xamarin.Forms.Platform.Android, Version=2.0.0.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Xamarin.Forms.2.4.0.74863\lib\MonoAndroid10\Xamarin.Forms.Platform.Android.dll</HintPath>
+      <HintPath>..\..\packages\Xamarin.Forms.2.5.0.91635\lib\MonoAndroid10\Xamarin.Forms.Platform.Android.dll</HintPath>
```
martijn00: We should update the nuspec as well if this changes to 2.5

MvvmCross-Forms/MvvmCross.Forms.Droid/Platform/MvxFormsAndroidSetup.cs
```diff
@@ -51,13 +53,16 @@ public MvxFormsApplication FormsApplication
...
-                    _formsApplication = CreateFormsApplication();
+                {
+                    _formsApplication = _formsApplication ?? CreateFormsApplication();
+                }
                 return _formsApplication;
             }
         }
 
-        protected virtual MvxFormsApplication CreateFormsApplication() => new MvxFormsApplication();
+        protected virtual TForms CreateFormsApplication() => new TForms();
```
martijn00: Maybe we should make this abstract if it gives bugs?

MvvmCross-Forms/MvvmCross.Forms/Views/MvxFormsPagePresenter.cs
```diff
@@ -251,18 +298,27 @@ public virtual bool CloseContentPage(IMvxViewModel viewModel, MvxContentPagePres
...
                     masterDetailHost = new MvxMasterDetailPage();
-                    if (attribute.Position == MasterDetailPosition.Master)
-                        masterDetailHost.Master = page;
-                    else
-                        masterDetailHost.Master = new MvxContentPage() { Title = !string.IsNullOrEmpty(attribute.Title) ? attribute.Title : nameof(MvxMasterDetailPage) };
-                    if(attribute.Position == MasterDetailPosition.Detail)
-                        masterDetailHost.Detail = page;
-                    else
-                        masterDetailHost.Detail = new MvxNavigationPage(new MvxContentPage() { Title = !string.IsNullOrEmpty(attribute.Title) ? attribute.Title : nameof(MvxMasterDetailPage) });
+                    //if (attribute.Position == MasterDetailPosition.Master)
```
martijn00: Can you remove any out commented code?

MvvmCross-Forms/MvvmCross.Forms/Views/MvxFormsPagePresenter.cs
```diff
@@ -230,14 +277,14 @@ public virtual bool CloseContentPage(IMvxViewModel viewModel, MvxContentPagePres
...
             var page = CreatePage(view, request, attribute);
 
-            if(attribute.Position == MasterDetailPosition.Root)
+            if (attribute.Position == MasterDetailPosition.Root)
             {
                 if (page is MasterDetailPage masterDetailRoot)
                 {
                     if (masterDetailRoot.Master == null)
-                        masterDetailRoot.Master = new MvxContentPage() { Title = !string.IsNullOrEmpty(attribute.Title) ? attribute.Title : nameof(MvxMasterDetailPage)  };
+                        masterDetailRoot.Master = new ContentPage() { Title = !string.IsNullOrEmpty(attribute.Title) ? attribute.Title : nameof(MvxMasterDetailPage) };
```
martijn00: Why change from using `MvxContentPage` to `ContentPage`?

MvvmCross/Core/Core/Platform/LogProviders/NullLogProvider.cs
```diff
@@ -0,0 +1,7 @@
+namespace MvvmCross.Core.Platform.LogProviders
+{
+    internal sealed class NullLogProvider : MvxBaseLogProvider
```
martijn00: This should be removed, and instead set the logger to None.

MvvmCross-Forms/MvvmCross.Forms/Views/MvxFormsPagePresenter.cs
```diff
@@ -441,101 +528,208 @@ protected virtual void CloseModalStack(IReadOnlyList<Page> modals)
...
+        //        }
+        //    }
+
+        //    PushOrReplacePage(rootPage, page, attribute);
+        //}
+
         public virtual void PushOrReplacePage(Page rootPage, Page page, MvxPagePresentationAttribute attribute)
         {
-            if (attribute.NoHistory)
+            // Make sure we always have a rootPage
```
martijn00: I'm a bit unsure if this behaviour works in all situations. Can you confirm?

martijn00:
There is definitely some breaking changes in here in terms of behaviour. When testing the playground the modal navigation is different, and the master detail views don't show up with a toolbar with hamburger icon anymore. Also when navigating and coming back the buttons stay disabled. I didn't notice these things before so i think there might be some bugs in here.

MvvmCross-Forms/MvvmCross.Forms.Droid/Platform/MvxFormsAndroidSetup.cs
nickrandolph: I'd suggest I revert to non-generic and that we have a separate generic version of the setup class.

MvvmCross-Forms/MvvmCross.Forms.Droid/Platform/MvxFormsAndroidSetup.cs
nickrandolph: As per my other comment, I think we should revert this change. In the generic setup class the CreateFormsApplication can return a new instance of the generic TForms class

MvvmCross-Forms/MvvmCross.Forms/Views/MvxFormsPagePresenter.cs
nickrandolph: We were seeing issues where the View Model was being set on the MvxContentPage (passed down from parent to the Master page) which resulted in the OnAppearing method being called multiple times on the master detail root view model

MvvmCross-Forms/MvvmCross.Forms/Views/MvxFormsPagePresenter.cs
nickrandolph: Will do

MvvmCross/Core/Core/Platform/LogProviders/NullLogProvider.cs
nickrandolph: Without this class, the None option for logging will break any code which requests ILogProvider as an injected value. Do you have another alternative to this?

nickrandolph:
A couple of questions/comments:
- Can you be specific on the changes to modal navigation - I tried to make it more consistent as there were cases where it wouldn't show a modal, or when the modal was closed it didn't go back to the master detail
- I'll fix the issue with the icon
- The disabled buttons are actually due to a Delay that's in the Initialize method in one of the view models. Should this prevent navigation until this has completed?

PR #2399, opened 2017-11-22
Fix logging in netstandard

willsb:
This turns `IMvxLog` into the de facto logging mechanism for MvvmCross for version 6.0

Trace is deprecated, but exists
Some issues caused by heavy usage of reflection

netstandard allows us to have a reliable console based logging mechanism in every platform, meaning we can ditch 

Possibly! Thorough reviews are appreciated.

Using the test apps once everything is ported to see if nothing goes boom boom

- [ ] All projects build<sup>1</sup>
- [ ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contribute/mvvmcross-docs-style-guide))<sup>2</sup>
- [X] Rebased onto current ~develop~ netstandard

// TODO:
- [ ] Remove MvxTrace from all the Platforms targets
- [ ] Improve the registration mechanism now that `IMvxLogProvider` is mandatory

<sup>1</sup> As it is right now, only the core project builds on my mac, as support for multi targeting is bad on macOS and the plugins are not building properly
<sup>2</sup> Note to self: Include proper instructions on how to override the default MvxLog behavior.


NetStandard/MvvmCross/Platform/Logging/MvxLog.cs
```diff
@@ -1,12 +1,14 @@
...
+using MvvmCross.Platform.Logging.LogProviders;
 
-namespace MvvmCross.Core.Platform.LogProviders
+namespace MvvmCross.Platform.Logging
 {
     internal class MvxLog : IMvxLog
     {
         internal const string FailedToGenerateLogMessage = "Failed to generate log message";
 
+        internal static IMvxLog InternalLogInstance { get; set; }
```
martijn00: I would probably just call this `Instance` instead of `InternalLogInstance`

NetStandard/MvvmCross/Binding/Binders/MvxNamedInstanceRegistryFiller.cs
```diff
@@ -88,7 +87,7 @@ where type.IsConventional()
                 try
                 {
                     var converter = Activator.CreateInstance(pair.Type) as T;
-                    MvxSingleton<IMvxLog>.Instance.Trace("Registering value converter {0}:{1}", pair.Name, pair.Type.Name);
```
softlion: Why are you removing the usage of MvxSingleton ? Is MvxLog not in MvxCore ?

PR #2400, opened 2017-11-22
Fix Droid Activities when navigating for result

### LINKED ISSUE #2384: Navigation with Results and Configuration Change cause premature delivery of null result

1.
Provided that you have ViewModel A and a ViewModel B. Implement a navigation pattern so that A calls B and waits for B to deliver a result, i.e. B is derived from MvxViewModelResult
On Android, assume that both are Activities (I have not tested otherwise).

2.
Let A call B and wait for a result, when on Activity for B, turn the screen. 

3.
You'll notice that the TaskCompletionSource gets triggered by MvxAppCompatActivity::OnDestroy -> MvxViewModelResult::OnViewDestroy. In Fact the TCS is canceled so that A gets a null result. But in reality, Android is just re-creating the Activity for B with a different orientation.

The TCS should only be canceled if moving away from ViewModel B and not when the Activity of B is destroyed and recreated

### Actual behavior :bug:
Since ViewModel B is not semantically moved away from, but just the View for B is destroyed and recreated it is a wrong behavior leading to haveing a View for B and an intact ViewModel B but A won't get the real result as the TCS has already been canceled by the orientation change.

In android documentation there is noted on how to distinguish between closing a view with finish and the view being destroyed for other reasons:
https://developer.android.com/reference/android/app/Activity.html#onDestroy()

So maybe one approach could be to follow that guide and have some additional lifetime semantic to the viewmodel which understands the difference between the "view has been destroyed" and the "the user has backed out from / closed the view".

### Configuration :wrench:

**Version:** 5.3

**Platform:** 
- [ ] :iphone: iOS
- [x ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

nmilcoff:
Bug fix.

When navigating for result to a new Activity, if the device is rotated, there is a premature call to ViewModel.ViewDestroy.

Above is fixed. Solution is to call ViewModel.ViewDestroy only if the Activity is being finished.

No.

Playground TestProject: Make TabsRootViewModel derive from `MvxViewModelResult<bool>` and watch for the navigation result from RootViewModel. When using this branch, View rotation will work as expected. 

Fixes #2384.

- [X] All projects build

### DIFF
--- a/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.Fragment/MvxFragmentActivity.cs
+++ b/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.Fragment/MvxFragmentActivity.cs
@@ -84,35 +84,37 @@ namespace MvvmCross.Droid.Support.V4
             ViewModel?.ViewCreated();
         }
 
-		protected override void OnDestroy()
-		{
-			base.OnDestroy();
-			ViewModel?.ViewDestroy();
-		}
+        protected override void OnDestroy()
+        {
+            base.OnDestroy();
 
-		protected override void OnStart()
-		{
-			base.OnStart();
-			ViewModel?.ViewAppearing();
-		}
+            if (IsFinishing)
+                ViewModel?.ViewDestroy();
+        }
 
-		protected override void OnResume()
-		{
-			base.OnResume();
-			ViewModel?.ViewAppeared();
-		}
+        protected override void OnStart()
+        {
+            base.OnStart();
+            ViewModel?.ViewAppearing();
+        }
 
-		protected override void OnPause()
-		{
-			base.OnPause();
-			ViewModel?.ViewDisappearing();
-		}
+        protected override void OnResume()
+        {
+            base.OnResume();
+            ViewModel?.ViewAppeared();
+        }
 
-		protected override void OnStop()
-		{
-			base.OnStop();
-			ViewModel?.ViewDisappeared();
-		}
+        protected override void OnPause()
+        {
+            base.OnPause();
+            ViewModel?.ViewDisappearing();
+        }
+
+        protected override void OnStop()
+        {
+            base.OnStop();
+            ViewModel?.ViewDisappeared();
+        }
     }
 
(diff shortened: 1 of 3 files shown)


ElteHupkes:
Just when I was getting used to tracking my view state through my view model...

`ViewDestroyed` is used for other things than activity navigation guys! Event subscriptions, for one thing. In my opinion this breaks consistency in favor of supporting a navigation pattern that is fundamentally broken on Android anyway.

nmilcoff:
ViewDestroyed was being called more than once, more specifically each time
the device is rotated. Can you explain why is this an issue for you?

ElteHupkes:
Sorry if I sounded a bit harsh - HAXM keeps BSOD'ing my machine every time I boot the emulator which isn't so good for my mood ;).

I regularly test with "don't keep activities" to see what happens if Android tombstones (a part of) my app. Concrete example of a problem: I use `Acr.UserDialogs` to display a message when a connection error occurs. When you do this without a non-destroyed activity though, it outright crashes the app because obviously there is no activity available to display the dialog. Say the app is minimized, which kills all activities without finishing them. The services are still active, a network request fails and I want to display a dialog - that's a crash if I can't detect if that view has been destroyed. I have a couple more events like these that only make sense when an actual view is around, and I simply ignore or queue them in that case. 

In any case I think `ViewDestroy` should just be called when activities are destroyed - and they *are* when orientation changes, they're just recreated right after - no matter what the cause. I'd imagine there are use cases I can't even think of. This change removes that consistency to support a navigation pattern that, in my opinion, you shouldn't use anyway if you're serious about supporting Android (I can elaborate on that if you like).

nmilcoff:
No problem about that really, it didn't sound bad 🙂.

I understand your concern, I'll make a new PR trying to address this using an alternative approach.

ElteHupkes:
Awesome! That seems like a better solution.

Issue #2108, opened 2017-08-09
Update tvOS presenters to the new iOS presenter


martijn00:
Reference https://github.com/MvvmCross/MvvmCross/issues/1724

- [x] Update tvOS View Presenter
- [X] Support Root ViewControllers
- [X] Support Child ViewControllers
- [X] Support Tab ViewControllers
- [X] Support Modal ViewControllers
- [x] Support SplitView
- [ ] Create documentation

PR #2414, opened 2017-11-28
tvOS presentation update to match features in other platforms

biozal:
This change includes a new Presenter for tvOS along with changes to the existing Presentation framework to bring it up to speed and feature match the current release of iOS.

The current behavior doesn't support attributes or the new navigation life cycle events or using attributes for things like tabs and modal dialogs.

The new presentation  object(s) brings tvOS inline with the same features as iOS - as prohibited by what the tvOS supports including a new tab presentation support and modal using attributes.

YES - the old presentation objects were removed to make way for the new ones.  This means that if you have a custom presentation class you wrote for tvOS support it probably needs reworking to work with the new stuff - and if you wrote a custom presentation class to work with tabs you probably don't need it anymore.  Also the button event for binding was set to Touch Up Inside and based on Xamarin's documentation - this is wrong so I switched it to the Xamarin recommended event:

https://developer.xamarin.com/guides/ios/tvos/user-interface/buttons/

See section:
NOTE: While it is possible to assign actions such as TouchUpInside to a UIButton, it will never be called because Apple TV doesn't have a touch screen or support touch events. You should always use events such as AllEvents or PrimaryActionTriggered.

I switched the binding event to go off when PrimaryActionTriggered is fired and this has allowed for button binding to Command's in the ViewModel without mapping using the For method.  I haven't updated the documentation yet on the website for this - and the startup nuget package's AppDelegate directions need to be updated.  I'll have to figure out where they are and put in a different pull request to update those.

I included a new Playground tvOS app that has the same features as the iOS playground app.  Each screen was tested to make sure they behave like the iOS counter part.

Not sure - asked the team today on Slack about tvOS support and they told me it needed worked on, so I worked on it.

- [X] All projects build


nmilcoff:
PR for: #2108

biozal:
One more thing - this does not support split view controller yet.  I'll patch that in next week.

Cheesebaron:
@biozal Can you please take a look at some of the CodeFactor issues. Just ignore any issues in the PlayGround project as this has been added to the ignore pile.

biozal:
@cheesebaron I’ll check tonight but 80% of these are because the iOS  and the tvOS objects are the same name - so 90% of the code is going to look the same.  Suggestions?

nmilcoff:
Awesome job @biozal! Please enable editions to your PR so I can commit some cleanups

biozal:
@nmilcoff I should have edits enabled for all team members.  I'm back at the hotel know and I'm looking at things.

nmilcoff:
@biozal Please check that your develop branch doesn't have any special policies / permissions, I still can't push anything into it.

biozal:
Here is the setting on the side of this PR.  I see nothing special in my permissions.  Weird...  I'll check my fork permissions.

screen shot 2017-11-28 at 10 49 46 pm

PR #2441, opened 2017-12-05
MvxTableViewSource.CollectionChangedOnCollectionChanged: Allow execution from a worker thread

### LINKED ISSUE #2360: MvxTableViewSource CollectionChangedOnCollectionChanged can be executed on a worker thread
When using an ObservableCollection with a MvxTableViewSource, the collection changed event is subscribed by MvxTableViewSource but its handler does not check if it is run on the main thread.

Using the Task based binder, the viewmodel should fire and forget events on any thread.

### Actual behavior :bug:
If the event is received on a worker thread, a UIKit consistency exception occurs.

### Fix

In MvxTableViewSource change the CollectionChangedOnCollectionChanged handler:

        protected override void CollectionChangedOnCollectionChanged(object sender, NotifyCollectionChangedEventArgs args)
        {
            void Action()
            {
                if (!UseAnimations)
                    ReloadTableData();
                else
                {
                    if (TryDoAnimatedChange(args))
                        return;
                    ReloadTableData();
                }
            }

            if (NSThread.IsMain)
                Action();
            else
                InvokeOnMainThread(Action);
        }

### Configuration :wrench:

**Version:** 5.4

**Platform:** 
- [X] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

nmilcoff:
Code improvement.

If CollectionChangedOnCollectionChanged is called from a worker thread, the app will 💥 .

CollectionChangedOnCollectionChanged can be called from a worker thread.

No.

/

Fixes #2360 

- [X] All projects build

### DIFF
--- a/MvvmCross/Binding/Mac/Views/MvxTableViewSource.cs
+++ b/MvvmCross/Binding/Mac/Views/MvxTableViewSource.cs
@@ -28,12 +28,12 @@ namespace MvvmCross.Binding.Mac.Views
 
         public MvxTableViewSource(NSTableView tableView) : base()
         {
-            this._tableView = tableView;
+            _tableView = tableView;
         }
 
         public override nint GetRowCount(NSTableView tableView)
         {
-            return this.ItemsSource.Count();
+            return ItemsSource.Count();
         }
 
         [MvxSetToNullAfterBinding]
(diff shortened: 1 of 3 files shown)


MvvmCross/Binding/Mac/Views/MvxTableViewSource.cs
```diff
@@ -89,24 +88,41 @@ private NSView GetOrCreateViewFor(NSTableView tableView, NSTableColumn tableColu
...
+                        return;
+
+                    ReloadTableData();
+                }
+            };
+
+            if (NSThread.IsMain)
+                action();
+            else
+                InvokeOnMainThread(action);
```
ElteHupkes: FYI, I recently found out the hard way that this might result in having the table display the wrong items. I'll give you an example.

Let's say you have a background service returning items as they become available and inserting them into the table (no surprise, we have such a service in our app). The insertion is sorted, so items may be added anywhere in the list. Now let's assume that a first item comes in, inserted at position `0`. An `Add at position 0` collection change event is scheduled. Another item arrives shortly after, which will also insert at position `0` because it precedes the previous item in the ordering. Another `Add at position 0` event. Because the collection changed events are executed on the main thread (or their side effects on the table are, it makes little difference in this case), both events may be executed after the items source has already been updated (this turned out to be very likely in our case). Two add events are successfully executed, but both will read their data context from position `0` in the items source. The table will have 2 items, but both are the same item. It's not a problem for complete table reloads of course, but those may be undesirable for large data lists.

I thought about this for a while and don't really see an easy / intuitive way around this that doesn't require forcing access to the items source to occur on the main thread. Personally I have implemented a custom observable collection which maintains both an "up to date" list and a list as it is currently viewed by the UI thread. Changes to the former are immediate, whereas changes to the latter happen on the UI thread and are synchronized with the collection change events. This way I can make immediate changes on any thread which will eventually be efficiently synchronized with the data table.

nmilcoff: Please open a new issue linking this PR instead of commenting here (it's really difficult to be up to date with GitHub notifications and you can easily miss anything, specially if it's a comment on an already merged PR).

Looks like you've a very special situation there, and I think you solved it in a very efficient way 👌. But still, I think this PR was an improvement over the previous behavior (it just made apps crash in the past).

If you think your solution may fit into the framework, I would love to see it in a PR!

ElteHupkes: @nmilcoff Yeah you're absolutely right, I just saw this pop up on the go and had to comment it somewhere quickly. I'll see if I can create something; the thing is it's kind of a complex problem and unless you understand why it's needed the solution doesn't make an awful lot of sense. So there will have to be some explaining too ;).

PR #2454, opened 2017-12-11
Fixing launch when splash screen isn't used

nickrandolph:
Bug fix

Android with Xamarin Forms doesn't load when there's no splash screen.

Android with Xamarin Forms launches correctly

Yes, IMvxStart.Start is now async

Run the Playground.Forms.Droid

- [ ] All projects build

### DIFF
--- a/MvvmCross-Forms/MvvmCross.Forms.Droid/Views/MvxFormsAppCompatActivity.cs
+++ b/MvvmCross-Forms/MvvmCross.Forms.Droid/Views/MvxFormsAppCompatActivity.cs
@@ -26,7 +26,7 @@ namespace MvvmCross.Forms.Droid.Views
         {
             BindingContext = new MvxAndroidBindingContext(this, this);
             this.AddEventListeners();
-            _resourceAssembly = Assembly.GetCallingAssembly();
+            _resourceAssembly = this.GetType().Assembly;
         }
 
         public object DataContext
--- a/MvvmCross/Core/Core/ViewModels/IMvxAppStart.cs
+++ b/MvvmCross/Core/Core/ViewModels/IMvxAppStart.cs
@@ -1,14 +1,16 @@
-﻿// IMvxAppStart.cs
+// IMvxAppStart.cs
 
 // MvvmCross is licensed using Microsoft Public License (Ms-PL)
 // Contributions and inspirations noted in readme.md and license.txt
 //
 // Project Lead - Stuart Lodge, @slodge, me@slodge.com
 
+using System.Threading.Tasks;
+
 namespace MvvmCross.Core.ViewModels
 {
     public interface IMvxAppStart
     {
-        void Start(object hint = null);
+        Task Start(object hint = null);
     }
 }
--- a/MvvmCross/Core/Core/ViewModels/MvxAppStart.cs
+++ b/MvvmCross/Core/Core/ViewModels/MvxAppStart.cs
@@ -1,4 +1,4 @@
-﻿// MvxAppStart.cs
+// MvxAppStart.cs
 
 // MvvmCross is licensed using Microsoft Public License (Ms-PL)
 // Contributions and inspirations noted in readme.md and license.txt
@@ -6,6 +6,7 @@
 // Project Lead - Stuart Lodge, @slodge, me@slodge.com
 
 using System;
+using System.Threading.Tasks;
 using MvvmCross.Platform.Platform;
 
 namespace MvvmCross.Core.ViewModels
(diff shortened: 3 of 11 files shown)


MvvmCross-Forms/MvvmCross.Forms.Droid/Views/MvxFormsAppCompatActivity.cs
```diff
@@ -26,7 +26,7 @@ protected MvxFormsAppCompatActivity()
         {
             BindingContext = new MvxAndroidBindingContext(this, this);
             this.AddEventListeners();
-            _resourceAssembly = Assembly.GetCallingAssembly();
+            _resourceAssembly = this.GetType().Assembly;
```
martijn00: I don't think this is going to work. We need to get the actual assembly of the app project. @johnnywebb thoughts?

MvvmCross/Core/Core/ViewModels/IMvxAppStart.cs
```diff
@@ -1,14 +1,16 @@
...
 // Project Lead - Stuart Lodge, @slodge, me@slodge.com
 
+using System.Threading.Tasks;
+
 namespace MvvmCross.Core.ViewModels
 {
     public interface IMvxAppStart
     {
-        void Start(object hint = null);
+        Task Start(object hint = null);
```
martijn00: We should not change this at this time.

MvvmCross/Core/Core/ViewModels/MvxAppStart.cs
```diff
@@ -15,7 +16,7 @@ public class MvxAppStart<TViewModel>
         : MvxNavigatingObject, IMvxAppStart
         where TViewModel : IMvxViewModel
     {
-        public void Start(object hint = null)
+        public async Task Start(object hint = null)
```
martijn00: This won't be async this way, can you revert it?

MvvmCross/Droid/Droid/Platform/MvxAndroidSetupSingleton.cs
```diff
@@ -43,6 +44,10 @@ public virtual void EnsureInitialized()
                 {
                     IsInitialisedTaskCompletionSource = new TaskCompletionSource<bool>();
                     _setup.Initialize();
+
+                    var start = Mvx.Resolve<IMvxAppStart>();
+                    start.Start().Wait();
```
martijn00: Doing a Wait is bad practice.

martijn00:
Fixed in https://github.com/MvvmCross/MvvmCross/pull/2457

MvvmCross-Forms/MvvmCross.Forms.Droid/Views/MvxFormsAppCompatActivity.cs
johnnywebbbankmobile: @martijn00 you are correct. The objective here is to obtain the project assembly containing the android resources. In most cases, the app assembly contains a MainActivity derived from MvxFormsAppCompatActivity; hence, why `Assembly.GetCallingAssembly` is use in the constructor. [More info can be found here](https://github.com/MvvmCross/MvvmCross/blob/8146d2ca600e043d3e1dfbe0ef096dbf26332973/docs/_documentation/platform/forms/xamarin-forms-customization.md)

`this.GetType().Assembly` will return the `MvvmCross.Forms.Droid` assembly and the droid resources, such as images, will fail to load.

nickrandolph: @johnnywebbbankmobile I don't think that's true.... this.GetType().Assembly will get the Type for the sub-type (ie the class that inherits from MvxFormsAppCompatActivity, not MvFormsAppCompatActivity), so will pick up the correct assembly

Assembly.GetCallingAssembly will pick up the assembly of the calling method. This fails because if you use MvxFormsAppCompatActivity<TViewModel> because it picks up the ctor for that class which is in MvvmCross.Forms.Droid.

I'm not fussed about this particular change as it didn't turn out to be the cause of the Droid XF not rendering but I don't believe the current code is correct and should be fixed

johnnywebbbankmobile: ah yeah you are right @nickrandolph . 👍 good catch, I'll fix it.

johnnywebb: @nickrandolph good catch. I fixed here: https://github.com/MvvmCross/MvvmCross/pull/2461

PR #2461, opened 2017-12-12
mvxforms/droid-resources

johnnywebb:
Code improvement

MvxForms Activities currently leverages GetCallingAssembly() which must be called from the constructor... (not very flexible, and is called regardless of override)

updated logic to use this.GetType().Assembly which is only called within GetResourceAssembly().

No.

1. Launch the forms playground
1. View image on root

- [x] All projects build


MvvmCross-Forms/MvvmCross.Forms.Droid/Views/MvxFormsAppCompatActivity.cs
```diff
@@ -113,7 +111,7 @@ public virtual void InitializeForms(Bundle bundle)
 
         protected virtual Assembly GetResourceAssembly()
         {
-            return _resourceAssembly;
+            return this.GetType().Assembly;
```
Cheesebaron: This will get the assembly of MvvmCross.Forms.Droid and not the Assembly containing the Android resources. The reason it is virtual is so you can modify the behavior if needed. I don't think this is a valid change.

MvvmCross-Forms/MvvmCross.Forms.Droid/Views/MvxFormsApplicationActivity.cs
```diff
@@ -111,7 +109,7 @@ public virtual void InitializeForms(Bundle bundle)
 
         protected virtual Assembly GetResourceAssembly()
         {
-            return _resourceAssembly;
+            return this.GetType().Assembly;
```
Cheesebaron: Same as the comment for MvxFormsAppCompatActivity

MvvmCross-Forms/MvvmCross.Forms.Droid/Views/MvxFormsAppCompatActivity.cs
nickrandolph: @Cheesebaron the only time the proposed change will get the MvvmCross.Forms.Droid assembly is if this class isn't sub-classed, which I'm not sure is even possible for an Android XF app.
The previous code breaks as it tries to retrieve the assembly of the calling method; if the calling method is in a different library (for example the ctor of the generic version of MvxFormsAppCompatActivity which is in MvvmCross.Forms.Droid) it'll pick the wrong assembly.

johnnywebb: @Cheesebaron I thought the same thing. I'm still confused how this is working... I tested the Playground.Forms.Droid with this several times and it is returning the `Playground.Forms.Droid` assembly as expected.

nickrandolph: @johnnywebb I think it might be working because the generic MvxFormsAppCompatActivity doesn't have it's own ctor, which means the calling assembly will be the assembly where the sub-classed activity lives. However, this isn't a great assumption to make. "this.GetType" should always return the sub-classed activity, which should be the correct assembly unless I've missed something

johnnywebb: shouldn't these MvxForms activity classes be abstract anyways? @Cheesebaron @nickrandolph

johnnywebb: Should MvxFormsAppCompatActivity and MvxFormsApplicationActivity be abstract? @Cheesebaron @nickrandolph

nickrandolph: I don't see why they can't be. I don't see a scenario where they'd be used without being sub-classed.

johnnywebb: @nickrandolph would you happen to have a branch reproducing the GetCallingAssembly error? That code seems to work for me as well...

nickrandolph: @johnnywebb you just need to run the playground.forms.droid sample and put a breakpoint at the point in the ctor of MvxFormsAppCompatActivity where it attempts to call  Assembly.GetCallingAssembly - this will return Mvvmcross.Forms.Droid instead of the Playground.Forms.Droid assembly

nickrandolph: @Cheesebaron are you still against making this change?

Cheesebaron: I guess it is fine if marked abstract.

PR #2477, opened 2017-12-19
MvxRecyclerViewAdapter: Support worker threads

### LINKED ISSUE #2144: MvxRecyclerAdapter NotifyDataSetChanged not called on mainthread
When an obserbable collection is changed on a worker thread, MvxRecyclerAdapter.NotifyDataSetChanged is called on the same worker thread instead of switching to the main thread.

Following the same philosophy that drove the introduction of the task based binder which eliminates the need to fire notifypropertychange events on the main thread, it would be better if the MvxRecyclerAdapter test if it is on the main thread and switch to it automatically if not before calling the NotifyXXXX native methods on the base android class, which requires the main thread.

Also, https://github.com/MvvmCross/MvvmCross/blob/master/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V17.Leanback/Listeners/MvxOnChildViewHolderSelectedListener.cs should use IMvxRecyclerAdapter instead of MvxRecyclerAdapter on line 21.

Also, there should not be any `new MvxRecyclerAdapter()` code in MvxHorizontalGridView.cs or MvxRecyclerView.cs, instead it should use iOC instantiation. It would be much easier to switch the implementation of MvxRecyclerAdapter.

### Configuration

**Version:** 5.1

**Platform:** 
- [ ] iOS
- [X] Android
- [ ] WPF
- [ ] UWP

nmilcoff:
Code improvement

When a user tries to update a bound ItemsSource collection from a worker thread, MvxRecyclerViewAdapter just 💥💥. 

MvxRecyclerViewAdapter handles collection changes from worker threads.

No.

Run TestProjects/Android-Support. `ExampleRecyclerViewModel` updates a collection from background.

Closes #2144

- [X] All projects build

### DIFF
--- a/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.RecyclerView/MvxRecyclerAdapter.cs
+++ b/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.RecyclerView/MvxRecyclerAdapter.cs
@@ -9,6 +9,8 @@ using System;
 using System.Collections;
 using System.Collections.Specialized;
 using System.Windows.Input;
+using Android.App;
+using Android.OS;
 using Android.Runtime;
 using Android.Views;
 using MvvmCross.Binding;
@@ -26,7 +28,7 @@ using Object = Java.Lang.Object;
 namespace MvvmCross.Droid.Support.V7.RecyclerView
 {
     [Register("mvvmcross.droid.support.v7.recyclerview.MvxRecyclerAdapter")]
-    public class MvxRecyclerAdapter 
+    public class MvxRecyclerAdapter
         : Android.Support.V7.Widget.RecyclerView.Adapter, IMvxRecyclerAdapter, IMvxRecyclerAdapterBindableHolder
     {
         private readonly IMvxAndroidBindingContext _bindingContext;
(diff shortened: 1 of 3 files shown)


nmilcoff:
@softlion can you maybe review this PR? 🙂

MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.RecyclerView/MvxRecyclerAdapter.cs
```diff
@@ -241,7 +243,15 @@ protected virtual void SetItemsSource(IEnumerable value)
 
         protected virtual void OnItemsSourceCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
         {
-            NotifyDataSetChanged(e);
+            if (Looper.MainLooper == Looper.MyLooper())
```
martijn00: Isn't there a Mvx method that checks the thread and runs it on main if necessary?

softlion:
I've stumbled upon other crashes in MvxRecyclerViewAdapter. Will post them.

MvvmCross-AndroidSupport/MvvmCross.Droid.Support.V7.RecyclerView/MvxRecyclerAdapter.cs
Cheesebaron: There is `MvxAndroidMainThreadDispatcher`. However this class can't inherit from `MvxMainThreadDispatchingObject`. So this is a good solution.

PR #2482, opened 2017-12-20
[WIP] Added support for nested fragments (ChildFragmentManager)

Qwin:
Feature

Current behavior only supports FragmentManager and SupportFragmentManager on 1 nested level.
It also doesn't support nested fragments in PageViewer. 

The new behavior adds support for nested fragments (multi level) (for now only MvxAppCompatPresenter supports it).
Adds correct tagging on MvxCachedFragmentPagerAdapter so that nested fragments can be supported in PageViewer. 

Not yet (depends if we want to remove 2 of the PageAdapters into making only 2 available) 

Problems with the Android API I am using (found this like 2 hours ago) : 
https://stackoverflow.com/questions/6102007/is-there-a-way-to-get-references-for-all-currently-active-fragments-in-an-activi
When we look at the link above it seems that FragmentManager.Fragments is maybe not supposed to be used as its hidden (well its not really hidden in Xamarin). But I am concerned of future updates it might break the implementation. (if they end up making it private field). Also the non-AppCompat version we would need to use reflection to get the field with mActive as said per post. So I am not sure if this is a good solution.  

Didnt have time to do the below yet.  I currently implemented it in my own project by just re-implementing the MvxCachedFragmentPagerAdapter and overwriting the MvxAppCompatPresenter. It works as far as I can see but to be added to MvvmCross it needs more work and maybe a different approach. 

- [ ] All projects build

### DIFF
--- a/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.Fragment/MvxCachingFragmentPagerAdapter.cs
+++ b/MvvmCross-AndroidSupport/MvvmCross.Droid.Support.Fragment/MvxCachingFragmentPagerAdapter.cs
@@ -122,7 +122,7 @@ namespace MvvmCross.Droid.Support.V4
             return fragment;
         }
 
-        private static string FragmentJavaName(Type fragmentType)
+        protected string FragmentJavaName(Type fragmentType)
         {
             return Class.FromType(fragmentType).Name;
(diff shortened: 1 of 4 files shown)


martijn00:
Can you do a rebase?

Issue #2408, opened 2017-11-27
MvxExpandableListView GroupClick binding replaces group expanding functionality


xHype:
### Steps to reproduce :scroll:

1. Bind 'GroupClick' on an MvxExpandableListView

2. MvxExpandableListView group doesn't expand any more, but the code inside the command works

### Expected behavior :thinking:
Additionally expand the group besides executing the code in the command

### Actual behavior :bug:
The code inside the command works, but the group doesn't get expanded

### Configuration :wrench:

**Version:** 5.4.2

**Platform:** 
- [x] :robot: Android
- [ ] :iphone: iOS
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

I am trying to do some additional stuff when a user taps on a group of an MvxExpandableListView, but apparently using the code below makes the group unable to expand a group anymore (if I delete the code below, the list works just fine; if I keep it, clicking the group won't expand it)

    public IMvxCommand SelectGroupCommand => new MvxCommand<DrinkViewModel>(SelectGroup);
    public void SelectGroup(DrinkViewModel drink)
    {

    }

My layout:

    <Mvx.MvxExpandableListView
    xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:local="http://schemas.android.com/apk/res-auto"
    android:id="@+id/DrinksExpandableListView"
    android:minWidth="25px"
    android:minHeight="25px"
    android:layout_width="match_parent"
    android:layout_height="match_parent"
    android:groupIndicator="@null"
    android:childDivider="@android:color/transparent"
    local:MvxBind="ItemsSource Drinks; ItemClick SelectDrinkCommand; GroupClick 
    SelectCategoryCommand"
    local:MvxItemTemplate="@layout/drinks_listitem"
    local:MvxGroupItemTemplate="@layout/drinks_listheader"/`

Cheesebaron:
Might have something to do with our `GroupOnClick` does not set `GroupClickEventArgs.Handled` explicitly. Needs a bit of investigation.

PR #2489, opened 2017-12-22
Fix for Expand/Colapse problem when GroupClick is bound.

AlexStefan:
Bug fix

When GroupClick is bound the group is not expanding/collapsing

When GroupClick is bound the group will still expand/collapse

No

https://github.com/MvvmCross/MvvmCross/issues/2408

- [x] All projects build

### DIFF
--- a/MvvmCross/Binding/Droid/Views/MvxExpandableListView.cs
+++ b/MvvmCross/Binding/Droid/Views/MvxExpandableListView.cs
@@ -151,7 +151,8 @@ namespace MvvmCross.Binding.Droid.Views
 
         private void GroupOnClick(object sender, GroupClickEventArgs e)
         {
-           ExecuteCommandOnGroup(GroupClick, e.GroupPosition);
+            ExecuteCommandOnGroup(GroupClick, e.GroupPosition);
+            e.Handled = false;
         }
 
         private void EnsureItemLongClickOverloaded()


MvvmCross/Binding/Droid/Views/MvxExpandableListView.cs
```diff
@@ -151,7 +151,11 @@ private void EnsureGroupClickOverloaded()
 
         private void GroupOnClick(object sender, GroupClickEventArgs e)
         {
-           ExecuteCommandOnGroup(GroupClick, e.GroupPosition);
+            if (IsGroupExpanded(e.GroupPosition))
```
Cheesebaron: Is this change really needed, don't we just need to set `GroupClickEventArgs.Handled` property to true or false to signal whether we handled the event?

AlexStefan: Tried to find a place to set that property but I could not find one as the args are already in there when the method is getting called. By looking around I saw that there is a difference over how we use ItemClick and GroupClick, but will look more into it tomorrow. (GroupClick exists in both base and Mvx when ItemClick exists only in Mvx, while we have ChildClick in base, and maybe something goes wrong around this?) Any other idea?

Cheesebaron: Just as I suggested, it is enough to just add `e.Handled = false;` after `ExecuteCommandOnGroup(..)`.

AlexStefan:
Not ok, working to remove the bad commits and apply your suggestion.

PR #2490, opened 2017-12-23
New events binding

Saratsin:
Feature:
I've created new target event binding, which uses event name (string) instead of EventInfo and also can use EventArgs parameter as Command parameter

Current MvxEventHandlerEventInfoTargetBinding and MvxEventInfoTargetBinding uses EventInfo reflection property as an argument for binding. Also it doesn't support command parameter in binding.

My approach is to use event name only. It can be simply for use as here: new MvxEventNameTargetBinding(button, nameof(button.Clicked)). Also by default it uses EventHandler EventArgs or TEventArgs and passes it as a CommandParameter.

No, only adding new target binding

- [x] All projects build

### DIFF
--- a/MvvmCross/Core/Binding/Bindings/Target/MvxEventNameTargetBinding.cs
+++ b/MvvmCross/Core/Binding/Bindings/Target/MvxEventNameTargetBinding.cs
@@ -1,0 +1,112 @@
+﻿using System;
+using System.Reflection;
+using System.Windows.Input;
+
+namespace MvvmCross.Binding.Bindings.Target
+{
+    public class MvxEventNameTargetBinding : MvxTargetBinding
+    {
+        private readonly EventInfo _targetEventInfo;
+        private readonly bool _useEventArgsAsCommandParameter;
+
+        private ICommand _currentCommand;
+
+        public MvxEventNameTargetBinding(object target, string targetEventName, bool useEventArgsAsCommandParameter = true) : base(target)
+        {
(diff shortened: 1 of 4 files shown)


Cheesebaron:
1. Would it be possible to make these even subscriptions weak instead?
2. Could you please update docs with relevant information about what has been added and usage

MvvmCross/Core/Binding/Bindings/Target/MvxEventNameTargetBinding.cs
```diff
@@ -0,0 +1,59 @@
...
+        private readonly bool _useEventArgsAsCommandParameter;
+
+        private ICommand _currentCommand;
+
+        public MvxEventNameTargetBinding(object target, string targetEventName, bool useEventArgsAsCommandParameter = true) : base(target)
+        {
+            _targetEventInfo = target.GetType().GetTypeInfo().GetDeclaredEvent(targetEventName);
+            _useEventArgsAsCommandParameter = useEventArgsAsCommandParameter;
+
+            //  addMethod is used because of error:
```
Cheesebaron: According to the link, this should not happen anymore and using `AddEventHandler` should be OK to use

Saratsin: With weak event subscriptions this code is not needed anymore, as all reflection is doing inside weak subscribing logic

Saratsin:
According to your recomendations, I've changed event subscriptions to weak and also update the revelant documentation. WIll be waiting for response)

Saratsin:
Also can't understand why AppVeyor build has failed

PR #2498, opened 2017-12-30
Making it easier to override default IMvxViewModelTypeFinder implementation

nickrandolph:
Feature 

Currently in order to override the implementation of IMvxViewModelTypeFinder you have to override the entire InitializeViewModelTypeFinder method in each setup for your solution

Can now simply override CreateViewModelByNameLookup, CreateViewModelByNameRegistry or RegisterViewTypeFinder to change the lookup, registry or finder

No

N/A

- [ ] All projects build


martijn00:
I'm not sure this is a good idea, every sort of view in MvvmCross is Mvx based. This is so we can internally make changes to fix bugs. If we are supporting non mvx views it might even get harder for us to keep up to date with external changes.

What's wrong with leaving it as is?

nickrandolph:
There's just too much friction involved in getting started with MVX with Forms.
I submitted this PR mainly to get the conversation started on this - MVX is currently a chore to work with and we need to reduce the drag on developers wanting to use it.
If there's resistance to removing the dependency on Mvx based application and pages I can revert those changes and just leave the changes to the Setup that allows for it to be overridden easier. Let me know what you think

nmilcoff:
If anything is difficult to override or if anything fails, we should just fix it / improve it / document it.

But other than that I think MvvmCross has always been more of a full framework than something you can just plug in in an app. I'd say that discussion should be moved into a separate issue.

nickrandolph:
@martijn00 @nmilcoff I've reverted the changes concerning MvxApplication and MvxContentPage, leaving on the change that makes it easier to override the default finder.

PR #2511, opened 2018-01-06
 Make Mvx...Cell inherit from IMvxCell instead of IMvxElement

mubold:
Bug fix, currently all Mvx...Cells are unusable in XamarinForms because MvxCellAdapter expects a IMvxCell and currently all cells inherit from IMvxElement instead of IMvxCell. See
https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross-Forms/MvvmCross.Forms/Views/MvxCellAdapter.cs#L13

I have also changed the playground ListViewPage to use a MvxViewCell instead to test this.

Exception thrown when using a Mvx...Cell

Nope

The ListViewPage in Playground now works with MvxViewCells

- [x] All projects build


UnreachableCode:
I still get an error thrown up on MvxCellAdapter on this one, "eventSource should be a IMvxCell". But this is in a check for cell != IMvxCell, even though it should be one now? Need to do some more digging.

nmilcoff:
Are you using v6 @UnreachableCode?

Issue #1958, opened 2017-06-16
Add nested categories to the documentation


martijn00:
We should add something to have nested categories in the documentation menu. This would enable us to split up certain topics and group them in sub-categories.

An idea could be to use a `category` and `parent-category` tag in the documents. The menu would filter out if a doc had a parent or not, and nest it.

nmilcoff:
We definitely need this done to improve the docs on each platform!  We need a separate folder for each platform inside "Platform specifics"

It would be really helpful if any contributor with web skills could work on it!

PR #2513, opened 2018-01-07
Improve implementation of IMvxAndroidCurrentTopActivity

### LINKED ISSUE #783: IMvxAndroidCurrentTopActivity Activity doesn't always return the correct top activity
I have an activity with 'Theme.Holo.Light.Dialog' theme which makes the activity look like a dialog, hovering an activity.

as you know MvxAndroidLifeTimeMonitor (which implements IMvxAndroidCurrentTopActivity) uses different activity lifetime events to set the IMvxAndroidCurrentTopActivity:Activity 

https://github.com/MvvmCross/MvvmCross/blob/e6d10972b5c28e00e80acc3d9e8910961aa813d6/Cirrious/Cirrious.MvvmCross.Droid/Views/MvxAndroidLifeTimeMonitor.cs

In the case of the dialog however, if I reorient the device, what happens is the dialog activity gets first all the lifetime events, from destruction to creation, followed by the hovered activity.
This makes the hovered activity become the top activity reported by IMvxAndroidCurrentTopActivity.

My specs:
latest stable MvvmCross from NuGet
testing on Nexus 10, Android 4.3
app target Android version is Automatic

I stumbled on this because I have a Close button on dialog which calls MvxViewModel.Close() and because of incorrect top activity reported by IMvxAndroidCurrentTopActivity, it gives this very useful warning:

<code>
mvx:Warning: 71.15 Ignoring close for viewmodel - rootframe's current page is not the view for the requested viewmodel
</code>

https://github.com/MvvmCross/MvvmCross/blob/e6d10972b5c28e00e80acc3d9e8910961aa813d6/Cirrious/Cirrious.MvvmCross.Droid/Views/MvxAndroidViewPresenter.cs#L73

I haven't found for an exact solution for this.
I will make a sample app if that's helpful.
In my app, a dirty & temporary solution is to add ConfigurationChanges attribute for both the activity dialog and for the activity below it.

### LINKED ISSUE #1044: IMvxAndroidCurrentTopActivity > Activity = null during Fragment.OnCreate()
Hello everyone, 
I think I just found an issue in MVVMCross Fragments and unfortunately, after spending some time on it, I wasn't able to fix it on my own... 

When the "Developers Options > Don't keep activities" flag is set and the Activity is brought back from background to foreground, in the "Fragment > OnCreate()" the IMvxAndroidCurrentTopActivity.Activity should point to the current activity, but it returns null instead :(

As my understanding, it happens because the LifeTimeMonitor.OnCreate() is called only after the Activity.OnCreate() has been completed.
So, when the Fragment.OnCreate() gets called, the LifeTimeMonitor.Activity is still null.
The LifeTimeMonitor.OnCreate() should be called after the Activity.OnCreate and before the Fragment.OnCreate()
But I don't know how to properly fix it...

Above all, the issue affects the ViewModel side, because since all the CIRS event are called during the View.OnCreate(), it is impossible to safely use any Gui service from a ViewModel during the CIRS...

Did anyone already face such issue?
Marco.

nmilcoff:
Code improvement / bug fix

As explained in #783, `IMvxAndroidCurrentTopActivity` doesn't always return the correct value.

Above is fixed, using the solution provided by @dazinator (if you read this, I'd appreciate if you could review this PR)

No. The old implementation is still there and is still the default. The new implementation will only be used if the user has a custom `Application` class that extends `MvxAndroidApplication`.

Run Playground.Droid. Check it works. Comment out `MainApplication.cs` and check the app still works.

I'd like to mention that I couldn't reproduce the problem described in the issue (it may be possible to add a repro case in Playground but I don't feel like there is enough information). So I'm trusting @dazinator's implementation on this one :)

Fixes #783.
Probably fixes #1044 too.

-----

I recognize we need to document this stuff, but I'm not really sure where should I write it down. As pointed out in #1958, it would be awesome to have sub-categories for platforms inside "Platform specifics"

- [X] All projects build

### DIFF
--- a/MvvmCross/Droid/Droid/MvvmCross.Droid.csproj
+++ b/MvvmCross/Droid/Droid/MvvmCross.Droid.csproj
@@ -111,8 +111,8 @@
     <Compile Include="Views\Fragments\MvxFragmentExtensions.cs" />
     <Compile Include="Views\Fragments\MvxFragmentViewExtensionMethods.cs" />
     <Compile Include="Views\Fragments\MvxPreferenceFragment.cs" />
-    <Compile Include="Views\MvxMonitorBasedAndroidCurrentTopActivity.cs" />
-    <Compile Include="Views\MvxLifecycleCallbacksBasedAndroidCurrentTopActivity.cs" />
+    <Compile Include="Views\MvxLifecycleMonitorCurrentTopActivity.cs" />
+    <Compile Include="Views\MvxApplicationCallbacksCurrentTopActivity.cs" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\Platform\Droid\MvvmCross.Platform.Droid.csproj">
--- a/MvvmCross/Droid/Droid/MvvmCross.Droid.csproj
+++ b/MvvmCross/Droid/Droid/MvvmCross.Droid.csproj
@@ -111,6 +111,8 @@
     <Compile Include="Views\Fragments\MvxFragmentExtensions.cs" />
     <Compile Include="Views\Fragments\MvxFragmentViewExtensionMethods.cs" />
     <Compile Include="Views\Fragments\MvxPreferenceFragment.cs" />
+    <Compile Include="Views\MvxMonitorBasedAndroidCurrentTopActivity.cs" />
+    <Compile Include="Views\MvxLifecycleCallbacksBasedAndroidCurrentTopActivity.cs" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\Platform\Droid\MvvmCross.Platform.Droid.csproj">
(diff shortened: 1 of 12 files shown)


nmilcoff:
As we discussed in slack (but for future reference), we'll keep both implementations to support apps that won't be able to use our base Android Application class, and also to provide backwards compatibility. 

The PR is now ready to be reviewed again.

dazinator:
Feel free to compare with implementation here: https://github.com/dazinator/MonoDroid.ViewLifecycleManager/tree/develop/src/MonoDroid.ViewLifecycleManager

Also one thing to be careful with (perhaps you want to tweak) is that, say you launch a new intent to show an external activity (for example android account manager), you end up in the situation that the current activity gets paused, and this marks paused activities as not current. When the external activity then finishes with a result code, if you are overriding the method on your activity to receive that result code, then, as it seems to fire before the activity is "resumed", there is no "current" activity at this point in time. I have worked around this in my application by ensuring that when my activity receives a result code, it just temporarily assigns it, amd then later processes it in onResume() in which the activity will be marked current again.

jaykannan:
Came of use today, thanks @xleon

Issue #2520, opened 2018-01-09
Inconsistency between MvxCommand<T> and MvxAsyncCommand<T> implementing IMvxCommand


anne-k:
### Steps to reproduce :scroll:

In #1948, generic interfaces were added for IMvxCommand. Previously, I could use a property typed IMvxCommand and back it with an MvxCommand, MvxCommand\<T>, MvxAsyncCommand or MvxAsyncCommand\<T>. After this change, MvxAsyncCommand\<T> no longer implements the non-generic interface IMvxCommand, but MvxCommand\<T> still does.

### Expected behavior :thinking:
MvxCommand and MvxAsyncCommand are synchronous and asynchronous versions of each other and should exhibit the same behaviour as much as possible. Either the generic typed versions of both commands should implement the non-generic interface, or both of them shouldn't.

### Actual behavior :bug:
MvxCommand\<T> implements the interface IMvxCommand as well as IMvxCommand\<T>, whereas MvxAsyncCommand\<T> only implements IMvxCommand\<T>. This is inconsistent.

The best solution in my opinion would be to have MvxCommand\<T> no longer implement IMvxCommand either, though it would force us (and probably others too) into a lot of refactoring (probably for the better). But either way, they should be consistent.

### Configuration :wrench:

**Version:** 5.6.3

**Platform:** 
- [x] :iphone: iOS
- [x] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

martijn00:
Can you make a PR to fix this? I think we can make this a breaking change for 6.0.

From what i looks like it is fairly easy, just remove the second interface, and add `RaiseCanExecuteChanged` to the `IMvxCommand<T>` interface.

softlion:
I do not agree. The best would be to implement the non generic interface for many reasons:
- AOT compilation size will decrease
- You can create an array of IMvxCommand to store all commands, which has many benefits

nickrandolph:
@softlion thanks for that - I was just about to come back and say something similar

anne-k:
@softlion - your first point: I didn't think of that, how big of a difference do you think it will make?
[...]

PR #2536, opened 2018-01-12
Remove IMvxCommand implementation from MvxCommand<T>

anne-k:
Inconsistency fix

MvxCommand\<T> implements both IMvxCommand and IMvxCommand\<T>.

MvxCommand\<T> no longer implements IMvxCommand, consistent with MvxAsyncCommand\<T> not implementing IMvxAsyncCommand and it forces typing.

Yes, all MvxCommand\<T> can no longer be cast to IMvxCommand. Use the typed version IMvxCommand\<T>.

#2520 

- [ ] All projects build

### DIFF
--- a/MvvmCross/Core/Core/ViewModels/IMvxCommand.cs
+++ b/MvvmCross/Core/Core/ViewModels/IMvxCommand.cs
@@ -27,6 +27,8 @@ namespace MvvmCross.Core.ViewModels
         [Obsolete("Use the strongly typed version of CanExecute instead", true)]
         new bool CanExecute(object parameter);
 
+        void RaiseCanExecuteChanged();
+
         void Execute(T parameter);
 
         bool CanExecute(T parameter);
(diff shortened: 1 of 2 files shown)


nickrandolph:
Would we not want to go the other way and add the non-generic IMvxAsyncCommand interface to the MvxAsycCommand? Having the ability to case a generic to the non-generic version is very useful when you don't have enough type information but want to invoke the command without resorting to reflection.
If you take IEnumerable<T> and IEnumerable, they're linked in this way - https://msdn.microsoft.com/en-us/library/9eekhta0(v=vs.110).aspx

anne-k:
I considered that, but they can all be cast to ICommand if that's what you want. Is there a good reason to be able to cast to a non-generic while still making the distinction between sync and async commands?

I've been updating MvvmCross in our project to 5.6 from 4.4 the past week and a theme in this update seems to be stronger typing, especially with the new navigation system. I think this change fits into that. In what situation would you want to invoke a command without knowing which type its parameter should be (and that it takes a parameter at all)? Currently in our project you have to 'just know' and code can easily break if someone changes the underlying command and forgets about one of the places where it's invoked; this way, the correct type would be enforced by the compiler.

It could be I'm missing a use case though. To be honest I'm fine with going the other way too, the most important thing for me is that it's consistent. And I've been moving over all of our async commands because they broke with 5.6 and I like what it does with our code, so I felt like enforcing it for both commands would be a good idea. I'm very much open for discussion :)

PR #2548, opened 2018-01-19
MvvmCross.Droid.Support.V14.Preference

andrechi1:
The class MvxPreferenceFragment throws an exception when it is created. The problem is the cross of MvvmCross.Droid.Support.V14 with MvvmCross.Droid.Support.V7

Using the MvvmCross.Droid.Support.V14

- [x] All projects build

### DIFF
--- a/MvvmCross.Android.Support/V14.Preference/EventSource/MvxBaseFragmentAdapter.cs
+++ b/MvvmCross.Android.Support/V14.Preference/EventSource/MvxBaseFragmentAdapter.cs
@@ -9,11 +9,10 @@ using System;
 using Android.App;
 using Android.Content;
 using Android.OS;
-//using Android.Support.V4.App;
 using MvvmCross.Droid.Views;
 using MvvmCross.Platform.Core;
 
-namespace MvvmCross.Droid.Support.V4.EventSource
+namespace MvvmCross.Droid.Support.V14.EventSource
 {
     public class MvxBaseFragmentAdapter
     {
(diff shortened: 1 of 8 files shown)


Cheesebaron:
You shouldn't mix Support.V14 with V7, instead if you want to use PreferenceFragment with V7, use the V7 PreferenceFragmentCompat.

I don't think this PR is valid at all, since it duplicates classes which are already present in the framework.

andrechi1:
I want to use Support.V14. But i can't create an intance of the MvxPreferenceFragment class. The problem is that in Support.V14 inherits of App.Fragment. But in Support.V7.Preference.MvxPreferenceFragmentCompat inherits of V4.App.Fragment.

Cheesebaron:
You still fail to explain why that is a problem. If you are already using V7 elsewhere it doesn't make sense to use V14.Preference.

andrechi1:
I already use Android.App.Fragment in other parts of the application and I have problems with  android.support.v4.app.Fragment. I use MvxAndroidViewPresenter instead MyMvxAppCompatViewPresenter. But not build with MvxPreferenceFragmentCompat because inherits of V4.App.Fragment. The Java.Lang.ClassCastException: com.****.SettingsView cannot be cast to android.support.v4.app.Fragment exception is thrown.

I've tried this. But it doesn't work.
  ```
protected override IMvxFragmentView CreateFragment(MvxBasePresentationAttribute attribute, string fragmentName)
        {
            try
            {
                if (typeof(Android.Support.V4.App.Fragment).IsAssignableFrom(attribute.ViewType))
                    return (IMvxFragmentView)Android.Support.V4.App.Fragment.Instantiate(CurrentActivity, fragmentName);
                else
                    return (IMvxFragmentView)Android.App.Fragment.Instantiate(CurrentActivity, fragmentName);
            }
            catch (Exception ex)
            {
                throw new MvxException($"Cannot create Fragment '{fragmentName}'. Use the MvxAppCompatViewPresenter when using Android Support Fragments", ex);
            }
        }
```

Issue #1999, opened 2017-06-28
UWP support for StarterPack package


Shtong:
### Steps to reproduce

1. Create a new projet of type "Cross Platform App (Xamarin)"

2. In the project creation options, leave everything at the default values (Blank App, Xamarin.Forms technology, Shared Project, no cloud)

3. After the projects have been created, install the MvvmCross.StarterPack in the Universal Windows project

### Expected behavior
A few files should get created in the project (like a setup.cs for example)

### Actual behavior
A new reference is added to the project, but everything else is left unchanged

### Configuration

**Version:** 5.0.5
**Visual Studio:** 15.2
**Xamarin:** 4.5.0.486

**Platform:** 
- [ ] iOS
- [ ] Android
- [ ] WPF
- [x] UWP

nmilcoff:
@willsb can you validate UWP is not currently supported by StarterPack? The platform is Forms though, and is marked as done in #1659

willsb:
@nmilcoff UWP is not supported. The mark is for WPF projects, not UWP. The reason is my lack of Windows machine :/

PR #2551, opened 2018-01-22
Add UWP StarterPack content

kipters:
Adds support for UWP projects in `MvvmCross.StarterPack` package

Nothing gets added to UWP projects if you install the `MvvmCross.StarterPack` package

The basic infrastructure (`Setup.cs`, `DebugTrace.cs`, `MainView`) gets added to UWP projects

No

None. But I had to override `GetDefaultLogProviderType()` in Setup.cs.

This should solve #1999, at least partially (doesn't cover Forms projects)

Not sure if I need to update docs

- [x] All projects build


martijn00:
Thanks! This is a very useful PR! I think we might have to update a couple of things when we refactor for .net standard, but it looks good!

kipters:
Ok, I'll watch development of other starter packs and update it when the others are ready to be ported to netstandard. I was also thinking I should update this one to use abstract classes for pages instead of the ugly Vm property, I'll send another PR later today

kipters:
There are a few issues with the UWP XAML compiler and hidden members, so until Microsoft fixes it the ugly property is a more viable choice than the generic MvxWindowsPage-inheriting-abstract-class suggested in the docs, unless you want to give up x:Bind 😞

Issue #2540, opened 2018-01-15
Switch NUnit tests to XUnit


Cheesebaron:
NUnit console runner does not seem to work well with `netcoreapp2.0`. The test engine consistently fails.

`dotnet` cli tool seems to be able to run the tests, however, it does not accept much configuration.

XUnit is much better in this regard, lets switch.

Related to #2530

PR #2566, opened 2018-01-29
Convert UnitTests to XUnit

Cheesebaron:
Chore

Mostly NUnit tests. NUnit console runner does not like .netcore2.0 😢 

Stuff has been converted to XUnit and enabled in build script

Nope

See unit tests being run on AppVeyor builds

#2540 

- [x] All projects build


MvvmCross.Tests/MvvmCross.Test/MvxTestLog.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace MvvmCross
 {
-    internal static class MvxTestLog
+    public static class MvxTestLog
```
martijn00: Please revert back to internal.

MvvmCross.Tests/MvvmCross.Test/MvvmCross.Test.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
martijn00: Why add this back in?

MvvmCross.Tests/MvvmCross.UnitTest/MvvmCross.UnitTest.csproj
```diff
@@ -2,17 +2,18 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.0</TargetFramework>
     <IsPackable>false</IsPackable>
-    <AssemblyName>MvvmCross.Tests</AssemblyName>
-    <RootNamespace>MvvmCross.Tests</RootNamespace>
+    <AssemblyName>MvvmCross.UnitTest</AssemblyName>
```
martijn00: This naming convention is not picked up by the target and prop files.

MvvmCross/MvvmCross.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net461;uap10.0.16299;Xamarin.iOS10;Xamarin.Mac20;Xamarin.TVOS10;Xamarin.WatchOS10;MonoAndroid80;netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net461;uap10.0.16299;Xamarin.iOS10;Xamarin.Mac20;Xamarin.TVOS10;Xamarin.WatchOS10;MonoAndroid80</TargetFrameworks>
```
martijn00: Why remove this target?

Cheesebaron: The target was _only_ for `MvxIoCSupportingTest.cs`

MvvmCross.Tests/MvvmCross.Test/MvvmCross.Test.csproj
Cheesebaron: This is a new Test project with shared test stuff. I don't think it should be in the core `MvvmCross` project.

MvvmCross.Tests/MvvmCross.Test/MvxTestLog.cs
Cheesebaron: Sure

MvvmCross/MvvmCross.csproj
martijn00: Yes, but we can add more test helpers in there. With that you could install the main MvvmCross lib into any project without pulling in a lot of other dependencies.

MvvmCross.Tests/MvvmCross.UnitTest/MvvmCross.UnitTest.csproj
Cheesebaron: Only 2 or 3 of the old assemblies were picked up either way, because of naming inconsistencies. Will adjust the prop.

MvvmCross/MvvmCross.csproj
Cheesebaron: That doesn't make sense. What if I add a shared Fixture that depends on Xunit? Then all of the sudden MvvmCross depends on Xunit everywhere.

martijn00: With multi target that would only add the dependency in the netcoreapp target.

Cheesebaron: It still doesn't make sense, because if we want to support netcoreapp2.0 in the future, not just use it for testing, we will need to break stuff again.

martijn00: Hmm, right you got a point there. So doesn't the new project need to target netcoreapp then?

Cheesebaron: netstandard there is fine. It is just our own test assemblies that need to be netcoreapp2.0 or net full framework

Cheesebaron:
Now to fix the tests... A whole bunch of them are failing 😭

Cheesebaron:
Looks pretty green to me now. @martijn00 can you review again please.

FinHorsley:
Does this PR mean that the [Testing](https://www.mvvmcross.com/documentation/fundamentals/testing) docs are out of date? (they are still recommending using NUnit)

Cheesebaron:
You can use whatever you want for _your_ unit tests. We don't impose any limitation to which testing framework you use here.

PR #2583, opened 2018-02-05
Move actual unit tests and test project to correct folders

martijn00:

- [x] All projects build


Directory.build.props
```diff
@@ -31,6 +31,6 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(IsTestProject)' != 'true' and '$(SourceLinkEnabled)' != 'false' and '$(IsLibraryProject)' == 'true'">
-    <PackageReference Include="SourceLink.Create.CommandLine" Version="2.7.6" PrivateAssets="All" /> 
+    <PackageReference Include="SourceLink.Create.CommandLine" Version="2.8.0" PrivateAssets="All" />
```
Cheesebaron: I don't think PackageReferences should be updated in this PR.

martijn00: Ooh crap, that should be another PR. It's not a big deal though?

Cheesebaron: I think it would be healthy to make a new PR with these changes :) Imagine what kind of example you are setting by just letting it slide.

PR #2587, opened 2018-02-06
[WIP] Add Async version of Start and mark non-async version obsolete

### LINKED ISSUE #2586: MvxAppStart swallows exceptions
@nickrandolph pinged us on Slack that the following was an issue:

1. Make `Start` crash in `MvxAppStart`
2. See exception crash the app, can't be caught from outer scope

Exception should bubble up to the caller, but doesn't because we use `async void`.

### Actual behavior :bug:
App crash

### Configuration :wrench:

**Version:** 5.x

Cheesebaron:
Bug fix

Exception swallowed

Added new async version of Start, which bubbles up exceptions

Kind of

Start your app

Fixes #2586 

- [x] All projects build


martijn00:
I vote to remove the old one in 6.0 if we break things anyway. Also StartAsync will not be called with this change.

Cheesebaron:
You are right StartAsync won't be called. Not sure what to do about the callee though. Do we fire and forget there? That is the current behavior at least, That might derp things up when rotating the device on Android, which is probably also an issue now.

Maybe we should look in to @garuma's LibActivityTask to help with these lifecycle issues.

🤔

nickrandolph:
I think LibActivityTask is something worth looking at long term. I don't think it's the right solution for solving this startup issue.
imho I think we should proactively drop the void Start method, rather than just deprecating it. However, we should make sure that all the playground samples run and behave correctly on startup to make sure this change doesn't introduce issues with startup

Cheesebaron:
Blocked by #2593

nickrandolph:
I think this PR might be able to be abandoned with the work that has gone into https://github.com/MvvmCross/MvvmCross/pull/2674
The new PR proposes a synchronous start that I feel is a better approach than attempting to make start async.

Cheesebaron:
I think we could put some effort into parallelizing some of the things in Setup. Right now everything runs sequentially, while we could probably do a bunch of it in parallel.

nickrandolph:
@Cheesebaron I agree there is work to be done in vnext around improving perf for setup. Just think that making things async isn't something that should be pushed due to the limitations on most platform of doing a synchronous startup. 
I'm not sure that doing things in parallel will improve perf on startup due to task switching that the OS would have to do. However, I think there is a lot we can do to weed our unnecessary use of reflection for locating views/viewmodels/dependencies etc. If we were to use something like Fody, or other code generation, we can generate code that would register view/viewmodel pairs at build time, rather than having to do a reflection lookup for them. I believe this would improve startup perf, which still making it simple for developers (ie they still wouldn't have to manually register each views/viewmodels).
Anyhow, I'd suggest closing this PR and we can open an issue around improving startup perf

PR #2593, opened 2018-02-07
Updating various startup logic

nickrandolph:
I've started going through various pieces of startup logic. My concern is that Android isn't starting correctly but VS is being a pain and I haven't been able to even step through the logic successfully.


nickrandolph:
@Cheesebaron hey sorry about being a bit absent, where's this set of changes at now? Let me know what it's waiting on and I can help finish it.

nickrandolph:
@Cheesebaron hey sorry about being a bit absent, where's this set of changes at now? Let me know what it's waiting on and I can help finish it.

Cheesebaron:
I merged it into my branch here: https://github.com/MvvmCross/MvvmCross/pull/2587

nickrandolph:
@Cheesebaron I was chatting with @martijn00 about this and we concluded we'd wait for after v6 is baked before we pursue this.

Cheesebaron:
I don't disagree. It also needs to be tested well, so we don't encounter new gremlins by making it asynchronous.

PR #2597, opened 2018-02-09
Using IActivatedEventArgs in UWP

MartinZikmund:

**New feature**
This PR switches the `LaunchActivatedEventArgs` class in the constructor of `MvxFormsWindowsSetup` class for the more general `IActivatedEventArgs` interface.

The app can be launched in many **other ways than just by direct launch** (URI, File association, etc.), and instead of `LaunchActivatedEventArgs` you then have an instance of different class implementing `IActivatedEventArgs` and can't therefore create `MvxFormsWindowsSetup` instance.

Internally, `MvxFormsWindowsSetup` calls `Xamarin.Forms.Forms.Init` method, which expects an instance of `IActivatedEventArgs` and doesn't require an instance of `LaunchActivatedEventArgs` specifically.

The `MvxFormsWindowsSetup` constructor requires an instance of `LaunchActivatedEventArgs`.

The `MvxFormsWindowsSetup` constructor expects a more general instance of `IActivatedEventArgs`.

No, because `LaunchActivatedEventArgs` implements `IActivatedEventArgs`, the new interface is just more permissive and allows other activation types like URI, File, Search and many others.

[IActivatedEventArgs in Docs](https://docs.microsoft.com/en-us/uwp/api/Windows.ApplicationModel.Activation.IActivatedEventArgs).

[LaunchActivatedEventArgs in Docs](https://docs.microsoft.com/en-us/uwp/api/Windows.ApplicationModel.Activation.LaunchActivatedEventArgs).

- [x] All projects build

### DIFF
--- a/MvvmCross.Forms/Platform/Uap/Core/MvxFormsWindowsSetup.cs
+++ b/MvvmCross.Forms/Platform/Uap/Core/MvxFormsWindowsSetup.cs
@@ -48,7 +48,7 @@ namespace MvvmCross.Forms.Platform.Uap.Core
             {
                 if (_formsApplication == null)
                 {
-                    Xamarin.Forms.Forms.Init(ActivatedEventArgs);
+                    Xamarin.Forms.Forms.Init(ActivationArguments);
                     _formsApplication = _formsApplication ?? CreateFormsApplication();
                 }
                 return _formsApplication;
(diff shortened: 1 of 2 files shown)


nickrandolph:
Yeh I don't think we'll want to make that change in this PR but it does bode the question as to how we can improve UWP support in MVX going forward

martijn00:
I would vote for changing this in the base too, especially since we are working on 6.0.

MartinZikmund:
@martijn00 I will update the code to include it in the base too. Which approach sounds best to you? To pass it into the constructor and add a public `IActivatedEventArgs` property to the `MvxWindowsSetup` class?

MartinZikmund:
I have added a new constructor to `MvxWindowsSetup` with the `IActivatedEventArgs`. Also Added new protected property `ActivationArguments` that provides access to the passed in instance if any.

martijn00:
Looks good! Can you do a rebase?

MartinZikmund:
@martijn00 Sorry, I was away for the weekend. Thank you for merging 👍 !

Issue #2600, opened 2018-02-10
Fragment close does not work if fragment presentation attribute has backstack set to false


thefex:
### Steps to reproduce :scroll:

1. Use traditional Mvx fragment flow, but make sure your [MvxFragmentPresentationAttribute] has isBackStack field set to false. 

2. Close viewmodel attached to fragment

3. Activity is closed instead of fragment.

### Expected behavior :thinking:
Fragment should be closed.

### Actual behavior :bug:
Activity is closed.

### Configuration :wrench:

**Version:** most recent nuget + develop branch

**Platform:** 
- [ ] :iphone: iOS
- [X] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

thefex:
I'd like to do that but for unknown reason I can't load MvvmCross support projects -> "Project does not support framework _Version=8.0"

PR #2601, opened 2018-02-10
 #2600 no backstack fragment close does not work hotfix

thefex:
Fragment close does not work if fragment presentation attribute has backstack set to false

fix

- [ ] All projects build


nmilcoff:
I guess the same issue exists in the support presenter, right? If so, can you maybe apply the same fix?

thefex:
@nmilcoff I'd like to do that but for unknown reason I can't load MvvmCross support projects -> "Project does not support framework _Version=8.0"

Issue #2515, opened 2018-01-07
MvxObservableCollection reports wrong index when doing AddRange


aaaabcv:
### Steps to reproduce :scroll:

1.
Create observable collection using MvxObservableCollection

2.
Add items with AddRange

### Expected behavior :thinking:
NotifyCollectionChangedEventArgs should report the first index where the changed occured  and the items which are inserted

### Actual behavior :bug:
NotifyCollectionChangedEventArgs report the index which is the same as the end of the resulting collection

Fix suggestion:
https://github.com/MvvmCross/MvvmCross/blob/c4cbc21d8417abd16a0cca956f8bee722c8e4faa/MvvmCross/Core/Core/ViewModels/MvxObservableCollection.cs

In line 99 "this.Items.Count" is taken after the items have been inserted. If this value is captured before the imtes are inserted and them reported, behaviour would be ok, i believe.

Also refer to 
https://msdn.microsoft.com/en-us/library/ms653208(v=vs.110).aspx

### Configuration :wrench:

**Version:** 6.2

**Platform:** 
- [x ] :iphone: iOS
- [ x] :robot: Android
- [ x] :checkered_flag: WPF
- [ x] :earth_americas: UWP
- [ x] :apple: MacOS
- [ x] :tv: tvOS
- [ x] :monkey: Xamarin.Forms

nmilcoff:
Can you please confirm this issue still exists in v5.6.3? #2455 should have solved it.

aaaabcv:
Yes, it appears to solve the issue, thanks! And sorry, I didn't try 5.6.3 earlier.
I still have concerns about the change of #2455 since the index should indicate the start index where the change occured. The change indicates the index after the end of the list (Items.Count). Apparently the RecylerView Adapter can cope with that but it might be that other lists may have problems with that kind of hint.
I might add that to the change of #2455

martijn00:
@aaaabcv Can you make a PR for that?

ctartamella:
FWIW, I am on 5.6.3 and seeing this issue when calling AddRange() on a previously empty list in the case I'm initially populating it (~1500 items).  Calling ```data.ForEach(d => collection.Add(d));``` alleviates it as a workaround (but looks kinda ugly).

PR #2614, opened 2018-02-13
Fixed issue #2515 where MvxObservableCollection.AddRange() reports wrong index

Strifex:

Bug Fix

The index is incorrectly reported when AddRange is called, which could cause issues with certain collections responding to NotifyCollectionChangedEventArgs events.  

MvxObservableCollection will stop reporting the wrong index

Not that I'm aware of, I've tested it against a MvxRecyclerView and haven't seen any regression in behavior.  Unless there is a custom control using the index parameter incorrectly, this should work as expected.

Bind a MvxRecyclerView to an MvxObservableCollection<T> and call AddRange multiple times and check the index being reported.

Original Issue:
https://github.com/MvvmCross/MvvmCross/issues/2515

See this MSDN for details on how index parameter should be passed: https://msdn.microsoft.com/en-us/library/ms653208(v=vs.110).aspx

- [x] All projects build

### DIFF
--- a/MvvmCross/ViewModels/MvxObservableCollection.cs
+++ b/MvvmCross/ViewModels/MvxObservableCollection.cs
@@ -84,6 +84,7 @@ namespace MvvmCross.ViewModels
                 throw new ArgumentNullException(nameof(items));
             }
 
+            int startingIndex = this.Items.Count;
             var itemsList = items.ToList();
             using (SuppressEvents())
             {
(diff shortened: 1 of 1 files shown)


Cheesebaron:
Would be really neat if we could cover `MvxObservableCollection`  with unit tests 😄 @Strifex do you want to add some?

PR #2615, opened 2018-02-13
Reducing code to get started for UWP projects

nickrandolph:
Feature

Currently developers have to put the right code into the UWP app.xaml.cs etc in order to get started with MVX

Simplified the code with extensibility if required

No

Review playground.UWP project

- [ ] All projects build


nickrandolph:
@MartinZikmund  let me know what you think of these changes - keen to reduce the clutter for new projects
@martijn00 pls confirm the MvxBaseApplication name and folder is in line with mvx style/naming

MartinZikmund:
I really like this :-) . It makes it easy to start a fresh new project and can be extended if necessary, great it has the `MvxSuspensionManager` built in, too. And great it already supports the `IActivatedEventArgs`

MvvmCross/Platform/Uap/Core/MvxTypedWindowsSetup.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Collections.Generic;
+using System.Reflection;
+using MvvmCross.Platform.Uap.Views;
+using MvvmCross.ViewModels;
+using System.Linq;
+using Windows.UI.Xaml.Controls;
+
+namespace MvvmCross.Platform.Uap.Core
+{
+    public class MvxTypedWindowsSetup<TApplication>: MvxWindowsSetup
```
martijn00: I think that having a typed version of the setup could be an idea, but then it should be over all the platforms. Why in particular do you want this? Most people need to have a custom setup anyway. 

For the class, please make the changes to the normal `MvxWindowsSetup` file and add only the generic overrides in there as a new class. Both should have the same name.

MvvmCross/Platform/Uap/Views/MvxBaseApplication.cs
```diff
@@ -0,0 +1,131 @@
...
+using MvvmCross.ViewModels;
+using Windows.ApplicationModel;
+using Windows.ApplicationModel.Activation;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Navigation;
+
+namespace MvvmCross.Platform.Uap.Views
+{
+    public abstract class MvxBaseApplication: Application
```
martijn00: I would remove the word `Base` here. We just put Mvx in front of every class throughout the framework.

MvvmCross/Platform/Uap/Views/MvxBaseApplication.cs
```diff
@@ -0,0 +1,131 @@
...
+    {
+        protected MvxWindowsSetup InternalSetup { get; set; }
```
martijn00: InternalSetup seems a bit strange as name. Especially since it is protected.

Projects/Playground/Playground.Uwp/App.xaml.cs
```diff
@@ -1,107 +1,19 @@
...
-            //TODO: Save application state and stop any background activity
-            deferral.Complete();
-        }
-
-        private async void OnResuming(object sender, object e)
+        protected override MvxWindowsSetup CreateSetup(Frame rootFrame, string suspension)
         {
-            var suspension = Mvx.GetSingleton<IMvxSuspensionManager>() as MvxSuspensionManager;
-            await suspension.RestoreAsync();
+            return new MvxTypedWindowsSetup<Core.App>(rootFrame, suspension);
```
martijn00: The setup is normally automatically called from somewhere if i remember correctly. Why does this need to be created here if that is the case?

MvvmCross/Platform/Uap/Core/MvxTypedWindowsSetup.cs
nickrandolph: Ok will change name back - I was having a lot of issues with multi-targetting breaking builds and thought that it was due to it not being able to resolve the generic override :-(

MvvmCross/Platform/Uap/Views/MvxBaseApplication.cs
nickrandolph: ok

MvvmCross/Platform/Uap/Views/MvxBaseApplication.cs
nickrandolph: Just "Setup" then?

nmilcoff:
I think this is a nice improvement, but we shouldn't hide the Setup class behind the curtain. Although it may be bit of a boost to get started, we've always implicitly encouraged users to add their own code to that class (register IoC stuff, custom bindings, converters, ...).

nickrandolph:
@martijn00 I've rebased and have updated most of the names. With the additional classes for UWP Forms I think we may need to think about naming again for these

nickrandolph:
@nmilcoff I think it should be possible to get started without setup (as per the current playground UWP sample) but I take your point about making it easy for devs to override the setup. I think compromise would be to add the setup back into the sample just to demonstrate how you can override it. Does that work for you?

nmilcoff:
@nickrandolph I personally think we should still force users have their own Setup class. 
Our documentation is not great - no surprises there - but the Setup class has always been there around and I don't think beginners find it difficult to understand.

As I previously said, Setup has always worked as a place to initialize and register your own stuff. So I'm a bit concerned we'll start getting issues / questions about where to do that if we provide a default implementation. You'll actually need to dig more into the docs / source code to know how to use a custom Setup.

I'd like to know what does @MvvmCross/core think about this. If I'm the only one concerned about this change, I'm fine having this IN 🙂. We just need to ensure we provide support for that on all/most platforms to avoid too much fragmentation.

nickrandolph:
@nmilcoff I definitely agree with regards to making all the platforms as similar as possible - right now this is definitely not the case, and the work here is much more in line with some of the way Android works.

I disagree with forcing the inclusion of a setup file for every project. You're right in that a lot of projects do require it. However, particularly with XF projects, this is not the case, since most of the DI actions can be done within the XF world. 

Fundamentally my thoughts are that MVX is way to hard for new devs to ramp up. Even experienced devs, there are way to many init steps required to get all platforms working. I'd prefer to see less code/files required initially, supplemented with documentation/FAQ that is structured around "how do I do XYZ" (which is much easier to maintain that documentation that describes the overall framework etc)

martijn00:
Merged this for now. It still needs some extra work. We should remove the need to set a Setup like that, and align the other platforms like this.

Issue #2512, opened 2018-01-06
MvxFormsApplication Start, Sleep and Resume gets not called on iOS


Sebastian1989101:
### Steps to reproduce :scroll:

No steps required.

### Expected behavior :thinking:
Start, Sleep and Resume gets called at the correct moment like it is on a non MVX Forms app.

### Actual behavior :bug:
All three methods are ignored by MVX on iOS. It works on Android (at least Start and Sleep; got no call for Resume there too). 

### Configuration :wrench:

**Version:** tested with 5.5.0-5.6.3 - all versions had this problem

**Platform:** 
- [ ] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [x] :monkey: Xamarin.Forms

Issue #2223, opened 2017-09-25
OnStart is not called for a Xamarin Forms iOS application


MichaelBTR:
### Steps to reproduce
1. Create a new Xamarin Forms project using MvvmCross

2. Derive from the MvxFormsApplication class and override the OnStart method so that there is code that should be executed when the app starts.  

3. Run the app on iOS

**Note**: an example is available from https://github.com/MichaelBTR/MvxForms. I've forked the repo from https://github.com/martijn00/MvxForms, updated to MvvmCross 5.2 and added local notifications to be displayed. I believe the issue also occurred in 5.1.1

### Expected behavior
The statements within the OnStart override is executed

### Actual behavior
The statements within the OnStart override is not executed

### Configuration

**Version:** 5.2

**Platform:** 
- [x] iOS
- [ ] Android
- [ ] WPF
- [ ] UWP

martijn00:
The code for OnStart is called from the Xamarin.Forms platform. If it is not called in your case i would think this is a bug on their side. Can you look into that to confirm?

MichaelBTR:
Apologies, I should've clarified that I'm aware it's a Xamarin.Forms platform method and that I've noticed that it works fine when creating a Forms app where the application class doesn't derive from MvxFormsApplication (e.g. using the out of the box Master Detail template in Visual Studio). I've updated the sample that I've forked from your repository to include another folder/projects to demonstrate that it works when not deriving from MvxFormsApplication

sergio11:
I have the same problem for MvvmCross.Forms 5.1.0

PR #2617, opened 2018-02-13
Use the basic Forms application type instead of the Mvx one

martijn00:

Bug fix

You need to extend the Mvx type. Forms can't handle that one correctly

Use the basic Forms type. People could still use the mvx type if they want.

Yes

Run the playground

Fixes https://github.com/MvvmCross/MvvmCross/issues/2512
Fixes: https://github.com/MvvmCross/MvvmCross/issues/2223

- [x] All projects build


nmilcoff:
Shouldn't we just remove our application class?

nickrandolph:
@martijn00 @nmilcoff we should see whether this issue (ie Onstart not being called when using MvxApplication) has been fixed in the latest version of XF that just came out. If it has, I still think we accept this PR but I would leave the MvxApplication just in case devs want to use it

martijn00:
@Cheesebaron It is all done, but Forms is not working, but neither is it in the develop branch, so i'm not sure what to do here. I'm trying to fix the playground some time soon.

nickrandolph:
Hey guys apologies I’ve been a bit absent. I’ll try to jump in tomorrow and see if I can work out the issue with XF.

Projects/Playground/Playground.Forms.Uwp/Setup.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Collections.Generic;
```
nickrandolph: @martijn00 this file shouldn't be included in the UWP project any more

nickrandolph:
@martijn00 can you rebase and remove the UWP setup class and then this should be good to go.

martijn00:
@nickrandolph done!

nmilcoff:
Sorry guys but I still don't get why are keeping our own FormsApplication class if it's buggy? 🤔

martijn00:
It's actually not buggy, it works as expected. I think it might have been a bug in Xamarin.Forms. Our version has some advantages, so we could keep it, but i think we want to use the native type so people can choose themselves.

PR #2632, opened 2018-02-23
Binding types fix

Saratsin:
Bug fix

Weak subscription couldn't find event info, because it uses typeof instead of GetType

Now you implicitly passing view type in generic constructor, so binding goes normal

No. It's fixes EventName bindings which were not used before in projects

- [x] All projects build


martijn00:
Looks good, but should you maybe use `MvxTargetBinding<TTarget, TValue>` instead?

https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Binding/Bindings/Target/MvxTargetBinding.cs#L39

Saratsin:
Well, in case of generic `MvxTargetBinding`, `TargetType` property would be by default `typeof(TTarget)`, when I use `typeof(ICommand)` instead. So I think, I can't use here generic version

PR #2635, opened 2018-02-25
Adding support for custom renderer assemblies to MvxFormsWindowsSetup

MartinZikmund:

**Feature**

When Xamarin.Forms UWP projects are built in Release mode, .NET Native compilation is used and that causes problems when there are secondary assemblies with Xamarin.Forms `ViewRenderers`.

Developers can provide the additional renderer assemblies in similar manner to `ViewAssemblies` - by overriding the new `GetRendererAssemblies` method in `MvxFormsWindowsSetup`.

The `Xamarin.Forms.Forms.Init` method is called with only the `IActivatedEventArgs` parameter.

`Xamarin.Forms.Forms.Init` method is called with the optional `rendererAssemblies` parameter which is necessary to make view renderers in secondary assemblies to be successfully discovered by Xamarin.Forms in UWP Release mode.

No, the parameter is optional, by default the `GetRendererAssemblies` method returns `null` which is the default value for the parameter.

Discussed for example [here](https://forums.xamarin.com/discussion/73132/map-doesnt-work-in-release-mode-uwp) on Xamarin Forums

Also the default cross-platform template for Xamarin.Forms mentions this in a comment.

- [x] All projects build

### DIFF
--- a/MvvmCross.Forms/Platform/Uap/Core/MvxFormsWindowsSetup.cs
+++ b/MvvmCross.Forms/Platform/Uap/Core/MvxFormsWindowsSetup.cs
@@ -30,6 +30,11 @@ namespace MvvmCross.Forms.Platform.Uap.Core
         {
         }
 
+        /// <summary>
+        /// Override to provide list of assemblies to search for views. 
+        /// Additionally for UWP .NET Native compilation include the assemblies containing custom controls and renderers to be passed to <see cref="Xamarin.Forms.Forms.Init" /> method. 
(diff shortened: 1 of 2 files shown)


martijn00:
Isn't it possible to just pass in the `ViewAssemblies` into that? The renderers are views anyways.

MartinZikmund:
@martijn00 You are absolutely right there, it makes sense. I will update the branch accordingly :-) .

PR #2658, opened 2018-03-06
Add readme.txt file to open on nuget install

martijn00:

Feature

This softens the pain of not having a starterpack, or install nuget files.

No

- [x] All projects build


MvvmCross/MvvmCross.csproj
```diff
@@ -74,6 +74,10 @@
     <Compile Include="Platform\Xamarin-common\**\*.cs" />
     <PackageReference Include="Tizen.NET" Version="4.0.0" />
   </ItemGroup>
+  
+  <ItemGroup>
+    <None Include="..\readme.txt" pack="true" PackagePath="." />
```
nickrandolph: @martijn00 is there a reason why the readme is up a folder level? is this just so it doesn't get included as source code?

martijn00: I thought to add it as solution item, so it could work for every project. I could move it into the project itself.

nickrandolph: Given that it has a specific list of files I think move to project and then we may look at having an additional readme for mvvmcross.forms

martijn00:
@nickrandolph Feedback is implemented

Projects/Playground/Playground.Forms.iOS/Setup.cs
```diff
@@ -26,12 +26,5 @@ protected override IEnumerable<Assembly> GetViewAssemblies()
         protected override MvxFormsApplication CreateFormsApplication() => new FormsApp();
 
         protected override IMvxApplication CreateApp() => new Core.App();
-
```
nickrandolph: @martijn00 was this intentionally added to this PR?

Issue #2624, opened 2018-02-16
Lack of Initialization from MvxSplashScreenActivity causes App start from external Intent (ie, Uri routing) to fail in Forms app


borbmizzet:
### Steps to reproduce :scroll:

1.Create a Xamarin.Forms/MvvmCross solution, **including** android MvxSplashScreenActivity
2. Add IntentFilter Attribute to `MainActivity : MvxFormsAppCompatActivity<MyViewModel>`
```
[IntentFilter(new[] { Intent.ActionView },
        Categories = new []
        {
            Intent.CategoryDefault,
            Intent.CategoryBrowsable
        },
        DataScheme = "mvx")]
```
3.
```
[MvxContentPagePresentation(WrapInNavigationPage = true, Animated = false, NoHistory = true)]
public partial class MyPage : MvxContentPage<MyViewModel>
{...}
```
4.  Add Url routing to MyViewModel
```
[assembly: MvxNavigation(typeof(MyViewModel), @"mvx://auth\?data=(?<data>.*)$")]
namespace *.ViewModels{
public class MyViewModel :MvxViewModel{
...
//Some Usage of Mvx.Resolve<IRegisteredInMvxFormsAndroidSetup>() 
during viewmodel prepare/initialization/display steps
...
public void Init(string data){...}
}}
```
5. Use "mvx://auth?data=foo" routing link from browser to run app from **closed** state

### Expected behavior :thinking:
MainActivity should properly initialize the application, run Setup, etc., then properly navigate to and display MyPage after calling MyViewModel:Init(foo)

### Actual behavior :bug:
MainActivity<MyViewModel> opens directly, but fails to display anything except a blank page.  Likely caused by skipping the splash screen and all the setup it entails.  Debug Console doesn't say anything went wrong.

Note that if you click the "mvx://auth?data=foo" link after manually launching the application and starting MainActivity after displaying the splash screen, it works fine.

### Configuration :wrench:

**Version:** 5.6.3

**Platform:** 
- [ ] :iphone: iOS
- [*] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [*] :monkey: Xamarin.Forms

Issue #2622, opened 2018-02-15
Xamarin.Forms Android - First page cannot reference Application level StaticResources


duckwaffle:
Repo for repro: https://github.com/duckwaffle/MvxFormsVmDuplication

### Steps to reproduce :scroll:

1. Add resources (any resource will do, color, converter whatever) to your Xamarin.Forms App.xaml

2. Go to the XAML of your first launched page (whatever page is for RegisterNavigationServiceAppStart<>) and add an element which references the StaticResource

3. Start the app

### Expected behavior :thinking:
Tell us what should happen
The app should start, and the visual element that references the StaticResource will be visible

### Actual behavior :bug:
Tell us what happens instead
App fails to start. Occasionally you will see a TargetInvokationException and something in the output will say that it cannot find the StaticResource for the key you provided, but more often than not you will just see a blank screen because the first navigation never happened.

When running the UWP project, everything works fine, however the Android one should show the behaviour I'm describing.

### Configuration :wrench:

**Version:** 5.6.3

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [X] :monkey: Xamarin.Forms

nmilcoff:
I'm hitting this issue too, easy to reproduce. I'll try to dig into it in the next couple of days.

PR #2665, opened 2018-03-08
Assure Xamarin's Current Application is set when Intiailize is called

jtillman:
Bug Fix For: https://github.com/MvvmCross/MvvmCross/issues/2622 and also makes https://github.com/MvvmCross/MvvmCross/issues/2624 possible.

The Application.Current Context is lost in the base implementation of `OnCreate` by Xamarin's Built in FormsActivity

If no Current Application is set, we will set it to be the current FormsApplication we have associated with the Activity

Shouldn't for Forms Applications unless people are detecting this edge case manually

Create an Xamarin Forms Application with Android that doesn't have a Splash Screen and its first page depends on Static Resources. This PR, should make the app work properly.

- [ ] All projects build

### DIFF
--- a/MvvmCross.Forms/Platform/Android/Views/MvxFormsAppCompatActivity.cs
+++ b/MvvmCross.Forms/Platform/Android/Views/MvxFormsAppCompatActivity.cs
@@ -105,6 +105,11 @@ namespace MvvmCross.Forms.Platform.Android.Views
 
         public virtual void InitializeForms(Bundle bundle)
         {
+            if(null == global::Xamarin.Forms.Application.Current)
+            {
+                global::Xamarin.Forms.Application.Current = FormsApplication;
+            }
+
             if (FormsApplication.MainPage != null)
             {
                 global::Xamarin.Forms.Forms.Init(this, bundle, GetResourceAssembly());


Cheesebaron:
Looks good to me.

nickrandolph:
@dnfclas are you able provide a repo that shows the scenario this PR fixes? my concern is that we're plugging this the wrong way. In the constructor of the FormsApplication it sets the Application.Current, so we shouldn't have to explicitly set it (tbh I'm not even sure why it has an exposed setter).

jtillman:
@nickrandolph I repo'd this by using the Playground Project by taking away the splash screen and adding a single StaticResource reference on the RootView. as detailed in https://github.com/MvvmCross/MvvmCross/issues/2622

This does feel weird to me that Application.Current is set when creating but if you look at the base FormsAppCompactActivity, they take it away right after: https://github.com/xamarin/Xamarin.Forms/blob/e65b48b7194aff9287aae8619a3964fbf71993fe/Xamarin.Forms.Platform.Android/AppCompat/FormsAppCompatActivity.cs#L168

nickrandolph:
@jtillman if we were to change OnCreate to do the Mvx init after the call to the base OnCreate, would this break things. Right now the setup process isn't consistent with the way Xamarin Forms is supposed to be initialized (see the default File->New->Xamarin Forms Project). What I'm suggesting is:

        protected override void OnCreate(Bundle bundle)
        {
            base.OnCreate(bundle);

            **// Required for proper Push notifications handling      
            var setupSingleton = MvxAndroidSetupSingleton.EnsureSingletonAvailable(ApplicationContext);
            setupSingleton.EnsureInitialized();**

            ViewModel?.ViewCreated();
            InitializeForms(bundle);
        }

I haven't had a chance to test this, and don't know what side effects this would have but it seems this would be a better way than simply setting the Application.Current at some random point in the lifecycle

jtillman:
Should we close this PR @nickrandolph . I'm assuming your changes in https://github.com/MvvmCross/MvvmCross/pull/2674 cover what I'm attempting to address here.

nickrandolph:
@jtillman yes I think we can close this PR - would love you to take a look at my PR as it makes some significant changes to startup and the more feedback/testing the better

PR #2668, opened 2018-03-08
Add generic setup to all platforms

martijn00:

Feature

You can now set the app class as generic.

No

- [ ] All projects build


nickrandolph:
@martijn00 should we look at adjusting the CodeFactor rules so that we don't see errors when we have generic classes in same file as the non-generic class?

martijn00:
@nickrandolph that would be a good idea! Not sure how to do that.

PR #2670, opened 2018-03-08
Add check in navigation service to see if viewmodels are available

martijn00:

Feature

You can only check uri paths

Also possible to check for normal viewmodels

A little to the interface

- [ ] All projects build


nickrandolph:
@martijn00 we should add calls to this service into the playground sample - I think some of the native projects are missing quite a lot of the views

PR #2674, opened 2018-03-10
Ensuring Start completed before initializing Xamarin Forms

nickrandolph:
bug fix

Start calls asynchronous navigation method. If this method invokes a method that takes some time to complete, it is possible for XF to crash because the page attempts to load XF before MainPage is set

Setup blocks after calling Start until the async navigation completes

No

Run playground forms

Related PR (https://github.com/MvvmCross/MvvmCross/pull/2665) - this PR should no longer be required after the changes in this PR which ensure the current application is set whenever the FormsApplicaiton property is accessed.
Related Issues (these need to be tested against these changes)
https://github.com/MvvmCross/MvvmCross/issues/2622
https://github.com/MvvmCross/MvvmCross/issues/2624

- [X] All projects build


MvvmCross/ViewModels/IMvxAppStart.cs
```diff
@@ -2,10 +2,14 @@
...
 
+using System.Threading.Tasks;
+
 namespace MvvmCross.ViewModels
 {
     public interface IMvxAppStart
     {
         void Start(object hint = null);
+
+        Task WaitForStart();
```
martijn00: I would rather make `void Start(object hint = null);` a `Task`. That might be difficult to make async all the way, so maybe another idea is to create a `IMvxFormsAppStart` for now and make it a Task there or add WaitForStart there. Also we could cast it to make it available.

nickrandolph:
@martijn00 can you take another look - somehow in making changes I ended up re-surfacing the nullreferenceexception when calling LoadApplication for Xamarin Forms on Android. Any suggestions on why this might be happening?

MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -2,33 +2,56 @@
...
+            // Check whether Start has commenced, and return if it has
+            if (Interlocked.CompareExchange(ref startHasCommenced, 1, 0) == 1) return;
+
             if (hint != null) {
                 MvxLog.Instance.Trace("Hint ignored in default MvxAppStart");
             }
+            
             try {
-                await NavigationService.Navigate<TViewModel>();
+                _startTaskNotifier = MvxNotifyTask.Create(async ()=> await NavigationService.Navigate<TViewModel>());
```
nmilcoff: The try/catch here won't work, as MvxNotifyTask doesn't re-throw exceptions. You can use the newly added `onException` callback for that.

MvvmCross/ViewModels/MvxNotifyTask.cs
```diff
@@ -31,7 +31,6 @@ private MvxNotifyTask(Task task, Action<Exception> onException)
         {
             try
             {
-                await Task.Yield();
```
nmilcoff: Why did you remove this `Task.Yield`? It was added in purpose.

MvvmCross.Forms/Platform/Android/Views/MvxFormsAppCompatActivity.cs
```diff
@@ -93,26 +94,43 @@ protected override void AttachBaseContext(Context @base)
...
+        {
             ViewModel?.ViewCreated();
+
+            await StartSetup();
+
             InitializeForms(bundle);
         }
 
-        public virtual void InitializeForms(Bundle bundle)
+        protected virtual async Task StartSetup()
```
nmilcoff: Maybe this method should be called something like `RunAppStart`, the `Setup` keyword is a bit confusing in this context

MvvmCross.Forms/Platform/Android/Views/MvxFormsApplicationActivity.cs
```diff
@@ -48,6 +49,8 @@ public IMvxViewModel ViewModel
             }
         }
 
+        private MvxNotifyTask _startTaskNotifier;
```
nmilcoff: This member isn't used anywhere.

MvvmCross/ViewModels/MvxNotifyTask.cs
nickrandolph: Two things:
- my understanding is that calling Yield will cause an unnecessary latency in the task being invoked
- if it is there for a reason (eg best practice or some other guidance) we should add it back with an appropriate comment and/or reference

MvvmCross.Forms/Platform/Android/Views/MvxFormsAppCompatActivity.cs
nickrandolph: Good call. Will fix

MvvmCross.Forms/Platform/Android/Views/MvxFormsApplicationActivity.cs
nickrandolph: Thanks. Will clean it up now that it’s no longer used

MvvmCross/ViewModels/MvxNotifyTask.cs
nickrandolph: Is the point of the Yield call to force the code to run asynchronously? If so, this should be an option that's exposed when creating the MvxNotifyTask, rather than being the default.

nickrandolph:
@nmilcoff The point of these changes are to ensure that exceptions aren't lost deep within mvx where developers can't intercept them. We also want to give developers a known point where they can run code after the first navigation has completed. Both of these scenarios are difficult to accomplish as the Start method is async void with no way of capturing exceptions, nor waiting for it to complete.
The ideal solution is to change to StartAsync but that's a much larger piece of work. I'm looking at this as an interim solution with a view to switching to StartAsync in a future revision.

MvvmCross/Binding/Binders/MvxNamedInstanceRegistryFiller.cs
```diff
@@ -82,6 +82,8 @@ where type.IsConventional()
             {
                 try
                 {
+                    if (pair.Type.ContainsGenericParameters) continue;
```
martijn00: What does this do?

MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -14,21 +16,31 @@ public class MvxAppStart<TViewModel>
...
+            // Check whether Start has commenced, and return if it has
+            if (Interlocked.CompareExchange(ref startHasCommenced, 1, 0) == 1) return;
+
             if (hint != null) {
                 MvxLog.Instance.Trace("Hint ignored in default MvxAppStart");
             }
+            
             try {
-                await NavigationService.Navigate<TViewModel>();
+                NavigationService.Navigate<TViewModel>().Wait();
```
martijn00: I think using GetAwaiter().GetResult() is prefered over .Wait()

Projects/Playground/Playground.Forms.iOS/AppDelegate.cs
```diff
@@ -11,22 +12,15 @@ public partial class AppDelegate : MvxFormsApplicationDelegate
...
-            startup.Start();
-
-            LoadApplication(setup.FormsApplication);
-
-            Window.MakeKeyAndVisible();
-
             return base.FinishedLaunching(app, options);
         }
+
+        protected override MvxIosSetup CreateSetup(IMvxApplicationDelegate applicationDelegate, UIWindow window)
```
martijn00: We should probably move this into the normal appdelegate as well.

MvvmCross.Forms/Platform/Ios/Core/MvxFormsApplicationDelegate.cs
```diff
@@ -31,11 +33,45 @@ public override UIWindow Window
 
         public override bool FinishedLaunching(UIApplication uiApplication, NSDictionary launchOptions)
         {
+            Window = new UIWindow(UIScreen.MainScreen.Bounds);
+
+            UINavigationBar.Appearance.BarTintColor = UIColor.FromRGB(65, 105, 225);
```
martijn00: Should this be in here?

MvvmCross.Forms/Platform/Android/Views/MvxFormsApplicationActivity.cs
```diff
@@ -92,23 +91,38 @@ protected override void AttachBaseContext(Context @base)
         }
 
         protected override void OnCreate(Bundle bundle)
+        {
+            base.OnCreate(bundle);
+
+            InternalOnCreate(bundle);
```
martijn00: InternalOnCreate sounds a bit weird as name

MvvmCross/Binding/Binders/MvxNamedInstanceRegistryFiller.cs
nickrandolph: It prevents exceptions being raised when the Type is a generic - In the playground forms droid sample it is throwing an exception.

MvvmCross/ViewModels/MvxAppStart.cs
nickrandolph: Done

Projects/Playground/Playground.Forms.iOS/AppDelegate.cs
nickrandolph: Done

MvvmCross.Forms/Platform/Ios/Core/MvxFormsApplicationDelegate.cs
nickrandolph: no - I've removed these lines

MvvmCross/ViewModels/MvxNotifyTask.cs
nmilcoff: It was added here #2642 (also I don't see any mention to this change on this PR other than the change in the source code).

You're right about why is it there. The problem is that if the code isn't run asynchronously, there is a (quite high) chance the exception will be thrown before MvxNotifyTask starts monitoring the task. This is highly unexpected when you are delegating all error handling responsibilities to the object.

I agree this adds a (very) small latency to all operations running inside of it, but MvxNotifyTask is not just common async/await.

MvvmCross.Forms/Platform/Android/Views/MvxFormsApplicationActivity.cs
```diff
@@ -82,8 +82,7 @@ protected virtual void OnViewModelSet()
 
         protected override void AttachBaseContext(Context @base)
         {
-            if (this is IMvxAndroidSplashScreenActivity)
-            {
+            if (this is IMvxAndroidSplashScreenActivity) {
```
nmilcoff: `{` should be placed in a new line

MvvmCross.Forms/Platform/Android/Core/MvxFormsAndroidSetup.cs
```diff
@@ -54,13 +55,15 @@ public Application FormsApplication
...
+                    var activity = Mvx.Resolve<IMvxAndroidCurrentTopActivity>()?.Activity ?? ApplicationContext;
                     Xamarin.Forms.Forms.Init(activity, null, activity.GetType().Assembly);
                 }
-
                 if (_formsApplication == null)
                 {
-                    _formsApplication = _formsApplication ?? CreateFormsApplication();
+                    _formsApplication = CreateFormsApplication();
+                }
+                if (Application.Current != _formsApplication) {
```
nmilcoff: `{` should be placed in a new line

MvvmCross.Forms/Platform/Android/Views/MvxFormsAppCompatActivity.cs
```diff
@@ -84,8 +84,7 @@ public override void SetContentView(int layoutResId)
 
         protected override void AttachBaseContext(Context @base)
         {
-            if (this is IMvxAndroidSplashScreenActivity)
-            {
+            if (this is IMvxAndroidSplashScreenActivity) {
```
nmilcoff: `{` should be placed in a new line

MvvmCross.Forms/Platform/Android/Views/MvxFormsApplicationActivity.cs
```diff
@@ -53,8 +54,7 @@ protected Application FormsApplication
         {
             get
             {
-                if (_formsApplication == null)
-                {
+                if (_formsApplication == null) {
```
nmilcoff: `{` should be placed in a new line

MvvmCross.Forms/Platform/Android/Views/MvxFormsApplicationActivity.cs
```diff
@@ -99,16 +98,34 @@ protected override void OnCreate(Bundle bundle)
...
+        protected virtual object GetAppStartHint(object hint = null)
+        {
+            return null;
+        }
+
         public virtual void InitializeForms(Bundle bundle)
         {
-            if (FormsApplication.MainPage != null)
-            {
+            if (!Xamarin.Forms.Forms.IsInitialized) {
```
nmilcoff: `{` should be placed in a new line

MvvmCross.Forms/Platform/Android/Views/MvxFormsApplicationActivity.cs
```diff
@@ -99,16 +98,34 @@ protected override void OnCreate(Bundle bundle)
...
                 global::Xamarin.Forms.Forms.Init(this, bundle, GetResourceAssembly());
-                LoadApplication(FormsApplication);
             }
+
+            if (Xamarin.Forms.Application.Current != FormsApplication) {
```
nmilcoff: `{` should be placed in a new line

MvvmCross.Forms/Platform/Ios/Core/MvxFormsApplicationDelegate.cs
```diff
@@ -3,17 +3,31 @@
...
 using UIKit;
 using Xamarin.Forms.Platform.iOS;
 
 namespace MvvmCross.Forms.Platform.Ios.Core
 {
-    public class MvxFormsApplicationDelegate : FormsApplicationDelegate, IMvxApplicationDelegate
+    public abstract class MvxFormsApplicationDelegate : FormsApplicationDelegate, IMvxApplicationDelegate
     {
+        private MvxFormsIosSetup _setup;
+        protected MvxFormsIosSetup Setup { get
```
nmilcoff: `{` and `get` should be placed in new lines

MvvmCross.Forms/Platform/Ios/Core/MvxFormsIosSetup.cs
```diff
@@ -60,7 +60,10 @@ public Application FormsApplication
                     Xamarin.Forms.Forms.Init();
                 if (_formsApplication == null)
                 {
-                    _formsApplication = _formsApplication ?? CreateFormsApplication();
+                    _formsApplication = CreateFormsApplication();
+                }
+                if (Application.Current != _formsApplication) {
```
nmilcoff: `{` should be placed in a new line

MvvmCross.Forms/Platform/Mac/Core/MvxFormsMacSetup.cs
```diff
@@ -60,7 +60,10 @@ public Application FormsApplication
                     Xamarin.Forms.Forms.Init();
                 if (_formsApplication == null)
                 {
-                    _formsApplication = _formsApplication ?? CreateFormsApplication();
+                    _formsApplication = CreateFormsApplication();
+                }
+                if (Application.Current != _formsApplication) {
```
nmilcoff: `{` should be placed in a new line

MvvmCross.Forms/Platform/Uap/Core/MvxFormsWindowsSetup.cs
```diff
@@ -52,10 +52,15 @@ public Application FormsApplication
...
+                if (!Xamarin.Forms.Forms.IsInitialized)
+                    Xamarin.Forms.Forms.Init(ActivationArguments, GetViewAssemblies());
+
                 if (_formsApplication == null)
                 {
-                    Xamarin.Forms.Forms.Init(ActivationArguments, GetViewAssemblies());
-                    _formsApplication = _formsApplication ?? CreateFormsApplication();
+                    _formsApplication = CreateFormsApplication();
+                }
+                if(Application.Current != _formsApplication) {
```
nmilcoff: `{` should be placed in a new line

MvvmCross/Core/MvxSetupExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+using MvvmCross.IoC;
+
+namespace MvvmCross.Core
+{
+    public static class MvxSetupExtensions
+    {
+        public static TSetup CreateSetup<TSetup>(Assembly assembly, params object[] parameters) where TSetup : MvxSetup
+        {
+            var setupType = FindSetupType<TSetup>(assembly);
+            if (setupType == null) {
```
nmilcoff: `{` should be placed in a new line

MvvmCross/Core/MvxSetupExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+                throw new MvxException("Could not find a Setup class for application");
+            }
+
+            try {
```
nmilcoff: `{` should be placed in a new line

MvvmCross/Platform/Uap/Views/MvxApplication.cs
```diff
@@ -48,33 +61,30 @@ protected override void OnLaunched(LaunchActivatedEventArgs activationArgs)
...
         }
 
-        protected abstract MvxWindowsSetup CreateSetup(Frame rootFrame, IActivatedEventArgs activationArgs, string suspension);
-
-        protected virtual void StartSetup(Frame rootFrame, IActivatedEventArgs activationArgs)
+        protected virtual void RunAppStart(IActivatedEventArgs activationArgs)
         {
-            if (rootFrame.Content == null) {
-                Setup = CreateSetup(rootFrame, activationArgs, nameof(Suspend));
+            if (RootFrame.Content == null) {
```
nmilcoff: `{` should be placed in a new line

MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -14,21 +16,31 @@ public class MvxAppStart<TViewModel>
...
 
-        public async void Start(object hint = null)
+
+
+        public void Start(object hint = null)
         {
+            // Check whether Start has commenced, and return if it has
+            if (Interlocked.CompareExchange(ref startHasCommenced, 1, 0) == 1) return;
+
             if (hint != null) {
```
nmilcoff: `{` should be placed in a new line

MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -14,21 +16,31 @@ public class MvxAppStart<TViewModel>
...
                 MvxLog.Instance.Trace("Hint ignored in default MvxAppStart");
             }
+            
             try {
```
nmilcoff: `{` should be placed in a new line

MvvmCross/Platform/Ios/Core/MvxApplicationDelegate.cs
```diff
@@ -63,5 +61,30 @@ private void FireLifetimeChanged(MvxLifetimeEvent which)
         }
 
         public event EventHandler<MvxLifetimeEventArgs> LifetimeChanged;
+
+        protected virtual void CreateSetup(IMvxApplicationDelegate applicationDelegate, UIWindow window)
+        {
+            var setupType = FindSetupType();
```
nickrandolph: We should not be doing this - reflection is slow. We can have this as a fallback but the need the ability for the setup class to be specified (see UWP)

Projects/Playground/Playground.Forms.iOS/AppDelegate.cs
```diff
@@ -17,10 +17,5 @@ public override bool FinishedLaunching(UIApplication app, NSDictionary options)
 
             return base.FinishedLaunching(app, options);
         }
-
-        protected override MvxIosSetup CreateSetup(IMvxApplicationDelegate applicationDelegate, UIWindow window)
```
nickrandolph: I think this should remain, or find an alternative way for the application to define the setup class

martijn00: It has just gone from `abstract` to `virtual`. This makes it so much easier for beginners to start on Mvx.

MvvmCross/Platform/Ios/Core/MvxApplicationDelegate.cs
martijn00: This is outdated code. Look at the whole PR.

nickrandolph:
For the most part Setup was treated as a singleton from what I can see. Particularly on Android where it was wrapped in a singleton class to call initialise. I should have sometime over the weekend if any further changes are needed

nickrandolph:
@martijn00 do you think this PR is ready, or are there other adjustments to be made? It'd be great to get this merged so we can get a beta out with these changes. I'll then start to work through the issues to see what else we can get resolved before v6 is done

MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatSetup.cs
```diff
@@ -50,4 +46,15 @@ protected override void FillBindingNames(IMvxBindingNameRegistry registry)
             base.FillBindingNames(registry);
         }
     }
+
+    public abstract class MvxAppCompatSetup<TApplication> : MvxAppCompatSetup
+        where TApplication : IMvxApplication, new()
+    {
+        protected override IMvxApplication CreateApp() => Mvx.IocConstruct<TApplication>();
```
nmilcoff: @nickrandolph are you sure this will work? At first glance looks like [this](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Mvx.cs#L157) won't be ready at this point

MvvmCross.Forms/Platform/Mac/Core/MvxFormsApplicationDelegate.cs
```diff
@@ -3,22 +3,61 @@
...
     public abstract class MvxFormsApplicationDelegate : FormsApplicationDelegate, IMvxApplicationDelegate
     {
+        private MvxFormsMacSetup _setup;
+        protected MvxFormsMacSetup Setup
+        {
+            get
+            {
+                if (_setup == null)
+                    _setup = CreateSetup(this, MainWindow);
+                return _setup;
```
nmilcoff: Can we make Mac work the same way as Android and iOS? (use a `InitializePlatform` method instead of ctor arguments) It seems cleaner to me to unify all setup changes on the same PR

MvvmCross/Platform/Tvos/Core/MvxApplicationDelegate.cs
```diff
@@ -3,14 +3,47 @@
...
-                                          , IMvxApplicationDelegate
+    public abstract class MvxApplicationDelegate : UIApplicationDelegate, IMvxApplicationDelegate
     {
+        private MvxTvosSetup _setup;
+        protected MvxTvosSetup Setup
+        {
+            get
+            {
+                if (_setup == null)
+                    _setup = CreateSetup(this, Window);
```
nmilcoff: Can we make tvOS work the same way as Android and iOS? (use a `InitializePlatform` method instead of ctor arguments) It seems cleaner to me to unify all setup changes on the same PR

MvvmCross/Platform/Wpf/Views/MvxApplication.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public abstract class MvxApplication : Application
+    {
+        private MvxWpfSetup _setup;
+        protected MvxWpfSetup Setup
+        {
+            get
+            {
+                if (_setup == null)
+                    _setup = CreateSetup(Dispatcher, MainWindow);
```
nmilcoff: Can we make WPF work the same way as Android and iOS? (use a `InitializePlatform` method instead of ctor arguments) It seems cleaner to me to unify all setup changes on the same PR

MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatSetup.cs
nickrandolph: This ran up ok for me on Android - this is late executed, so the IoC container has initialized by the time this is called.

MvvmCross.Forms/Platform/Mac/Core/MvxFormsApplicationDelegate.cs
nickrandolph: Done - I've made the changes but can someone please test both Mac and Mac Forms as I don't have access to a Mac this weekend.

MvvmCross/Platform/Tvos/Core/MvxApplicationDelegate.cs
nickrandolph: Done - as per mac, can someone please test to make sure my changes work

nickrandolph:
Ok, I think I've covered all the platforms with the updated setup structure. I'd be great if someone could test the iOS, iOS Forms, Mac, Mac Forms, TvOS builds

martijn00:
This is a huge improvement for new developers starting on MvvmCross! We might need to fix up some more Activity's in Android but for now this will be good to get merged!

PR #2675, opened 2018-03-10
Move hints and attributes into the right folder

martijn00:

Cleanup

Not the same folder used as the other platforms

Yes a bit.

- [ ] All projects build


MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs
```diff
@@ -202,6 +203,15 @@ public override void RegisterAttributeTypes()
                 }
                 return;
             }
+            if (hint is MvxPopRecursivePresentationHint popRecursiveHint) {
+                var levels = popRecursiveHint.LevelsDeep;
+                if (levels > navigation.NavigationStack.Count())
+                    levels = navigation.NavigationStack.Count();
```
nickrandolph: @martijn00 shouldn't this be count-1. If levels==Count, won't it attempt to pop off all pages?

martijn00: Yes it will pop it all off. It's up to the user if they want that, and to control it.

nickrandolph: In that case should we be doing the check for levels > count at all? I think popping the last page fails doesn't it?

martijn00: I'm not sure if popping the last page fails. Popping more than that should not work in any case i guess.

PR #2717, opened 2018-03-23
allow only one fragment instatnce and example in playground

### LINKED ISSUE #2694: Add the ability to only use one instance of a given fragment
It would be great if the `MvxFragmentPresentation` attribute had a flag to tell the presenter to only allow one instance of the fragment at a time. An example scenario where this feature would be useful:

1. A `HomeViewModel` is navigated to when the program starts.
2. The user opens the hamburger menu and clicks the Home item to display `HomeViewModel`.
3. Now there are two instances of the `HomeViewModel` when it would be desirable to only have the one instance active at a time.

### Configuration :wrench:

**Version:** 6.x

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

orzech85:
feature

There is no a flag to tell the presenter to only allow one instance of the fragment at a time.

`allowOnlyOneInstance` flag added to `MvxFragmentPresentationAttribute`

no

SplitView (master/detail) in Playground has an example

fixes #2694 

- [x] All projects build

### DIFF
--- a/MvvmCross/Platforms/Android/Presenters/Attributes/MvxFragmentPresentationAttribute.cs
+++ b/MvvmCross/Platforms/Android/Presenters/Attributes/MvxFragmentPresentationAttribute.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MS-PL license.
 // See the LICENSE file in the project root for more information.
 
@@ -27,7 +27,8 @@ namespace MvvmCross.Platforms.Android.Presenters.Attributes
             int popExitAnimation = int.MinValue,
             int transitionStyle = int.MinValue,
             Type fragmentHostViewType = null,
-            bool isCacheableFragment = false
+            bool isCacheableFragment = false,
+            bool allowOnlyOneInstance = false
         )
         {
             ActivityHostViewModelType = activityHostViewModelType;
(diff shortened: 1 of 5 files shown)


MvvmCross/Platforms/Android/Presenters/Attributes/MvxFragmentPresentationAttribute.cs
```diff
@@ -90,6 +94,12 @@ public MvxFragmentPresentationAttribute()
         /// </summary>
         public bool AddToBackStack { get; set; } = DefaultAddToBackStack;
 
+        public static bool DefaultAllowOnlyOneInstance = false;
+        /// <summary>
+        /// Allow only one instance at a time. False by default.
+        /// </summary>
+        public bool AllowOnlyOneInstance { get; set; } = DefaultAllowOnlyOneInstance;
```
martijn00: I think we might need to rethink this a bit. IsCacheableFragment should already indicate if it is a single instance or reused. Because this problem exists on all platforms i was thinking to make 2 new properties in a base class:

```
public bool CacheView = false;
public bool CacheViewModel = false;
```
We could then implement this over all the platforms and make a stable way to cache views and viewmodels.

nickrandolph: @martijn00 which class would these properties reside?

martijn00:
We need to make a broader proposal for this. Closing it for now.

PR #2720, opened 2018-03-23
Helpers for testing RaiseCanExecuteChanged

jacobduijzer:
Adds a helper and extension methods to make it possible to unit test the behaviour of RaiseCanExecuteChanged. 

Currently it is not possible to check (in a unit test) if CanExecuteChanged has been raised.

With the use of the helper (which is also documented in the Testing chapter of the MvvmCross documentation) it is possible to assert that CanExecuteChanged has been raised.

No

- [x ] All projects build
- [x ] Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))
- [x ] Rebased onto current develop

### DIFF
--- a/MvvmCross.Tests/MvxUnitTestCommandHelper.cs
+++ b/MvvmCross.Tests/MvxUnitTestCommandHelper.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using MvvmCross.Commands;
 
 namespace MvvmCross.Tests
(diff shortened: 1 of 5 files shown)


Cheesebaron:
How about also adding some actual tests?

jacobduijzer:
To test the code or to show how it works?

MvvmCross.Tests/MvxUnitTestCommandHelper.cs
```diff
@@ -0,0 +1,46 @@
...
+        public void WillCallRaisePropertyChangedFor(object item)
+        {
+            items.Add(item, 0);
+        }
+
+        public bool HasCalledRaisePropertyChangedFor(object item)
+        {
+            try
+            {
+                return items[item] > 0;
```
nickrandolph: There should be protection in here for when the item doesn't exist, rather than just allowing it to throw an exception

nickrandolph:
@jacobduijzer I agree with @Cheesebaron it's be good to actually include some tests to show it can be used. @martijn00 is it worth adding a tests project to the playground sample?

jacobduijzer:
I will take care of it!

jacobduijzer:
I added Unit tests for the helper but it's not really clear where I should
add an example.Wouldn't updating the documentation as I already did be
sufficient?

PR #2731, opened 2018-03-26
Setup fixes

martijn00:

- [x] All projects build

### DIFF
--- a/MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatActivity.cs
+++ b/MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatActivity.cs
@@ -129,7 +129,7 @@ namespace MvvmCross.Droid.Support.V7.AppCompat
 
         protected virtual void RegisterSetup<TMvxAndroidSetup>() where TMvxAndroidSetup : MvxAppCompatSetup, new()
         {
-            MvxSetup.RegisterSetupType<TMvxAndroidSetup>(GetType().Assembly);
+            this.RegisterSetupType<TMvxAndroidSetup>();
         }
     }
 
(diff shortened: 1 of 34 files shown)


MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatActivity.cs
```diff
@@ -87,6 +87,8 @@ protected override void OnCreate(Bundle bundle)
         {
             base.OnCreate(bundle);
             ViewModel?.ViewCreated();
+
+            //TODO: Add setup init and startup here
```
nmilcoff: What does this comment mean here? Is that needed to enable android apps start without splash screens, right?

nickrandolph: @martijn00 did you need me to add the setup logic, or are you going to add this before we merge this PR?

martijn00: Yeah it's pretty late here, would be good if you have time. Should be a matter of copying some logic in here and it should work.

nickrandolph:
@martijn00 I did a bit of refactoring of setup logic across some of the platform classes. However, I'm not sure whether we need to add RegisterSetupType to MvxFragmentActivity and MvxTabActivity. We only need to add RegisterSetup to any class that acts as an entry point for a specific platform (eg App class for UWP, AppDelegate for iOS, Application for Android).  I guess the difference is that Android can be launched with any Activity, (or Fragment??) - does this mean we should add RegisterSetup to every one of these? 

The RegisterSetup is really just a helper to make it easier for developers so that they don't need to register their Setup class. My thoughts are that if a developer is launching their application using anything other than the generic MvxAppCompatApplication or generic MvxAndroidApplication they should either implement the RegisterSetup method themselves, or call the new RegisterSetupType extension method from their platform class.

I've updated this PR by removing all RegisterSetupType methods, other than the extension method. If I've missed a scenario that you think is hard for developers to get started, let me know and I can come up with a solution.

PR #2732, opened 2018-03-27
MvxAppStart: Add non-generic version

nmilcoff:
Code improvement

Since all the Setup infrastructure has changed, not there's some code that needs to be placed in AppStart to ensure everything works fine.
The problem is that if an app implements its own MvxAppStart, that code needs to be copied/pasted or adequated on the app, because there's no base class to extend from.

Above is fixed. I added a non-generic version of MvxAppStart and applied inheritance.

It depends... but I'd say yes (although it's not this PR which breaks, this was caused by the PR which improved all the Setup stuff).
I needed to add a new method, which is wrapped in the classic `Start`. 

Apps upgrading to 6.0 will need to change the inheritance of their custom AppStart to be `MvxAppStart` instead of `IMvxAppStart` and also modify `Start` to `Startup`.

/

- [ ] All projects build

### DIFF
--- a/MvvmCross/ViewModels/MvxAppStart.cs
+++ b/MvvmCross/ViewModels/MvxAppStart.cs
@@ -3,25 +3,15 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Threading;
-using System.Threading.Tasks;
 using MvvmCross.Exceptions;
 using MvvmCross.Logging;
 using MvvmCross.Navigation;
 
 namespace MvvmCross.ViewModels
 {
-    public class MvxAppStart<TViewModel>
-        : IMvxAppStart
-        where TViewModel : IMvxViewModel
+    public abstract class MvxAppStart : IMvxAppStart
     {
-        protected readonly IMvxNavigationService NavigationService;
-
         private int startHasCommenced;
-
-        public MvxAppStart(IMvxNavigationService navigationService)
-        {
-            NavigationService = navigationService;
-        }
         
         public void Start(object hint = null)
         {
@@ -29,6 +19,31 @@ namespace MvvmCross.ViewModels
             if (Interlocked.CompareExchange(ref startHasCommenced, 1, 0) == 1)
                 return;
 
+            Startup(hint);
+        }
+
+        protected abstract void Startup(object hint = null);
+
+        public virtual bool IsStarted => startHasCommenced != 0;
+
+        public virtual void ResetStart()
+        {
+            Interlocked.Exchange(ref startHasCommenced, 0);
+        }
+    }
+
+    public class MvxAppStart<TViewModel> : MvxAppStart
+        where TViewModel : IMvxViewModel
+    {
+        protected readonly IMvxNavigationService NavigationService;
+        
+        public MvxAppStart(IMvxNavigationService navigationService)
+        {
+            NavigationService = navigationService;
+        }
+
+        protected override void Startup(object hint = null)
+        {
             if (hint != null)
             {
                 MvxLog.Instance.Trace("Hint ignored in default MvxAppStart");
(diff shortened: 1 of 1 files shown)


MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -3,32 +3,47 @@
...
 using MvvmCross.Navigation;
 
 namespace MvvmCross.ViewModels
 {
-    public class MvxAppStart<TViewModel>
-        : IMvxAppStart
-        where TViewModel : IMvxViewModel
+    public abstract class MvxAppStart : IMvxAppStart
     {
-        protected readonly IMvxNavigationService NavigationService;
```
nickrandolph: @nmilcoff Love this refactor. Is there a reason why the navigation service can't be pushed down to the non-generic version?

nmilcoff: Thanks! Yeah, so the problem is that there isn't enough information to use it... the generic version adds the ViewModel the app wants to navigate to. 
The non generic version will allow apps to take the decision without compromising stability.

nickrandolph: But won't the MvxAppStart be kinda useless if they don't have access to the NavigationService. Instead of having developers write Mvx.Resolve in order to get it, we might as well include it in the non-generic version so it can be injected into the constructor. We can still leave the Startupup method as abstract, so it shouldn't affect stability

MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -3,32 +3,47 @@
...
 
 using System.Threading;
-using System.Threading.Tasks;
 using MvvmCross.Exceptions;
 using MvvmCross.Logging;
 using MvvmCross.Navigation;
 
 namespace MvvmCross.ViewModels
 {
-    public class MvxAppStart<TViewModel>
```
nickrandolph: Whilst we're adjusting this, is it worth creating a generic version of IMvxAppStart eg IMvxAppStart<TViewModel> : IMvxAppStart?

MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -3,32 +3,47 @@
...
         public void Start(object hint = null)
         {
             // Check whether Start has commenced, and return if it has
             if (Interlocked.CompareExchange(ref startHasCommenced, 1, 0) == 1)
                 return;
 
+            Startup(hint);
+        }
+
+        protected abstract void Startup(object hint = null);
```
martijn00: This would break all apps that use it currently. Maybe make the first method Startup, and the abstract one Start so it stays the same?

MvvmCross/ViewModels/MvxAppStart.cs
nmilcoff: That should be up to the developer, who can easily inject the NavigationService on its own implementation. I don't see a reason to include a dependency on a class that won't be used.

Please look at my implementation for the [StarWarsSample](https://github.com/MvvmCross/MvvmCross-Samples/pull/114/files#diff-56bd2bea371da509b91bfb9c38d72189R7). Because there's no "base" class to inherit from, I had to include some logic for `IsStarted` by myself.

MvvmCross/ViewModels/MvxAppStart.cs
nmilcoff: Sorry but I don't get this one 🤔

MvvmCross/ViewModels/MvxAppStart.cs
nmilcoff: Indeed. But the breaking change was caused the Setup refactor... As the way AppStart is called has changed, there's now some necessary code to make it thread safe, and that code was only on the generic version of it. 

Same as the [StarWarsSample](https://github.com/MvvmCross/MvvmCross-Samples/pull/114/files#diff-56bd2bea371da509b91bfb9c38d72189R7), most complex apps do have a custom AppStart and therefore I think it's safer to break stuff and make developers inherit their custom AppStarts from `MvxAppStart` instead of `IMvxAppStart`

Regarding the naming... `Start` is now a "wrapper" call for `Startup`, which contains all the thread safe code verification. As the method is public and can be called from literally anywhere, and we want the code to be always thread safe, I made Startup `protected` and left Start public.

PR #2733, opened 2018-03-28
Update nuget packages

martijn00:

- [ ] All projects build


Cheesebaron:
It is almost as if we should just include this per default in build.props for Android.

martijn00:
@Cheesebaron yeah i was thinking of that, and then mark it as private. But it would be good for others to depend on it too. And we don't need this in the main lib.

PR #2735, opened 2018-03-29
Added generic versions of MvxApplication for WPF and UWP

nmilcoff:
Probably a feature, or a fix?

Classes with generic versions on most platforms. But UWP and WPF are missing

Above is fixed.

No

/

Previous PR: #2668 

- [X] All projects build

### DIFF
--- a/MvvmCross.Forms/Platforms/Uap/Views/MvxWindowsApplication.cs
+++ b/MvvmCross.Forms/Platforms/Uap/Views/MvxWindowsApplication.cs
@@ -4,11 +4,14 @@
 
 using System;
 using System.Threading.Tasks;
+using MvvmCross.Core;
+using MvvmCross.Forms.Platforms.Uap.Core;
 using MvvmCross.Forms.Views.Base;
 using MvvmCross.Platforms.Uap.Core;
 using MvvmCross.Platforms.Uap.Views;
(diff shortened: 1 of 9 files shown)


nickrandolph:
Interesting - I could have sworn that I tried this and had issues with Visual Studio not building with a generic Application class. Can you make sure resources defined in App.xaml can be loaded correctly - I won't get to it tonight but can add to this PR tomorrow morning if you want

nickrandolph:
Also, if this works, we should be able to do the same for the XF UWP application class too

nmilcoff:
Sure, I'll test that later today. I'm not really a UWP/WPF expert so any help here will be much appreciated @nickrandolph 🙂

nickrandolph:
As I suspected this gives a build error for the Playground.Uwp project:
7>C:\Source\github\btr-mvx\Projects\Playground\Playground.Uwp\App.xaml.cs(7,33,7,36): error CS0263: Partial declarations of 'App' must not specify different base classes
Resolving this would require changing the App.xaml to use the generic base class but that's not supported in UWP (XF has generics support in XAML but not UWP currently).
I've pushed one option for fixing this. 
The other (much less desirable) is to remove the App.xaml. This is fine, until you want to register application wide static resources... I would not recommend this approach

PR #2713, opened 2018-03-21
Refactoring Setup Singleton 

nickrandolph:
### :sparkles: What kind of change does this PR introduce? (Bug fix, feature, docs update...)
Refactor

### :arrow_heading_down: What is the current behavior?
Only Android uses a setup singleton wrapper

### :new: What is the new behavior (if this is a feature change)?
setup singleton can be reused for all platforms

### :boom: Does this PR introduce a breaking change?
No

### :thinking: Checklist before submitting

- [ ] All projects build
- [ ] Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))
- [ ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contribute/mvvmcross-docs-style-guide))
- [ X ] Rebased onto current develop

PR #2746, opened 2018-03-30
MvxSetupSingleton optimizations / Fix SplashScreen initialization on Android

nmilcoff:
Bug fix

1. Starting from SplashScreen doesn't work on Android
2. Setup is called multiple times during Android startup

Above is fixed, also made some optimizations and added some reference comments to MvxSetupSingleton.

Yes. TriggerFirstNavigate was removed.

Test the Playground apps

PR: #2713 

- [X] All projects build

### DIFF
--- a/MvvmCross.Android.Support/V7.AppCompat/MvxSplashScreenAppCompatActivity.cs
+++ b/MvvmCross.Android.Support/V7.AppCompat/MvxSplashScreenAppCompatActivity.cs
@@ -7,7 +7,6 @@ using Android.Runtime;
 using Android.Views;
 using MvvmCross.Core;
 using MvvmCross.Platforms.Android.Core;
-using MvvmCross.Platforms.Android.Views;
 using MvvmCross.ViewModels;
 
 namespace MvvmCross.Droid.Support.V7.AppCompat
@@ -19,6 +18,8 @@ namespace MvvmCross.Droid.Support.V7.AppCompat
         private const int NoContent = 0;
 
         private readonly int _resourceId;
+
+        private Bundle _bundle;
 
         public new MvxNullViewModel ViewModel
         {
@@ -39,6 +40,8 @@ namespace MvvmCross.Droid.Support.V7.AppCompat
         protected override void OnCreate(Bundle bundle)
         {
             RequestWindowFeatures();
+
+            _bundle = bundle;
 
             var setup = MvxAndroidSetupSingleton.EnsureSingletonAvailable(ApplicationContext);
             setup.InitializeAndMonitor(this);
@@ -77,14 +80,7 @@ namespace MvvmCross.Droid.Support.V7.AppCompat
             if (!_isResumed)
                 return;
 
-            TriggerFirstNavigate();
-        }
-
-        protected virtual void TriggerFirstNavigate()
-        {
-            var startup = Mvx.Resolve<IMvxAppStart>();
-            if (!startup.IsStarted)
-                startup.Start();
+            RunAppStart(_bundle);
         }
     }
 }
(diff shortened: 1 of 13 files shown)


nmilcoff:
@kjeremy 👍 ! It's splitted now

Projects/Playground/Playground.Droid/SplashScreen.cs
```diff
@@ -15,11 +15,16 @@ namespace Playground.Droid
...
         , ScreenOrientation = ScreenOrientation.Portrait)]
-    public class SplashScreen : MvxSplashScreenActivity
+    public class SplashScreen : MvxSplashScreenAppCompatActivity
     {
         public SplashScreen()
             : base(Resource.Layout.SplashScreen)
         {
         }
+
+        protected override void OnResume()
```
nickrandolph: Does this override need to be here?

MvvmCross/Core/MvxSetupSingleton.cs
```diff
@@ -61,57 +55,60 @@ public virtual void EnsureInitialized()
...
+                    if (_currentMonitor == null)
+                    {
+                        IsInitialisedTaskCompletionSource.Task.GetAwaiter().GetResult();
+                        MvxLog.Instance.Trace("EnsureInitialized has already been called so now waiting for completion");
+                    }
                 }
                 else
                 {
-                    MvxLog.Instance.Trace("EnsureInitialized has already been called so now waiting for completion");
+                    StartSetupInitialization();
```
nickrandolph: This beaks the behaviour of EnsureInitialized which should block until setup is completed. Currently this PR breaks more than it fixes.

MvvmCross.Android.Support/V7.AppCompat/MvxSplashScreenAppCompatActivity.cs
```diff
@@ -86,5 +85,10 @@ protected virtual void TriggerFirstNavigate()
             if (!startup.IsStarted)
                 startup.Start();
         }
+
+        // do nothing on this override, as initial navigation is managed by TriggerFirstNavigate
```
martijn00: Should we remove TriggerFirstNavigate?

MvvmCross/Core/MvxSetupSingleton.cs
nmilcoff: Sounds reasonable. We need to test PRs like #2713 or this one more before merging...

nmilcoff:
Ok guys, thank you all for your feedback! Honestly I feel like this PR is actually fixing some issues that were introduced just before... We need to be more careful with merging.

You'll notice I removed some calls to EnsureInitialized. This is because all Android Activities are already calling that [here](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Platforms/Android/Views/MvxActivityViewExtensions.cs#L138).

nickrandolph:
I think this is probably one of the more complex parts of mvx to get right. Mainly because each platform has slightly different startup behaviour and we’re trying to coerce into a single pattern to make it easy for devs.
I don’t think we’re quite there as there are still some race conditions with the latest edits but we’re getting there. I think it’s be good to have some tests around this, so I will try and get that started when I have some time.

kjeremy:
A big fat high level comment on how all this works might be good.

nickrandolph:
@martijn00 @nmilcoff I've done another pass through this PR and whilst there are some extreme edge cases where a race condition can occur, they're not easily resolvable without reworking the singleton significantly. I propose that this PR is accepted as is, and that we confirm that all platforms (inc Forms) are working well on startup.

PR #2754, opened 2018-04-02
Reverts IsXamarinForms changes in Playground

tbalcom:
Bug fix

<PackageReference/> to Xamarin.Forms is incorrectly being added to Playground.Droid.

<PackageReference/> to Xamarin.Forms is no longer being added to Playground.Droid.

No.

Rebuild and verify there is no longer a NuGet dependency to Xamarin.Forms in Playground.Droid.

- [x] All projects build

### DIFF
--- a/Directory.build.props
+++ b/Directory.build.props
@@ -35,8 +35,4 @@
   <ItemGroup Condition="'$(IsTestProject)' != 'true' and '$(SourceLinkEnabled)' != 'false' and '$(IsLibraryProject)' == 'true'">
     <PackageReference Include="SourceLink.Create.CommandLine" Version="2.8.0" PrivateAssets="All" /> 
   </ItemGroup>
-
-  <ItemGroup Condition="'$(IsXamarinForms)' == 'true'">
-    <PackageReference Include="Xamarin.Forms" Version="2.5.0.280555" />
-  </ItemGroup>
 </Project>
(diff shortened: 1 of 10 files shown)


nickrandolph:
@tbalcom are you sure that XF was being included in the Playground.Droid project? Looking at the props it looks like it shouldn't be. This PR is a regression on the the changes @martijn00  and I both proposed around simplifying the reference to XF, so I don't think this is a great fix. Better for us to find out why Playground.Droid would be getting a reference to XF.

tbalcom:
I brought it up on Slack this morning and he asked for a PR. You can see the Xamarin.Forms package shows up in Playground.Droid and can't be removed because of the .props file.

nickrandolph:
Ok, I just saw the conversation - reverting this change isn't the way to go. Instead I think we should add a separate targets file that gets included in the appropriate projects (mvvmcross.forms and the playground sample projects). This way we can still easily adjust forms version in one place.

nickrandolph:
FYI - MS is aware of this as an issue (see https://github.com/dotnet/project-system/issues/2129)

nickrandolph:
@tbalcom I hope you don't mind but I pushed changes to your branch that keep the intent of the original commit (central place to adjust XF versioning) whilst overcoming the issue that you quite rightly identified. I've confirmed the issue with the MS product team (see issue linke above), so for the moment we'll have to make do with the slightly less tidy workaround. 
Thanks again for picking this up and raising a PR to get it fixed

PR #2755, opened 2018-04-03
Adding Async dispatcher

nickrandolph:
Feature

The existing dispatcher invokes actions to run on the UI thread but it's fire and forget

The async dispatcher can be used to run an action on the UI thread and await for it to complete

No

Run all playground samples

- [ ] All projects build

### DIFF
--- a/MvvmCross/Base/MvxMainThreadAsyncDispatcher.cs
+++ b/MvvmCross/Base/MvxMainThreadAsyncDispatcher.cs
@@ -1,0 +1,43 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MS-PL license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace MvvmCross.Base
+{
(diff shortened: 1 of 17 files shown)


nickrandolph:
Note that this PR introduces the async dispatcher but doe **not** use it within any of the navigation framework.

MvvmCross/Base/MvxMainThreadAsyncDispatcher.cs
```diff
@@ -0,0 +1,43 @@
...
+            var syncAction = new Action(async () =>
+            {
+                await action();
+                completion.SetResult(true);
+            });
+            RequestMainThreadAction(syncAction, maskExceptions);
+
+            // If we're already on main thread, then the action will
+            // have already completed at this point, so can just return
+            if (completion.Task.IsCompleted)
```
martijn00: Can't we use ValueTask to do this?

Cheesebaron: `TaskCompletionSource` doesn't return a `ValueTask`? What are you proposing?

martijn00: If we use a `ValueTask` we could return the completed status right away if the Task was finished. Not sure if it makes sense to do here.

Issue #2745, opened 2018-03-30
[iOS] Using VS AppCenter "AppCenter.Start" while MvxApplication.Initialize results in deadlock since MVX 6 beta7


Sebastian1989101:
### Steps to reproduce :scroll:

1. Using VS AppCenter

2. Start the AppCenter within the MvxApplication.Initialize overwrite

### Expected behavior :thinking:
No deadlock like it was with MVX 6.0.0 beta6 and before.

### Actual behavior :bug:
Deadlock without any result (App stays on Splash Screen, there is no exception when debugging as soon as you step over AppCenter.Start nothing happens anymore). Is the MvxApplication.Initialize now in a different thread maybe? 

**Note**: This can be avoided by putting the VS AppCenter code somewhere else but maybe this is affecting other stuff too. 

### Configuration :wrench:

**Version:** 6.0.0.beta7

**Platform:** 
- [x] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [x] :monkey: Xamarin.Forms

borbmizzet:
This is a major concern for those of us who use VS App Center for diagnostics.  I was planning on migrating our Vanilla Xamarin.Forms code to MvvmCross v6 as soon as it was released but this regression is a major roadblock.

nmilcoff:
Also initializing AppCenter in the App class is probably too late (and dangerous) if you're using their crash logging feature.

Sebastian1989101:
@Cheesebaron The Xamarin.Forms OnStart() methods seems to be too late when using MVX. The push notifications wont work there. The `MvxApplication.Initialize()` is earlier but it is no longer possible to init VS AppCenter there (app freeze at launch since beta7, works with beta6 and earlier). 

I'm using the `AppDelegateFinishedLaunching()` and the `MvxSplashScreenActivity.RunAppStart()` since beta7 and it seems to work (I just have a few problems with push but I guess this comes from somewhere else). 

I just wanted to mention that some things might result in a freeze since beta7 with the report and not that VS AppCenter does this because you can move the code to another location very easily.

fedemkr:
[...]

PR #2761, opened 2018-04-04
Adding Application Startup method to be invoked on UI thread

nickrandolph:
Bug fix

Initialize is called on background thread which causes issues for any Application init actions that need to be done on UI thread

Startup method invoked on Application on the UI thread

No

https://github.com/MvvmCross/MvvmCross/issues/2745

- [ ] All projects build

### DIFF
--- a/MvvmCross/ViewModels/MvxAppStart.cs
+++ b/MvvmCross/ViewModels/MvxAppStart.cs
@@ -11,7 +11,14 @@ namespace MvvmCross.ViewModels
 {
     public abstract class MvxAppStart : IMvxAppStart
     {
+        protected readonly IMvxApplication Application;
+
         private int startHasCommenced;
+
+        public MvxAppStart(IMvxApplication application)
+        {
+            Application = application;
+        }
 
         public void Start(object hint = null)
         {
@@ -22,7 +29,16 @@ namespace MvvmCross.ViewModels
             Startup(hint);
         }
 
-        protected abstract void Startup(object hint = null);
+        protected virtual void Startup(object hint = null)
+        {
+            ApplicationStartup(hint);
+        }
+
+        protected virtual void ApplicationStartup(object hint = null)
+        {
+            MvxLog.Instance.Trace("AppStart: Application Startup - On UI thread");
+            Application.Startup(hint);
+        }
 
         public virtual bool IsStarted => startHasCommenced != 0;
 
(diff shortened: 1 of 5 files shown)


MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -28,35 +28,60 @@ public void Start(object hint = null)
 
         public virtual void ResetStart()
         {
+            Reset();
             Interlocked.Exchange(ref startHasCommenced, 0);
         }
+
+        protected virtual void Reset()
```
martijn00: Should this be empty?

MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -28,35 +28,60 @@ public void Start(object hint = null)
...
         }
 
         protected override void Startup(object hint = null)
+        {
+            ApplicationStartup();
+
+            NavigateToFirstViewModel(hint);
+        }
+
+        protected virtual void ApplicationStartup(object hint = null)
```
martijn00: Do we need this method? We could just call `Application.Startup(hint);` from `Startup `?

MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -28,35 +28,60 @@ public void Start(object hint = null)
...
+        {
+            MvxLog.Instance.Trace("AppStart: Application Startup - On UI thread");
+            Application.Startup(hint);
+        }
+
+        protected virtual void NavigateToFirstViewModel(object hint)
```
martijn00: Would it make sense to let this return the `StartupViewModel`, and call that from `Startup`?

kjeremy:
Are we still able to start MvvmCross from a background service?

MvvmCross/ViewModels/MvxAppStart.cs
nickrandolph: no longer empty in base mvxappstart class

MvvmCross/ViewModels/MvxAppStart.cs
nickrandolph: I think it makes more sense now after this has been refactored

MvvmCross/ViewModels/MvxAppStart.cs
nickrandolph: Not sure I understand what you're thinking here - take a look at my refactor and let me know

nickrandolph:
@kjeremy I think that should be fine but if you're able to generate a sample of background service and add to playground, that'd be great.

PR #2769, opened 2018-04-06
Make setup interface standard

martijn00:

Feature

MvxSetup interface not standard

Now it is

Yes

- [ ] All projects build

### DIFF
--- a/MvvmCross/Core/MvxSetup.cs
+++ b/MvvmCross/Core/MvxSetup.cs
@@ -207,6 +207,7 @@ namespace MvvmCross.Core
             // initialize the IoC registry, then add it to itself
             var iocProvider = CreateIocProvider();
             Mvx.RegisterSingleton(iocProvider);
+            Mvx.RegisterSingleton<IMvxSetup>(this);
         }
 
         protected virtual IMvxIocOptions CreateIocOptions()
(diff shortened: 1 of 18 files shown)


Projects/Playground/Playground.Wpf/App.xaml.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Playground.Wpf
 {
-    public partial class App : MvxApplication<MvxWpfSetup<Core.App>, Core.App>
+    public partial class App : MvxApplication
```
nickrandolph: Can this not inherit from the generic base?

martijn00: No i couldn't get the generic type arguments to work on WPF.

MvvmCross/Core/MvxSetup.cs
```diff
@@ -207,6 +207,7 @@ protected virtual void InitializeIoC()
             // initialize the IoC registry, then add it to itself
             var iocProvider = CreateIocProvider();
             Mvx.RegisterSingleton(iocProvider);
+            Mvx.RegisterSingleton<IMvxSetup>(this);
```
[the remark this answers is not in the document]
martijn00: Yes, so you can subscribe on this interface to callback when registered, then for example add a call to `StateChanged` that is now in the interface.

kjeremy:
EnsureInitialized was required from background services that dealt with FCM
messages.  Is there a new alternative?

nickrandolph:
@kjeremy yes, the MvxSetupSingleton should still be used imho. However, we don't currently have a sample of a background service for Playground. Any chance you can add one so we can confirm that it works?

PR #2775, opened 2018-04-08
Fixing issue with CurrentActivity being null in Playground.Droid

nickrandolph:
Bug fix

Exception raised when attempting to navigate to Tab or Split view in Playgroun.Droid

Navigation works.

No

Run Playground.Droid

https://github.com/MvvmCross/MvvmCross/issues/2782

- [ ] All projects build

### DIFF
--- a/MvvmCross.Android.Support/Fragment/EventSource/MvxEventSourceFragmentActivity.cs
+++ b/MvvmCross.Android.Support/Fragment/EventSource/MvxEventSourceFragmentActivity.cs
@@ -36,12 +36,6 @@ namespace MvvmCross.Droid.Support.V4.EventSource
         {
             DestroyCalled.Raise(this);
             base.OnDestroy();
-
-            if (IsTaskRoot)
-            {
-                var appStart = Mvx.Resolve<IMvxAppStart>();
-                appStart.ResetStart();
-            }
         }
 
         protected override void OnNewIntent(Intent intent)
--- a/MvvmCross.Android.Support/V7.AppCompat/EventSource/MvxEventSourceAppCompatActivity.cs
+++ b/MvvmCross.Android.Support/V7.AppCompat/EventSource/MvxEventSourceAppCompatActivity.cs
@@ -37,12 +37,6 @@ namespace MvvmCross.Droid.Support.V7.AppCompat.EventSource
         {
             DestroyCalled.Raise(this);
             base.OnDestroy();
-
-            if(IsTaskRoot)
-            {
-                var appStart = Mvx.Resolve<IMvxAppStart>();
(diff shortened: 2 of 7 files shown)


nickrandolph:
@martijn00 I raised this as a WIP because whilst my fix does get the playground.droid sample to work, I'm concerned about the use of CurrentActvity - it appears that in some scenarios this isn't set correctly until after OnCreate has completed, which means any code (such as navigating to tabs or the details view of split) will fail. Is there a better way for us to determine which activity the tabs should be added to (perhaps for beyond v6)?

Plac3hold3r:
@nickrandolph I have experienced similar issues with `CurrentActvity` being different to what previous versions of MvvmCross had. 

I was adding a sample using MvvmCross 6.0.0-beta.7 to my Android MvvmCross [SharedElement](https://github.com/Plac3hold3r/PH.SharedElement) sample repo. I ran into issue where previously (See MvvmCross version 5.x.x samples in repo) I could navigate directly to the ViewModel of a fragment in the AppStart. However, with MvvmCross 6.0.0-beta.7, this was no longer possible. I wanted to have a better look and see why it was happening. But in the mean time what I can share is that looking a the presenter the `CurrentActvity` was always being returned as the SplashScreenActivity, i.e. the activity prior to the one that gets created due to fragment ViewModel being specified first.

nmilcoff:
There has been an eternal discussion with issues around CurrentActivity in the past. I made the PR #2513 which may be conflicting here. What I did was basically bring into the fwk the improved version that @dazinator shared - but he left a comment there saying the implementation he shared may be outdated (and he shared a link to the latest one). I haven't checked that yet tbh.

Also the old mechanism wasn't removed, so as a fallback action we can just go back and remove the new implementation (which is only used if the project defines an "Application" class which inherits from `MvxAndroidApplication`).

dazinator:
Yeah probably worth checking over that new implementation.It could be marking the activity as no longer current as soon as its paused. This means when navigating, if the current activity is paused, and then you have some logic - it wont find a current activity. Until the new activity is Resumed. Its usually logic running in this gap that would encounter the null activity. I typically change my application logic so that things that rely on current activity run AFTER OnResume. However if thats not desirable you could just tweak the new implementation to not unmark an activity as current when it is paused. This would probably solve the null activity issue

nickrandolph:
@nmilcoff @dazinator are either of you able to adjust the behaviour of the IMvxAndroidCurrentTopActivity to avoid CurrentActivity ever being set to null (unless the application is being terminated I guess)?

nickrandolph:
Hi all, please review changes as this fixes some major issues with Droid

PR #2792, opened 2018-04-12
Update mvxnotifytask.md - Version specification

Ryujose:
Specificated the version when is available this feature. TAG 6.0.0-beta1 referenced. I think it's important to remark the feature as a beta and the version to not cause confusion on other people that's using another version.

URL Reference: https://github.com/MvvmCross/MvvmCross/blob/6.0.0-beta1/MvvmCross/ViewModels/MvxNotifyTask.cs

### DIFF
--- a/docs/_documentation/fundamentals/mvxnotifytask.md
+++ b/docs/_documentation/fundamentals/mvxnotifytask.md
@@ -5,6 +5,8 @@ category: Fundamentals
 order: 14
 ---
 
+[Available in version 6.0.0 beta1+] 
+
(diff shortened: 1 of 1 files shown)


nmilcoff:
Thanks! But `MvxNotifyTask` was actually added in v5.5.0

Ryujose:
Hello @nmilcoff 

Trought the tag of v5.5.0 I can't access to this file, maybe is on another folder? 

https://github.com/MvvmCross/MvvmCross/tree/5.5.0/MvvmCross/ViewModels/MvxNotifyTask.cs

Cheesebaron:
https://github.com/MvvmCross/MvvmCross/blob/5.5.0/MvvmCross/Core/Core/ViewModels/MvxNotifyTask.cs

Ryujose:
How do you see if we indicate the version where is available directly on top of the documentation? If you see that's good I'll do the change, if not, then I'll close the PR. As the version was clarified.

Issue #2829, opened 2018-04-24
Calling async method on ViewModel.Initialize never ends and InitializeTask properties never got updated


vackup:
### Steps to reproduce :scroll:

1. Create a ViewModel and override Initialize (this only happens if the first ViewModel is calling async code in Initialize method)

2. Call an async / await method (eg: refit service call)

```
public override async Task Initialize()
        {
            await base.Initialize();

            var response = await this.Service.GetDashboardValuesAsync();
        }
```

3. The device screen is black and never continue to the next line of code (eg: converting result to other model). InitializeTask status is always WaitingForActivation

### Expected behavior :thinking:
The execution flow should continue after awaited method complete its execution. This is working on MVVMCross 5.7

### Actual behavior :bug:
The device screen is black and never continue to the next line of code (eg: converting result to other model)

### Configuration :wrench:
VS 2017 15.6.6

**Version:** MvvmCross 6 using Xamarin.Forms 2.5.1.444934

**Platform:** 
- [ ] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ x] :monkey: Xamarin.Forms

nickrandolph:
A couple of things:
- For Xamarin Forms, please state which runtime platform as well (unless it's the same issue on all platforms in which case please state this)
- Please don't do long running activities in the Initialize method on the first view - this will cause problems on various devices and is not good practice (that's why it breaks in v6)
- If you must do long running actions, please look at https://nicksnettravels.builttoroam.com/post/2018/04/19/MvvmCross-Initialize-method-on-the-first-view-model.aspx for a suggestion on how to work around the issue. It shows it for Android, but the same basic concept workf for all platforms

ernestoyaquello:
Why isn't the workaround implemented by default MvvmCross?

nickrandolph:
@ernestoyaquello See the linked PR for what I'm thinking for v6.1. However, please note that you should really make use of a splash screen if you're going to be doing heavy lifting prior to the main view of your application.
The reason for the change in v6 is that there were a lot of issues due to the non-blocking nature of the Start call in MvxAppStart - doing long running activities were causing the app to not show the first view in time (particularly an issue with XF) and the app crashing or being terminated by the OS.
I'm closing this issue as the behaviour you're seeing is by design

vackup:
[...]

PR #2784, opened 2018-04-10
Making IMvxViewPresenter methods async

nickrandolph:
### :sparkles: What kind of change does this PR introduce? (Bug fix, feature, docs update...)
Feature

### :arrow_heading_down: What is the current behavior?
All view presenter methods are fire and forget (for async activities)

### :new: What is the new behavior (if this is a feature change)?
All methods return Task<bool> - bool indicates success/fail

### :boom: Does this PR introduce a breaking change?
Yes

### :bug: Recommendations for testing
Needs lots of help testing and confirming no breakage of playground samples

### :thinking: Checklist before submitting

- [ ] All projects build
- [ ] Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))
- [ ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contribute/mvvmcross-docs-style-guide))
- [ X ] Rebased onto current develop

PR #2830, opened 2018-04-24
[WIP] Add support for async startup

nickrandolph:
Feature

Currently the built in MvxAppStart only supports a synchronous startup. In cases where a splash screen is used, it should be possible to use an async startup which would allow the Initialize method to do long running actions. Currently this is achieved by overriding the default MvxAppStart.

Added support for StartAsync into IMvxAppStart
Altered SplashScreen (Android only) to make use of StartAsync
TODO: Add equivalent splash screen view to other platforms to show how this could work

Yes - some interfaces now return Task where previously was void

Follow up PR to: https://github.com/MvvmCross/MvvmCross/pull/2784
In response to Issue: https://github.com/MvvmCross/MvvmCross/issues/2829

- [ ] All projects build
Note: This needs to be rebased onto v6.1 after PR 2784 has been merged

### DIFF
--- a/MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatApplication.cs
+++ b/MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatApplication.cs
@@ -12,7 +12,7 @@ namespace MvvmCross.Droid.Support.V7.AppCompat
 {
     public abstract class MvxAppCompatApplication<TMvxAndroidSetup, TApplication> : MvxAndroidApplication
   where TMvxAndroidSetup : MvxAppCompatSetup<TApplication>, new()
-  where TApplication : IMvxApplication, new()
+  where TApplication : class, IMvxApplication, new()
     {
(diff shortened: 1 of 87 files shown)


nmilcoff:
This will be a really nice improvement. However I don't think we should include this PR on a minor version, as there will be a lot of breaking changes.

nickrandolph:
@nmilcoff what breaking changes do you see that will affect a large number of users? I'm really keen to get this into v6.1 as it addresses an area that v6 broke for a lot of people

nickrandolph:
@martijn00 what else do we need to do in order to get this PR completed?

nmilcoff:
@nickrandolph all apps using custom ViewPresenters will break, and the interface changes for `IMvxApplication` will also have an impact. Other than that can we please move all the IoC changes to a separate PR? There is no description about them, which will make them hard to track in the future (also probably a bit unrelated?)

PR #2837, opened 2018-05-01
housekeeping: use https

ghuntley:


martijn00:
We've got the same thing going on as https://github.com/reactiveui/ReactiveUI/pull/1602 and we were thinking to switch to MIT a while ago, but just not sure what the legal side is. How far are you guys on that?

ghuntley:
Should have an answer in next 12 hours or so re: MIT license switch. It may be more complicated for you depending on how IP was assigned. Will definitely share what we learn.

Issue #2842, opened 2018-05-02
UWP Release Builds are crashing at runtime


ueman:
Release builds on UWP are crashing at runtime at startup with a 'System.PlatformNotSupportedException' in System.Private.CoreLib.dll` exception. A repository to reproduce is here https://github.com/ueman/mvvmcross6crash

The UWP Playground should yield the same behaviour as the provided repository.

Upon further investigation this seems to be a problem with the build option `Compile with .NET Native tool chain`. If the option is turned off it works as expected, however you can't deploy to the Windows App Store. So this is still a problem.

### Steps to reproduce :scroll:

1. Create new project with MvvmCross 6. The Core project is a .Net Standard 2 Class Library and the UWP project is a Universal Windows Project. MvvmCross is included as NuGet.

2. Compile release build.

3. Run release build.

### Expected behavior :thinking:
The application should not crash.

### Actual behavior :bug:
The application crashes with a `Exception thrown: 'System.PlatformNotSupportedException' in System.Private.CoreLib.dll` Exception.

### Configuration :wrench:

**Version:** 6.0.0 and 6.0.1

**Platform:** 
- [ ] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [x] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

nickrandolph:
Good find - frustrating that it works in debug but not release (thanks for the great simple example)
Workaround - use a custom setup class that overrides getpluginassembly
public class CustomSetup: MvxWindowsSetup<ClassLibrary1.App>
    {
        public override IEnumerable<Assembly> GetPluginAssemblies()
        {
            return new[] { typeof(MvxPluginAttribute).Assembly };
        }
    }

ueman:
Thanks for the response.

After some research i think the exception comes from the code in this method https://github.com/MvvmCross/MvvmCross/blob/082ce799a17f31839b7c1221aacc4737942bc58e/MvvmCross/Core/MvxSetup.cs#L270 and has something to do with reflection.

nickrandolph:
@ueman as per my workaround, the issue is that GetPluginAssembly (which is called by InitializePluginFramework) is calling a method that isn't supported on UWP

ueman:
Closing because the issue was fixed.

PR #2844, opened 2018-05-02
Fixing crash on UWP in Release due to plugin

nickrandolph:
Bug fix

### Comments
Currently the plugin model uses reflection to look for plugin attributes, which is always going to be slow. Given issues such as this, I suggest we move away from this model to one where plugins need to be explicitly referenced by the head projects and/or core library.

UWP crashes when attempting to load plugins in Release (ie with .NET Native compile)

UWP no longer crashes

No

https://github.com/MvvmCross/MvvmCross/issues/2842

- [ ] All projects build

### DIFF
--- a/MvvmCross/Core/MvxSetup.cs
+++ b/MvvmCross/Core/MvxSetup.cs
@@ -291,12 +291,23 @@ namespace MvvmCross.Core
                 AppDomain.CurrentDomain
                     .GetAssemblies()
                     .AsParallel()
-                    .Where(AssemblyReferencesMvvmCross);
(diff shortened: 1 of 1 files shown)


nickrandolph:
@martijn00 any thoughts on how we want to fix this issue - my changes aren't really a good option as they just handle the exception that is raised on UWP.

martijn00:
I suggest we merge this for now. I can't think of anything better so far.

nickrandolph:
Sure - suggest we flag an issue for reviewing how plugins are registered (need an alternative that avoids all reflection calls to improve perf)

PR #2854, opened 2018-05-04
Mvx app start with params

LRP-sgravel:
Allows a parametrized app start hint and the possibility to provide the hint from shared code.

Non-existent

Start your app with a parametrized hint without using a custom app start

no

- [ X] All projects build
- [X ] Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))
- [] Rebased onto current develop


nickrandolph:
@LRP-sgravel is there a reason we can't push changes to your fork? when you submitted the PR did you uncheck the "Allow edits from maintainers." checkbox? If so, would you mind unchecking it so that we can help you get this PR merged?

nickrandolph:
@LRP-sgravel  I've made some adjustments to the way the startup parameter is passed from the application into the initial navigation.
The startup hint was intended for platform specific startup hints to be passed in (eg ActivationEventArgs for UWP), so I've reverted your changes around passing in a hint coming from the application object. Instead if the application is typed, the StartParameter will be retrieved and passed into the initial navigation. 
Hopefully this aligns with the intent behind your initial PR whilst still maintaining the ability for platform specific hints to be passed into appstart

sylvain-toptal:
I don't necessarily mind the changes.  However, if the hint is meant to be platform specific, why is it passed around as an ˋobject` everywhere, including to cross platform code like the MvxAppStart and MvxApp?  How would you handle an NSDictionnary (for an iOS push for instance) in there?  My reasoning was more along the lines that you would pass a cross platform conversion of whatever platform data you retrieved and then handle it in your first view model, which I assumed was a loading screen of some sort.  Is there a use-case where you would want platform specific AppStart?

nickrandolph:
Yep, but you were sourcing the start parameter from MvxApplication - which my changes still do.
The point of the startup hint is that you may want to vary startup behaviour for specific platforms and because each platform has different startup objects (Bundle for Android, ActivationArgs for UWP etc) it's not really possible to strongly type that object. 
Happy to consider other ways that we can handle the startup hint to make it more strongly typed but I do see it as being separate from a startup parameter coming from MvxApplication. 
The other thing I really wanted to avoid is calling MvxResolve in order to grab a reference to the MvxApplication - The MvxAppStart already has a reference to MvxApplication, so it seems better to call it directly.

sylvain-toptal:
Well, yes I don't think it is unreasonable as a default implementation to use the cross platform hint.  If you wanted to deal with the platform type, the app entry point has the GetAppStartHint you can override already.  Why add the burden of also subclassing AppStart to deal with a native type?

sylvain-toptal:
That is absolutely not what I am saying.  I'm saying that the AppStart should also be agnostic.  We already have all that we need to make the hint platform agnostic, so why not consider the hint in the AppStart as agnostic too and therefore we should keep the generic parameterized AppStart and it's registration method.  While we are at it, we can pass it to the first view model which happens to be the one we declare in our AppStart, basically grouping what will eventually go together.

nickrandolph:
So how does the updated code not do this? What further changes are you suggesting? (feel free to go ahead and make changes if you want)

sylvain-toptal:
It forces you to subclass AppStart because you can't set the Application StartParameter without subclassing the AppStart too.  I feel the changes add to the boilerplate code and reducing it seems to have been a focus of versions 5 & 6.

nickrandolph:
Mvx requires you to sub class the MvxApplication class which is all you need to do in order to set the StartParameter. I don't see the scenario (which the changes I've made) where you have to sub-class AppStart?

sylvain-toptal:
Hey, good morning!

You would need to subclass the AppStart to pass then .Net converted platform data to the first view model.  I am under the impression that we are duplicating an abstraction of launch data (the hint) with limited benefits.

So let’s get back to my original intention for the PR, which was to pass the hint to the first view model so it can be handled by shared code.  The hint would of course need to be converted to a shared code data structure to do so.  We lose this ability with your changes unlesss we subclass the AppStart to provide the hint to the Application subclass (or if we Resolve the Application class within the app entry point which is what you are also trying to avoid).  The fact that I proposed to introduce a hint at the Application level and to use it by default is a simple optional by-product made possible by this vision.  I’m fine with not introducing those changes if it’s what bothers you, I simply thought of it an extension that could make sense in certain circumstances (initalizing with a build-specific configuration constant for instance).

Basically, what I’m saying is that I don’t see the reason why the hint has to be considered platform specific.  This seems like a limited use of a generic mechanism, namely using an `object` and passing it everywhere for initialization, including to shared code. The fact that we already pass this abstraction to the AppStart and the Application leads me to believe we cannot rely on it being of a platform specific type, since we have no access to those types in the shared code.

If I get down to your issues with my proposed implementation, I agree that resolving the Application in the app entry point is a bit ugly.  I used that because there are already calls to Resolve at different places in the app entry points (i.e. https://github.com/MvvmCross/MvvmCross/blob/59809878c68b7e8f3648e54aca1dd921af17fbfe/MvvmCross/Platforms/Ios/Core/MvxApplicationDelegate.cs#L55) and the fact that I didn’t want to introduce breaking changes.  Maybe using `Mvx.CallbackWhenRegistered` would be a more acceptable approach, but then again it only hides the Resolve call.

I like the discussion by the way, it’s constructive and helps me better understand the philosophy behind the implementation.

nickrandolph:
Ok, I think I understand the scenario - am I right in saying that you're using a shared project and that your viewmodels can pick up platform specific parameters (ie using conditional compilation)? If this is the scenario my first thought is that I'm not a big fan of shared projects and that you should attempt to move to at a multi-targeted library but that's more of a personal dislike for shared projects.
You'll see I've made a small adjustment that hopefully should address some of what you want to do:
- You can now register a generic appstart even if your MvxApplication isn't generic
- If the type of starthint matches the generic parameter of appstart, it'll be sent through to the viewmodel.
Can you confirm that this is sufficient for your requirements? If not, what part of the scenario are we still missing.

LRP-sgravel:
No @nickrandolph, I don'T use conditional compiles.  I feel like a broken record so I'll just write it down in pseudo-code, using iOS as an example.

AppDelegate :
```
        object GetAppHintStart()(object hint)
        {
            NSDictionnary launchData = hint as NSDictionnary;
            MvxBundle bundle = launchData.ToMvxBundle();  // Convert the data to cross platform data structure we want to pass to the LoadingViewModel

            return bundle;
        }
```

in MvxApplication subclass :
```
        public override void Initialize()
        {
            CreatableTypes()
                .EndingWith("Service")
                .AsInterfaces()
                .RegisterAsLazySingleton();

            Mvx.RegisterSingleton<IMvxTextProvider>(new TextProviderBuilder().TextProvider);

            RegisterAppStart<RootViewModel, MvxBundle>();
        }
```

in RootViewModel:
```

        public override void Prepare(MvxBundle hint)
        {
            base.Prepare();

            _nextPath = hint.Data["deeplink"];
            _nextVMData = hint.Data["deeplinkdata"];;
        }

        private void LoadingCompleted()
        {
            NavigationService.Navigate<string>(_nextPath, _nextVMData);
        }
```

LRP-sgravel:
I don't think we need the generic application class. I'll submit a new proposition with this in mind.

Personnally, I think we should enforce `IMvxBundle` for the hint at the AppStart and Application level instead of an object.  This means the entry point classes GetAppStartHint should return a `IMvxBundle` instead of an `object` but this breaks backwards compatibility so I wasn't keen on making this change.

MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -88,4 +88,31 @@ protected virtual void NavigateToFirstViewModel(object hint)
...
+    {
+        public MvxAppStart(IMvxApplication application, IMvxNavigationService navigationService) : base(application, navigationService)
+        {
+        }
+
+        protected override void NavigateToFirstViewModel(object hint)
+        {
+            TParameter navParam = default;
+
+            hint = Application.GetAppStartHint(hint);
```
martijn00: I'm not sure why this has to go through the Application class. In the base `ApplicationStartup` already calls into that with the hint in it. I'm thinking that should maybe change to be able to customize the hint.

nickrandolph: Yeh we can just modify Startup to return the updated hint

LRP-sgravel: Yeah sure, that works for me.

nmilcoff: This line should be something like `hint = hint ?? Application.GetAppStartHint(hint);`. Otherwise the parameter is always replaced and there's no way to pass anything from the native platform

LRP-sgravel: I'm not sure I agree with this as the app start hint can be additive, especially when using a MvxBundle.

LRP-sgravel:
Pushed changes as proposed, using `Startup` instead of a new method.

MvvmCross/ViewModels/MvxAppStart.cs
LRP-sgravel: Also, notice that the default implementation is to return the provided hint, so nothing is lost.

LRP-sgravel:
@nickrandolph @nmilcoff I'm not sure I understand what your concerns are exactly.  Care to elaborate?

MvvmCross/ViewModels/IMvxAppStart.cs
```diff
@@ -6,7 +6,7 @@ namespace MvvmCross.ViewModels
 {
     public interface IMvxAppStart
     {
-        void Start(object hint = null);
+        object Start(object hint = null);
```
nickrandolph: Is there a good reason to return the updated hint here?

MvvmCross/ViewModels/MvxAppStart.cs
nmilcoff: You're right! Sorry

nmilcoff: That introduces a breaking change guys... Can we please avoid that?

MvvmCross/ViewModels/IMvxAppStart.cs
LRP-sgravel: Oops, that's just me going crazy on the copy-paste!  Reverting...

LRP-sgravel:
So I think we now all agree on the intention and use-case.  One last point I think we should discuss though.  My original changes did intriduce breaking changes.  The new ones do because of the change in return type for `Startup`.  I'm therefore coming back with a last proposition: Should we not enforce a `IMvxBundle` for the `AppStartHint` and `Startup` to make sure we don't get platform specific types passed in?  After all, as far as I know, all platform hints are a dictionnary of some sort, so we might as well enforce it.

What do you think?

MvvmCross/ViewModels/MvxAppStart.cs
LRP-sgravel: Yes it does and that is what drove my previous proposal, see my comment below.  If we accept breaking changes, we might as well do it for good and enforce the parameter type to `MvxBundle`.

MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -91,24 +99,42 @@ protected virtual void NavigateToFirstViewModel(object hint)
         }
     }
 
-    public class MvxAppStart<TViewModel, TParameter> : MvxAppStart<TViewModel> where TViewModel : IMvxViewModel<TParameter>
+    public class MvxAppStart<TViewModel, TParameter> : MvxAppStart<TViewModel> where TViewModel : IMvxViewModel
```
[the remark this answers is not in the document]
nickrandolph: Because developers may want to use the generic MvxApplication (so they get a strongly typed hint there) without having to use the generic IMvxViewModel

MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -91,24 +99,42 @@ protected virtual void NavigateToFirstViewModel(object hint)
...
         public MvxAppStart(IMvxApplication application, IMvxNavigationService navigationService) : base(application, navigationService)
         {
         }
 
+        protected override object ApplicationStartup(object hint = null)
+        {
+            if (hint is TParameter typedHint &&
+                Application is IMvxApplication<TParameter> typedApplication)
+            {
+                return typedApplication.StartupWithHint(typedHint);
```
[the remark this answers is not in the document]
nickrandolph: The typedApplication already calls the base Startup method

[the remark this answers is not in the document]
LRP-sgravel: But that doesn't forward the hint to the base `ApplicationStartup`.

MvvmCross/ViewModels/MvxAppStart.cs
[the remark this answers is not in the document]
LRP-sgravel: Legit!

martijn00:
@LRP-sgravel can you do a rebase?

sylvain-toptal:
Not sure that went as it should...

nickrandolph:
@LRP-sgravel you needed to do a force push (ie --force) in order to clean up the commit history

sylvain-toptal:
Oh

sylvain-toptal:
Hello guys, is there anything missing for this PR to go through?  I'm not sure where we are at...

nickrandolph:
@LRP-sgravel there a couple of things:
- Needs to be rebased with --force to clear commit history
- We need to get WPF and XF v3 stable for 6.1, so there's a risk that this might get bumped to 6.2

sylvain-toptal:
Hello @nickrandolph, @martijn00 told me he did the rebase...  It had some merge issues though. Is it fine now?

sylvain-toptal:
Thanks for the rebase again @martijn00

MvvmCross/ViewModels/IMvxApplication.cs
```diff
@@ -12,8 +12,13 @@ public interface IMvxApplication : IMvxViewModelLocatorCollection
 
         void Initialize();
 
-        void Startup(object hint);
+        void Startup();
```
martijn00: I would leave this `void Startup(object hint);`, for one to not break it, and another, maybe people just don't want to use a generic. Also adjust the trace messages to do this.

sylvain-toptal: I agree.  I think this change was propose by @nmilcoff perhaps he has an explanation.

nmilcoff: Mhmm it wasn't me 🤔

nickrandolph: If someone wants to pass a parameter in (even if just object) they should use the generic. I don’t see much value in having the old method syntax as passing around object is just bad practice and should be avoid if possible.

nmilcoff: That's a fair point, but it's still a breaking change. IMO we should try to avoid breaking users code as much as possible, unless it's a major release. Otherwise we'll loose their trust

nickrandolph: I disagree. Break without reason I totally agree, but the point of minor release is that you can do minimal api changes ie no architectural changes. 
What we can’t afford to do is to stagnate and not clean up code for fear of breaking developers. It’s OSS and comes “as is”

sylvain-toptal: We also broke the MvxAppStart Startupmethod by changing the return parameter to an `object`.  So either we accept or refuse both breaking changes.

builttoroamadmin:
Hey why was this closed? We’re these changes not going to be merged?

PR #2858, opened 2018-05-06
Improve MvvmCross Getting Started Experience (ReadMe Content & Sample Files)

andrewtechhelp:
Improves the Getting Started experience for MvvmCross by adding some basic set up instructions to the readme.txt file that automatically open when you add the MvvmCross or MvvmCross.Forms NuGet packages your project. These are a halfway step between the existing extremely brief instructions in the ReadMe and a full tutorial that explains what each step is doing.

I took a mix of some existing MvvmCross sample projects, plus some of the helpful instructions from Nick Randolph's blog posts and samples (to understand platforms I'd never worked with before such as macOS) to create these files. If you use these sample files and the ReadMe, you'll get a project with a text field that says "Hello MvvmCross!", with a Reset button to reset that text if modified.

ReadMe file is extremely brief and the sample files in the ContentFiles folder are out of date.

ReadMe file is now more detailed than previously and the sample files in the ContentFiles folder can be used to create a bare-bones MvvmCross project.

No

Try the instructions in the ReadMe and see if they make sense to you. Hopefully, this is useful.

- [ ] All projects build

### DIFF
--- a/MvvmCross/readme.txt
+++ b/MvvmCross/readme.txt
@@ -43,6 +43,12 @@ However, any $rootnamespace$ instances will need to be changed to your project's
 5. Add a Views folder to the root of the project and add at least one View class to this folder to correspond to the ViewModel class in the Core project (See Android/HomeView.cs.pp in sample files).
 6. Add a new Android layout to the Resources/layout folder to correspond to the View class created in the previous step. (See Android/HomeView.axml.pp in sample files).
 
+Note: If you wish to use the AppCompat versions of Android classes, you can follow the above instructions with the following modifications
+7. Add the MvvmCross v7 Android AppCompat Support Libraries NuGet package (MvvmCross.Droid.Support.v7.AppCompat) to to your Android platform specific project.
+8. When changing the the inheritance for the SplashScreen activity class, use MvxSplashScreenAppCompatActivity instead of MvxSplashScreenActivity.
+9. Add a new XML file that defines an AppCompat theme to the Resources/values folder called styles.xml. (See Android/styles.xml.pp in sample files)
+10. Ensure the theme created in the previous step is referenced in the Attribute for the MainActivity class (Theme = "@style/MainTheme").
+
 
 - iOS projects (ignore if not building for iOS) -
 1. Inside AppDelegate.cs, change the AppDelegate class to inherit from MvxApplicationDelegate<MvxIosSetup<Core.App>, Core.App> instead of ApplicationDelegate (See iOS/AppDelegate.cs.pp in sample files).
--- a/ContentFiles/Android/styles.xml.pp
+++ b/ContentFiles/Android/styles.xml.pp
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<resources>
+
+  <style name="MainTheme" parent="MainTheme.Base">
+  </style>
+  <!-- Base theme applied no matter what API -->
+  <style name="MainTheme.Base" parent="Theme.AppCompat.Light.DarkActionBar">
+    <!--If you are using revision 22.1 please use just windowNoTitle. Without android:-->
+    <item name="windowNoTitle">true</item>
+    <!--We will be using the toolbar so no need to show ActionBar-->
+    <item name="windowActionBar">false</item>
+    <!-- Set theme colors from http://www.google.com/design/spec/style/color.html#color-color-palette -->
+    <!-- colorPrimary is used for the default action bar background -->
+    <item name="colorPrimary">#2196F3</item>
+    <!-- colorPrimaryDark is used for the status bar -->
+    <item name="colorPrimaryDark">#1976D2</item>
+    <!-- colorAccent is used as the default value for colorControlActivated
+         which is used to tint widgets -->
+    <item name="colorAccent">#FF4081</item>
+    <!-- You can also set colorControlNormal, colorControlActivated
+         colorControlHighlight and colorSwitchThumbNormal. -->
+    <item name="windowActionModeOverlay">true</item>
+
+    <item name="android:datePickerDialogTheme">@style/AppCompatDialogStyle</item>
+  </style>
+
+  <style name="AppCompatDialogStyle" parent="Theme.AppCompat.Light.Dialog">
+    <item name="colorAccent">#FF4081</item>
+  </style>
+</resources>


martijn00:
Great work on this, we can use more PR's like this! One more optimisation we can do is include different readme on each platform. Using the platform specific sections we can pull that off. Could you work on that?

martijn00:
@andrewtechhelp please sign the CLA so we can merge this.

ContentFiles/Forms/AndroidContent/MainActivity.cs.pp
```diff
@@ -0,0 +1,11 @@
+﻿using Android.App;
+using MvvmCross.Forms.Platforms.Android.Core;
+using MvvmCross.Forms.Platforms.Android.Views;
+
+namespace $rootnamespace$
+{
+    [Activity(Label = "$rootnamespace$", MainLauncher = true, Theme = "@style/MainTheme", NoHistory = true)]
+    public class MainActivity : MvxFormsAppCompatActivity<MvxFormsAndroidSetup<Core.App, FormsUI.App>, Core.App, FormsUI.App, MainViewModel>
```
nmilcoff: `MainViewModel` doesn't exist anymore here, does it?

ContentFiles/Forms/FormsUIContent/App.xaml.cs.pp
```diff
@@ -0,0 +1,12 @@
+﻿using Xamarin.Forms;
+
+namespace $rootnamespace$
```
nmilcoff: I think this namespace should be `FormsUI`, as that is assumed later when referencing this `App` class

ContentFiles/Forms/FormsUIContent/App.xaml.pp
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<Application
+    xmlns="http://xamarin.com/schemas/2014/forms"
+    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+    x:Class="$rootnamespace$.App">
```
nmilcoff: I think this namespace should be `FormsUI`, as that is assumed later when referencing this `App` class

ContentFiles/Forms/FormsUIContent/HomeView.xaml.cs.pp
```diff
@@ -0,0 +1,12 @@
+﻿using MvvmCross.Forms.Views;
+
+namespace $rootnamespace$.Views
+{
+    public partial class HomeView : MvxContentPage
```
nmilcoff: Can we use the strong typed version here? Something like:

`public partial class HomeView : MvxContentPage<HomeViewModel>`

ContentFiles/Forms/FormsUIContent/HomeView.xaml.pp
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<views:MvxContentPage
+    xmlns="http://xamarin.com/schemas/2014/forms"
+    xmlns:views="clr-namespace:MvvmCross.Forms.Views;assembly=MvvmCross.Forms"
+    xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+    xmlns:local="clr-namespace:$rootnamespace$.Views"
+    x:Class="$rootnamespace$.Views.HomeView">
```
nmilcoff: If we were to use the strong typed version, this file needs to be edited to include `x:TypeArguments="viewModels:HomeViewModel"` and `xmlns:viewModels="clr-namespace:$rootnamespace$.ViewModels;assembly=$rootnamespace$"`

ContentFiles/WPF/HomeView.xaml.cs.pp
```diff
@@ -0,0 +1,15 @@
+using AlphanumericOz.Core.ViewModels;
```
nmilcoff: I think this should be `using $rootnamespace$.ViewModels;` 🙂

MvvmCross/readme.txt
```diff
@@ -2,36 +2,102 @@
...
+
+- Core project -
+1. Add an App class to the root folder (See Core/App.cs.pp in sample files).
+2. Add a ViewModels folder to the root of the project and add at least one ViewModel class to this folder (See Core/HomeViewModel.cs.pp in sample files).
+
+
+- Android projects (ignore if not building for Android) -
+1. Add a reference to the Mono.Android.Export assembly.
+2. Delete MainActivity.cs, and add a new activity class called SplashScreen to the root folder.
+3. Change the SplashScreen activity class to inherit from MvxSplashScreenActivity<MvxAndroidSetup<Core.App>, Core.App> instead of Activity (See Android/SplashScreen.cs.pp in sample files).
```
nmilcoff: Maybe it worths to leave a note about Support packages in here?

ContentFiles/WPF/HomeView.xaml.cs.pp
andrewtechhelp: yep, good pickup, accidentally left my sample project in there.

ContentFiles/Forms/FormsUIContent/HomeView.xaml.cs.pp
andrewtechhelp: What's the advantage to doing this? Happy to change it if you think it's better to include it, but am not sure why, since it adds more things you have to add to the XAML file (trying where possible to have this be the briefest amount of changes for the user as possible)

ContentFiles/Forms/FormsUIContent/App.xaml.cs.pp
andrewtechhelp: Ok, so this reply will cover all of the comments around the namespaces. Everything I've done is based on the assumption that $rootnamespace$ is the project name, not the solution name. So in this case '$rootnamespace$' should be replaced by the user with 'YourSolutionName.FormsUI' because that's the project name (meaning you would went up with 'YourSolutionName.FormsUI.App'). I think that by default when you create a new project the namespace for that project is the project name, hence why I went with that. I can change this to the solution name if you want, but I'd have to change every namespace in this pull request to make that consistent.

This brings up another question: Are these Content Files still automatically used by NuGet to insert into people's projects when they install any of the MvvmCross NuGet packages? I didn't see this happen with the main packages, but haven't poked around further to see if they're other projects. I kind of assume they don't currently get automatically copied and so are more just sample files people can reference, but might get automatically copied in the future.

andrewtechhelp:
@martijn00 thanks for that. Will have a look at that next week if I get time :)

PR #2862, opened 2018-05-08
Plugin loading on references of core assemblies

rdorta:
Fixes cases of plugin usage in core assembly with no explicit reference on the entry point assembly.
This makes MvxSetup load all referenced assemblies in the entry point referenced assemblies.
Fixes cases where in WPF as an example, we don't need to have a LinkerPleaseIncludeFile and also have no need to create an instance or static reference to a Plugin.

If you use a plugin in a WPF project as an example you have to create a strong reference to a Plugin type so that the plugin assembly can be loaded in AppDomain.

Now you don't need a strong reference, because you probably already have it in your Core project, so MvxSetup will look for plugins in the core project references as well.

No.

Running any plugin in all projects without having a strong reference type to any type in plugin assembly.

- [X ] All projects build
- [X ] Rebased onto current develop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,48 +1,135 @@
 # Change Log
 
+## [6.0.1](https://github.com/MvvmCross/MvvmCross/tree/6.0.1) (2018-04-29)
+[Full Changelog](https://github.com/MvvmCross/MvvmCross/compare/6.0.0...6.0.1)
+
+**Fixed bugs:**
+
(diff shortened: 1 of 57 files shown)


martijn00:
This is probably related to https://github.com/MvvmCross/MvvmCross/pull/2844

Can you rebase on 6.1 and check that change?

rdorta:
Did something wrong maybe? I rebased it on 6.1.0 and inherited lots of commits.

martijn00:
@rdorta Yeah you merged instead of rebase it seems. I took a look at rebasing, but got lots of conflicts. I think it might be best to make a new PR with 6.1 as base branch and copy the code changes into that new branch.

rdorta:
k

PR #2828, opened 2018-04-23
[WIP] consolidate samples into one playground

martijn00:
### :sparkles: What kind of change does this PR introduce? (Bug fix, feature, docs update...)

Samples

### :boom: Does this PR introduce a breaking change?

No

### :thinking: Checklist before submitting

- [ ] All projects build
- [ ] Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))
- [ ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contribute/mvvmcross-docs-style-guide))
- [ ] Rebased onto current develop

PR #2865, opened 2018-05-08
Added recursive method to search for referenced assemblies and load t…

rdorta:
Recreated based on 6.1.0

Plugin loading on references of core assemblies 

Fixes cases of plugin usage in core assembly with no explicit reference on the entry point assembly.
This makes MvxSetup load all referenced assemblies in the entry point referenced assemblies.
Fixes cases where in WPF as an example, we don't need to have a LinkerPleaseIncludeFile and also have no need to create an instance or static reference to a Plugin.

If you use a plugin in a WPF project as an example you have to create a strong reference to a Plugin type so that the plugin assembly can be loaded in AppDomain.

Now you don't need a strong reference, because you probably already have it in your Core project, so MvxSetup will look for plugins in the core project references as well.

No.

Running any plugin in all projects without having a strong reference type to any type in plugin assembly.

- [x] All projects build

### DIFF
--- a/MvvmCross/Core/MvxSetup.cs
+++ b/MvvmCross/Core/MvxSetup.cs
@@ -286,16 +286,32 @@ namespace MvvmCross.Core
         public virtual IEnumerable<Assembly> GetPluginAssemblies()
         {
             var mvvmCrossAssemblyName = typeof(MvxPluginAttribute).Assembly.GetName().Name;
-
+            var allAssemblies = LoadAllReferencedAssemblies(Assembly.GetExecutingAssembly());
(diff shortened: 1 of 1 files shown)


martijn00:
Thanks! This solves some problems!

rdorta:
I'll pick one of the up for grab cards now. :)

martijn00:
@rdorta You can always text me on the XamarinChat Slack if you are looking for things to work on! Maybe you can help finish https://github.com/MvvmCross/MvvmCross/pull/2828 for now?

PR #2872, opened 2018-05-09
[WIP] Making it easier to override presenters

nickrandolph:
Feature

In order to override MvxFormsPagePresenter you need to override a lot of other classes

Now you can just specify the type of the IMvxFormsPagePresenter you want to use when specifying the setup class

No

- [ ] All projects build

### DIFF
--- a/MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatSetup.cs
+++ b/MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatSetup.cs
@@ -11,12 +11,20 @@ using MvvmCross.Binding.Bindings.Target.Construction;
 using MvvmCross.Droid.Support.V4;
 using MvvmCross.Platforms.Android.Core;
 using MvvmCross.Platforms.Android.Presenters;
+using MvvmCross.Presenters;
 using MvvmCross.ViewModels;
 
 namespace MvvmCross.Droid.Support.V7.AppCompat
 {
     public abstract class MvxAppCompatSetup : MvxAndroidSetup
     {
+        protected override void InitializeIoC()
+        {
+            base.InitializeIoC();
+
+            Mvx.LazyConstructAndRegisterSingleton<IMvxViewPresenter, MvxAppCompatViewPresenter>();
+        }
+
         protected override IEnumerable<Assembly> AndroidViewAssemblies =>
             new List<Assembly>(base.AndroidViewAssemblies)
             {
@@ -24,11 +32,6 @@ namespace MvvmCross.Droid.Support.V7.AppCompat
                 typeof(DrawerLayout).Assembly,
                 typeof(MvxSwipeRefreshLayout).Assembly
             };
-
-        protected override IMvxAndroidViewPresenter CreateViewPresenter()
-        {
-            return new MvxAppCompatViewPresenter(AndroidViewAssemblies);
-        }
 
         protected override void FillTargetFactories(IMvxTargetBindingFactoryRegistry registry)
         {
--- a/MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatSetup.cs
+++ b/MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatSetup.cs
@@ -18,9 +18,9 @@ namespace MvvmCross.Droid.Support.V7.AppCompat
 {
     public abstract class MvxAppCompatSetup : MvxAndroidSetup
     {
-        protected override void InitializeIoC()
+        protected override void RegisterImplementations()
         {
-            base.InitializeIoC();
+            base.RegisterImplementations();
 
             Mvx.LazyConstructAndRegisterSingleton<IMvxViewPresenter, MvxAppCompatViewPresenter>();
         }
--- a/MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatSetup.cs
+++ b/MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatSetup.cs
@@ -9,6 +9,7 @@ using Android.Support.V7.Widget;
 using MvvmCross.Binding.BindingContext;
 using MvvmCross.Binding.Bindings.Target.Construction;
 using MvvmCross.Droid.Support.V4;
+using MvvmCross.IoC;
 using MvvmCross.Platforms.Android.Core;
 using MvvmCross.Platforms.Android.Presenters;
 using MvvmCross.Presenters;
@@ -18,11 +19,10 @@ namespace MvvmCross.Droid.Support.V7.AppCompat
 {
     public abstract class MvxAppCompatSetup : MvxAndroidSetup
     {
-        protected override void RegisterImplementations()
+        protected override void RegisterViewPresenter()
         {
-            base.RegisterImplementations();
-
-            Mvx.LazyConstructAndRegisterSingleton<IMvxViewPresenter, MvxAppCompatViewPresenter>();
+            base.RegisterViewPresenter();
+            Mvx.IoCProvider.LazyConstructAndRegisterSingleton<IMvxViewPresenter, MvxAppCompatViewPresenter>();
         }
 
         protected override IEnumerable<Assembly> AndroidViewAssemblies =>
@@ -49,7 +49,10 @@ namespace MvvmCross.Droid.Support.V7.AppCompat
     public class MvxAppCompatSetup<TApplication> : MvxAppCompatSetup
         where TApplication : class, IMvxApplication, new()
     {
-        protected override IMvxApplication CreateApp() => Mvx.IoCProvider.IoCConstruct<TApplication>();
+        protected override void RegisterApp()
+        {
+            Mvx.IoCProvider.LazyConstructAndRegisterSingleton<IMvxApplication, TApplication>();
+        }
 
         public override IEnumerable<Assembly> GetViewModelAssemblies()
         {
--- a/MvvmCross.Forms.Wpf/Platforms/Wpf/Core/MvxFormsWindowsSetup.cs
+++ b/MvvmCross.Forms.Wpf/Platforms/Wpf/Core/MvxFormsWindowsSetup.cs
@@ -34,9 +34,9 @@ namespace MvvmCross.Forms.Platforms.Wpf.Core
             }
         }
 
-        protected override void InitializeIoC()
+        protected override void RegisterImplementations()
         {
-            base.InitializeIoC();
+            base.RegisterImplementations();
             Mvx.RegisterSingleton<IMvxFormsSetup>(this);
 
             Mvx.LazyConstructAndRegisterSingleton<IMvxViewPresenter, MvxFormsWpfViewPresenter>();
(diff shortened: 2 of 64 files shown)


nickrandolph:
I think one of the issues for developers wanting to extend parts of mvvmcross is that there are cases where instances of Mvx objects are created and the philosophy has been to create methods that can be overridden. The issue is that this means that developers have to override a bunch of classes in order to override that one method. eg if you want to override a method in the formspagepresenter you have to override setup (multiple platforms) as well as the mvxformspagepresenter. 
An alternative to adding various generics to setup would be to have each of these types registered with the IoC container and then expose a point where developers can override any type they want. The big question is where would this go?

nickrandolph:
I've reworked this so that the MvxFormsPagePresenter is registered with the IoC container

nickrandolph:
@martijn00 I ended up registering all the presenters - I've still got a bit of testing to do on all platforms to make sure nothing's broken.

MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatSetup.cs
```diff
@@ -11,12 +11,20 @@
...
 
 namespace MvvmCross.Droid.Support.V7.AppCompat
 {
     public abstract class MvxAppCompatSetup : MvxAndroidSetup
     {
+        protected override void InitializeIoC()
+        {
+            base.InitializeIoC();
+
+            Mvx.LazyConstructAndRegisterSingleton<IMvxViewPresenter, MvxAppCompatViewPresenter>();
```
nmilcoff: It looks a bit weird to register all ViewPresenters from `InitializeIoC`... Although I suppose this is because we need it registered as early as possible (right?). 

Maybe we can at least create a separate method for that? That would make overriding the behavior cleaner

MvvmCross.Forms/Core/IMvxFormsSetupHelper.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using MvvmCross.Binding.BindingContext;
+using MvvmCross.Binding.Bindings.Target.Construction;
+using MvvmCross.Forms.Presenters;
+using MvvmCross.Presenters;
+using Xamarin.Forms;
+
+namespace MvvmCross.Forms.Core
+{
+    public interface IMvxFormsSetupHelper
```
nmilcoff: How about naming this new guy something like `IMvxFormsAgnosticSetup`? `Helper` sounds too general

MvvmCross.Forms/Platforms/Android/Core/MvxFormsAndroidSetup.cs
```diff
@@ -43,6 +54,16 @@ protected override void InitializeIoC()
         {
             base.InitializeIoC();
             Mvx.RegisterSingleton<IMvxFormsSetup>(this);
+
+            Mvx.LazyConstructAndRegisterSingleton<IMvxViewPresenter, MvxFormsAndroidViewPresenter>();
+            Mvx.LazyConstructAndRegisterSingleton<IMvxFormsSetupHelper, MvxFormsSetupHelper>();
+            Mvx.Resolve<IMvxFormsSetupHelper>().InitializeIoC();
```
nmilcoff: `LazyConstructAndRegisterSingleton<IMvxFormsSetupHelper, MvxFormsSetupHelper>` here doesn't make much sense if we're going to call `Resolve<IMvxFormsSetupHelper>` in the next line 🤔

MvvmCross/Core/MvxSetup.cs
```diff
@@ -54,7 +57,10 @@ public static IMvxSetup Instance()
 
         protected abstract IMvxViewsContainer CreateViewsContainer();
 
-        protected abstract IMvxViewDispatcher CreateViewDispatcher();
+        protected virtual IMvxViewDispatcher CreateViewDispatcher()
+        {
+            return Mvx.Resolve<IMvxViewDispatcher>();
```
nmilcoff: There must be a reason for doing / returning just `Mvx.Resolve<IMvxViewDispatcher>` inside a method that claims to create a ViewDispatcher, right? Would be nice to clarify this somehow

MvvmCross/Platforms/Android/Views/MvxAndroidViewDispatcher.cs
```diff
@@ -13,22 +14,17 @@ public class MvxAndroidViewDispatcher
         : MvxAndroidMainThreadDispatcher
         , IMvxViewDispatcher
     {
-        private readonly IMvxAndroidViewPresenter _presenter;
-
-        public MvxAndroidViewDispatcher(IMvxAndroidViewPresenter presenter)
```
nmilcoff: To my eyes the ViewDispatcher object looks easier to override / use if the ViewPresenter is a dependency declared in the constructor.

A property isn't clear enough most of the time, and it can make developers spend quite some time figuring out why their code doesn't work. Is this change necessary?

MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatSetup.cs
nickrandolph: I get you're point - the InitializeIoC doesn't indicate that it's going to register the view presenter. However, if we create methods for anything we register, we'll end up with a ton of overridable methods. imho it seems logical that all registration is done as part of initializing the IoC container.
I guess we could have a single overrideable method which is RegisterLazySingletons (or something like that) - it needs to indicate lazy because there's no guarrantee at the point this is done that all implementations will be available.

MvvmCross.Forms/Core/IMvxFormsSetupHelper.cs
nickrandolph: Hmm perhaps, although I feel that Helper is actually appropriate here since the class that implements this interface is not actually a setup class, it's a class designed to do some of the setup actions on behalf of the actual setup class (ie a helper). I think AgnosticSetup still implies that it is a setup class
Other options?

MvvmCross/Core/MvxSetup.cs
nickrandolph: Good point - worth having some comments/docs around this. The change is that the implementation is lazy registered, which means that calling resolve here will create the instance. This method is overridable to allow for developers to override the creation of the view dispatcher. In reality the best way to override the view presenter is to override the initializeIoC and register a different implementaiton

MvvmCross/Platforms/Android/Views/MvxAndroidViewDispatcher.cs
nickrandolph: Having constructor with parameters means that you have to override both the implementation and wherever in the framework the instance gets created. Preference is to override the implementation and simply register the new implementation. 
Unfortunately interfaces don't allow us to specify constructors with parameters - if they did, that would be best of both worlds

MvvmCross.Forms/Platforms/Android/Core/MvxFormsAndroidSetup.cs
nickrandolph: Agreed, it was more for consistency than anything - I'll adjust this

nickrandolph:
@Cheesebaron @martijn00 @nmilcoff can you guys take a look at this PR and provide feedback on what you like/dislike. imho this improves the setup as it makes it clearer how to override behaviour (ie by registering different impelmentations instead of overriding create methods). Interested to get feedback before I do any more work to split into smaller PRs and/or revert pieces that we're not happy with.

MvvmCross.Forms.Wpf/Platforms/Wpf/Core/MvxFormsWindowsSetup.cs
```diff
@@ -15,14 +15,42 @@
...
+        }
+
+        protected override void RegisterImplementations()
+        {
+            base.RegisterImplementations();
+            Mvx.RegisterSingleton<IMvxFormsSetup>(this);
+
+            Mvx.LazyConstructAndRegisterSingleton<IMvxViewPresenter, MvxFormsWpfViewPresenter>();
+            Mvx.LazyConstructAndRegisterSingleton<IMvxFormsSetupHelper, MvxFormsSetupHelper>();
+            Mvx.Resolve<IMvxFormsSetupHelper>().InitializeIoC();
```
nmilcoff: If `IMvxFormsSetupHelper` is going to be used on the next line, maybe `ConstructAndRegisterSingleton` should be used instead of the lazy version (it also returns the instance back)

MvvmCross.Forms.Wpf/Platforms/Wpf/Core/MvxFormsWindowsSetup.cs
```diff
@@ -15,14 +15,42 @@
...
+            Mvx.LazyConstructAndRegisterSingleton(() => FormsPresenter);
+        }
+
+        protected virtual void RegisterSetupHelper()
+        {
+            Mvx.LazyConstructAndRegisterSingleton<IMvxFormsSetupHelper, MvxFormsSetupHelper>();
```
nmilcoff: `IMvxFormsSetupHelper` is being registered twice 🤔

MvvmCross/Core/MvxSetup.cs
```diff
@@ -50,11 +53,44 @@ public static IMvxSetup Instance()
...
+            Mvx.IoCProvider.LazyConstructAndRegisterSingleton<IMvxSettings, MvxSettings>();
+            Mvx.IoCProvider.LazyConstructAndRegisterSingleton<IMvxPluginManager, MvxPluginManager>();
+            Mvx.IoCProvider.LazyConstructAndRegisterSingleton<IMvxViewModelLoader, MvxViewModelLoader>();
+            Mvx.IoCProvider.LazyConstructAndRegisterSingleton<IMvxNavigationService, MvxNavigationService>();
+            Mvx.IoCProvider.LazyConstructAndRegisterSingleton<IMvxNavigationSerializer, MvxStringDictionaryNavigationSerializer>();
+            Mvx.IoCProvider.LazyConstructAndRegisterSingleton<IMvxViewModelTypeFinder, MvxViewModelViewTypeFinder>();
+            Mvx.IoCProvider.LazyConstructAndRegisterSingleton<IMvxTypeToTypeLookupBuilder, MvxViewModelViewLookupBuilder>();
+            Mvx.IoCProvider.LazyConstructAndRegisterSingleton<IMvxCommandCollectionBuilder, MvxCommandCollectionBuilder>();
+            Mvx.IoCProvider.LazyConstructAndRegisterSingleton<IMvxChildViewModelCache, MvxChildViewModelCache>();
+            Mvx.IoCProvider.LazyConstructAndRegisterSingleton<IMvxStringToTypeParser, MvxStringToTypeParser>();
```
nmilcoff: I kind of prefer to have this registration lines splitted in separate methods (as it is now). That way the behavior is much easier to override and the user customizes only what he/she needs.

MvvmCross.Forms.Wpf/Platforms/Wpf/Core/MvxFormsWindowsSetup.cs
nickrandolph: Oops, will fix

MvvmCross/Core/MvxSetup.cs
nickrandolph: The point of them being lazy is that a developer wanting to override one or more should still call the base method. Then they just register the implementations they want, which will replace the prior registration

nickrandolph:
I agree this PR is huge atm. I'm just looking for feedback on what people like/dislike. The intention is then to break it up and submit as a sequence of PRs.

nickrandolph:
Closing this PR as it's very out of date and has in part been superseded by https://github.com/MvvmCross/MvvmCross/pull/3372

PR #2888, opened 2018-05-17
Issue 2884 reproduce

aoershov:
I add logs to MvxMultipleViewModelCache to see memory leak:
https://github.com/aoershov/MvvmCrossMemoryLeakBrowser/blob/master/MvxMultipleViewModelCacheWithLogs.cs

Steps to reproduce 📜
Press "Open second view"
Press "Open browser"
Browser is shown
Press back
Expected behavior 🤔
No memory leak.

Actual behavior 🐛
From logs I see than FirstViewModel and SecondViewModel added to cache on opening browser,
but never cleaned on back press.
This problem is serious for us because we call Dispose of view model on back press. So we get disposed view model from cache on second opening SecondViewModel.

Configuration 🔧
Version: 6.01

Platform:

 🤖 Android

### DIFF
--- a/Projects/Playground/Playground.Droid/App.cs
+++ b/Projects/Playground/Playground.Droid/App.cs
@@ -1,0 +1,13 @@
+﻿using MvvmCross.ViewModels;
+
+namespace Playground.Droid
+{
+    public class App : MvxApplication
+    {
+        public override void Initialize()
+        {
+            // register the appstart object
+            RegisterCustomAppStart<AppStart>();
+        }
+    }
+}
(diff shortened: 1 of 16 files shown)


nickrandolph:
@aoershov apologies for the delay in getting to this PR - can you please rebase against develop

nickrandolph:
@aoershov when you step through this code, does the GetAndClear method not remove the view model correctly? Or are you looking to extend the cache to support removing the view model on back button?

nickrandolph:
@Cheesebaron @martijn00 are you able to review what's going on here and offer a suggestion on how to fix this issue?

PR #2901, opened 2018-05-23
Typed hint on AppStart and documentation

martijn00:

Adds possibility to use a typed hint at appstart.

Yes, but tried to keep it minimal.

Fixes https://github.com/MvvmCross/MvvmCross/pull/2854
Fixes https://github.com/MvvmCross/MvvmCross/pull/2900

- [x] All projects build

### DIFF
--- a/MvvmCross/ViewModels/IMvxApplication.cs
+++ b/MvvmCross/ViewModels/IMvxApplication.cs
@@ -12,7 +12,7 @@ namespace MvvmCross.ViewModels
 
         void Initialize();
 
-        object Startup(object hint);
+        void Startup();
 
         void Reset();
     }
--- a/MvvmCross/ViewModels/IMvxApplication.cs
+++ b/MvvmCross/ViewModels/IMvxApplication.cs
@@ -16,4 +16,9 @@ namespace MvvmCross.ViewModels
 
         void Reset();
     }
+
+    public interface IMvxApplication<THint> : IMvxApplication
+    {
+        THint Startup(THint hint);
+    }
 }
--- a/MvvmCross/ViewModels/IMvxApplication.cs
+++ b/MvvmCross/ViewModels/IMvxApplication.cs
@@ -12,7 +12,7 @@ namespace MvvmCross.ViewModels
 
         void Initialize();
 
-        void Startup(object hint);
+        object Startup(object hint);
 
         void Reset();
     }
--- a/MvvmCross/ViewModels/MvxAppStart.cs
+++ b/MvvmCross/ViewModels/MvxAppStart.cs
@@ -33,22 +33,21 @@ namespace MvvmCross.ViewModels
 
         protected virtual void Startup(object hint = null)
         {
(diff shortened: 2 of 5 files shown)


MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -57,30 +68,50 @@ protected virtual void Reset()
...
+        protected override void ApplicationStartup(object hint = null)
         {
-            if (hint != null)
+            base.ApplicationStartup(hint);
+            if (hint is TParameter parameter && Application is IMvxApplication<TParameter> typedApplication)
             {
-                MvxLog.Instance.Trace("Hint ignored in default MvxAppStart");
-            }
+                //There is no way to pass the hint back
+                var typedHint = typedApplication.Startup(parameter);
```
sylvain-toptal: What's the point of doing this since we do nothing with the result?  Should it not be above the base call and pass the hint to the `base.ApplicationStartup` instead?

MvvmCross/ViewModels/IMvxApplication.cs
```diff
@@ -12,8 +12,13 @@ public interface IMvxApplication : IMvxViewModelLocatorCollection
 
         void Initialize();
 
-        void Startup(object hint);
+        object Startup(object hint);
```
nickrandolph: @martijn00 I'm not sure why we're keeping the object parameter and return type since this can be done by using MvxApplication<THint> with THint set to object

MvvmCross/ViewModels/MvxAppStart.cs
```diff
@@ -57,30 +67,52 @@ protected virtual void Reset()
...
-        protected virtual void NavigateToFirstViewModel(object hint)
+        protected override object ApplicationStartup(object hint = null)
         {
-            if (hint != null)
+            var applicationHint = base.ApplicationStartup(hint);
+            if (applicationHint is TParameter parameter && Application is IMvxApplication<TParameter> typedApplication)
             {
-                MvxLog.Instance.Trace("Hint ignored in default MvxAppStart");
+                //There is no way to pass the hint back
+                return typedApplication.Startup(parameter);
```
nickrandolph: Given this issue, perhaps we should consider passing around a new type, IStartupHint (and associated IStartupHint<THint>) that has a property Hint. Rather than passing back an updated hint, the Hint within the IStartupHint object could be updated.

MvvmCross/ViewModels/IMvxApplication.cs
martijn00: I just wanted to keep this to not break to many things. But it doesn't make a lot of sense. Developers could even just put `object` as type on the typed Startup.

MvvmCross/ViewModels/MvxAppStart.cs
martijn00: This was a leftover comment, not an issue anymore since breaking the interface method.

PR #2949, opened 2018-06-11
[WIP] Adding SignClient, updating unit test runner + security fixes

Cheesebaron:
Adds Sign Client so we can sign our NuGet packages with the signing cert we get from dotnet fdn.

Right now the step is disabled as creds are returning 401. Have e-mailed @onovotny 

Also removed binlogs from being generated. We can enable them when we need more details when builds fail.

Also switched to `dotnet test` instead of `dotnet xunit` as per @onovotny's request below.

No signing

Signing of NuGet packages

No

- [x] All projects build

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,8 +1,7 @@
 #tool nuget:?package=GitVersion.CommandLine
 #tool nuget:?package=vswhere
-#addin nuget:?package=Cake.Figlet
-#addin nuget:?package=Cake.Incubator&version=1.7.1
-#addin nuget:?package=Cake.Git&version=0.16.1
+#addin nuget:?package=Cake.Figlet&version=1.1.0
+#addin nuget:?package=Cake.Git&version=0.17.0
 #addin nuget:?package=Polly
 
 using Polly;
--- a/build.cake
+++ b/build.cake
@@ -3,6 +3,7 @@
 #addin nuget:?package=Cake.Figlet&version=1.1.0
 #addin nuget:?package=Cake.Git&version=0.17.0
 #addin nuget:?package=Polly
+#tool nuget:?package=SignClient&version=0.9.1&include=/tools/netcoreapp2.0/SignClient.dll
 
 using Polly;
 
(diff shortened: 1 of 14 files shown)


build.cake
```diff
@@ -1,9 +1,9 @@
 #tool nuget:?package=GitVersion.CommandLine
 #tool nuget:?package=vswhere
-#addin nuget:?package=Cake.Figlet
-#addin nuget:?package=Cake.Incubator&version=1.7.1
-#addin nuget:?package=Cake.Git&version=0.16.1
+#addin nuget:?package=Cake.Figlet&version=1.1.0
+#addin nuget:?package=Cake.Git&version=0.17.0
 #addin nuget:?package=Polly
+#tool nuget:?package=SignClient&version=0.9.1&include=/tools/netcoreapp2.0/SignClient.dll
```
onovotny: I would not do it this way. It's now a .NET Core tool. I would do it more like this: https://github.com/reactiveui/refit/blob/master/build/Sign-Package.ps1#L10

build.cake
```diff
@@ -133,7 +136,7 @@ Task("UnitTest")
         try 
         {
             DotNetCoreTool(project,
-                "xunit",  "-fxversion 2.0.0 --no-build -parallel none -configuration " + 
+                "xunit",  "-fxversion 2.1.0 --no-build -parallel none -configuration " +
```
onovotny: dotnet xunit is deprecated btw. Use dotnet test.

build.cake
Cheesebaron: Alrighty, will add that to our appveyor.yml

build.cake
Cheesebaron: OK, will try that, thanks for letting us know. I think the reason why we were using dotnet xunit, because dotnet test didn't support adding configuration and other parameters.

onovotny:
One thing to be aware of, and I think you're okay already, is that you should not have diagnostic/binlog artifacts as outputs. Can you confirm?

build.cake
onovotny: you should be able to add parameters to dotnet test...?

`dotnet test .\foo.csproj -c Release /p:SomeOtherParam=foo`

Cheesebaron:
@onovotny actually we do have that currently. But only when builds fail. I will remove that.

PR #2959, opened 2018-06-12
Binding to Child View's BindingContextProperty

borbmizzet:
Adds a demo to the playground for binding a child view's BindingContextProperty to a parent view model's member.

Xaml binding a child view's BindingContextProperty no longer properly synchronizes between parent view model and child view.

no

Playground.Forms.Droid ->Show Content View -> Tap "Change Child2 Text"

- [x] All projects build

### DIFF
--- a/Projects/Playground/Playground.Forms.UI/Pages/ParentContentPage.xaml
+++ b/Projects/Playground/Playground.Forms.UI/Pages/ParentContentPage.xaml
@@ -8,9 +8,11 @@
              x:Name="ParentPage">
     <ContentPage.Content>
         <StackLayout>
-            <views:ChildContentView ViewModel="{Binding ChildViewModel1}" IsEnabled="{Binding ViewModel.ChildViewModelEnabled, Source={x:Reference ParentPage}}"/>
-            <Button Text="Change Child Text" Command="{Binding ChangeButtonCmd}" />
-            <Button Text="Toggle Child Enabled" Command="{Binding ToggleChildEnabledCmd}" />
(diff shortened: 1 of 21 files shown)


borbmizzet:
Hold on, there is a typo in my demo that I'm about to push a correction for.

Projects/Playground/Playground.Forms.UI/Pages/ParentContentPage.xaml
```diff
@@ -8,9 +8,11 @@
              x:Name="ParentPage">
     <ContentPage.Content>
         <StackLayout>
-            <views:ChildContentView ViewModel="{Binding ChildViewModel1}" IsEnabled="{Binding ViewModel.ChildViewModelEnabled, Source={x:Reference ParentPage}}"/>
-            <Button Text="Change Child Text" Command="{Binding ChangeButtonCmd}" />
-            <Button Text="Toggle Child Enabled" Command="{Binding ToggleChildEnabledCmd}" />
+            <views:ChildContentView ViewModel="{Binding Source={x:Reference ParentPage}, Path=ViewModel.ChildViewModel1}" IsEnabled="{Binding ViewModel.ChildViewModelEnabled, Source={x:Reference ParentPage}}"/>
```
nickrandolph: Is there a reason why "Source" needs to be specified for either of these databinding expressions?

borbmizzet: Not in this specific context, but if Source is not specified then the binding works a grand total of one time, after which the child view's base.BindingContext is no longer the same as the parent view's ViewModel and it wouldn't be able to find the property ChildViewModel1 in the class 'ChildViewModel'.

nickrandolph:
@borbmizzet can you check over my changes and confirm that it fixes the issue with the binding? I've changed the binding expressions in the page to what they should be - there shouldn't be a need to reference the page as the source for all the expressions.

nickrandolph:
@martijn00 can you review my changes. Was there a reason for the original code?

Issue #2924, opened 2018-06-01
IMvxNavigationService.Navigate<TViewModel, TParam, TResult> deadlock if the back button is used


Sebastian1989101:
Is there any workaround or something I can do, so that I can still use this function? I don't like the MvxMessenger way which I use currently as an alternative (because it can easily fail due to lacks). 

### Steps to reproduce :scroll:

1. Use NavigationPage Back Button instead of a Close-Call with the result 

### Expected behavior :thinking:
Should just returns a default value 

### Actual behavior :bug:
Causes a deadlock within the MvxAsyncCommand if the NavigationPage back button is used and no return value is passed. 

### Configuration :wrench:

**Version:** 6.0.1

**Platform:** 
- [ ] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [x] :monkey: Xamarin.Forms

deepblueday:
Bump. This remains a problem with MVVMCross 7.0.1.

Issue #2833, opened 2018-04-25
Master Detail never cancel CloseCompletionSource


Thetyne:
### Steps to reproduce :scroll:

1. Make a Master detail Forms project on iOS. i haven't test on Android.

2. Add 2 bouton inside the master page who call one view model (with detail page with or without navigation page) like 
`var result = await NavigationService.Navigate<ViewModel, TModel, TResult>(data);`

3. Click on first button then one second button.

4. It appear that TResult CloseCompletionSource is never Canceled.

5. Same thing when i press on back button (with a navigation page on detail page)

### Expected behavior :thinking:
CloseCompletionSource must be canceled when mvvmcross pop View.
Normally, it must be call ViewModel.ViewDestroy method who called by UIViewController.DidMoveToParentViewController.

UIViewController.DidMoveToParentViewController is overrided by mvvmcross
```
public override void DidMoveToParentViewController(UIViewController parent)
        {
            base.DidMoveToParentViewController(parent);
            if(parent == null)
                ViewModel?.ViewDestroy();
        }
``` 

Problem: Maybe a memory leak but also my command is never released.

For test, i have override MvxFormsPagePresenter and the CloseMasterDetailPage method for call 
`ViewModel?.Destroy();` inside and the token is canceled.

### Actual behavior :bug:
it seem MvxFormsPagePresenter, never call UIViewController.DidMoveToParentViewController.

### Configuration :wrench:

**Version:** 6.0

**Platform:** 
- [x] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [x] :monkey: Xamarin.Forms

Thetyne:
I have fixed in my code by create my own MvxFormsPagePresenter and change code inside "ReplacePageRoot":


 ```
if (parent is MasterDetailPage rootMasterDetail)
                {
                    if (attribute is MvxMasterDetailPagePresentationAttribute masterDetailAttribute)
                    {
                        if (masterDetailAttribute.Position == MasterDetailPosition.Master)
                        {
                            rootMasterDetail.Master = page;
                        }
                        else if (masterDetailAttribute.Position == MasterDetailPosition.Detail)
                        {
                            rootMasterDetail.Detail = page;
                        }
                    }
                }
```

replace to

```
if (parent is MasterDetailPage rootMasterDetail)
                {
[...]

PR #3006, opened 2018-07-11
Make sure navigation is cancel when requested

martijn00:

Mostly a feature

Sometimes canceling a cancellationtoken wouldn't really cancel it.

It works now!

It breaks the interface, but the parameter is optional

Fixes https://github.com/MvvmCross/MvvmCross/issues/2924
Fixes https://github.com/MvvmCross/MvvmCross/issues/2833

- [x] All projects build

### DIFF
--- a/MvvmCross/Navigation/EventArguments/NavigateEventArgs.cs
+++ b/MvvmCross/Navigation/EventArguments/NavigateEventArgs.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Threading;
 using MvvmCross.ViewModels;
 
 namespace MvvmCross.Navigation.EventArguments
@@ -13,11 +14,13 @@ namespace MvvmCross.Navigation.EventArguments
         {
         }
 
-        public NavigateEventArgs(IMvxViewModel viewModel)
+        public NavigateEventArgs(IMvxViewModel viewModel, CancellationToken cancellationToken = default(CancellationToken))
         {
             ViewModel = viewModel;
+            CancellationToken = cancellationToken;
         }
 
         public IMvxViewModel ViewModel { get; set; }
+        public CancellationToken CancellationToken { get; set; }
     }
 }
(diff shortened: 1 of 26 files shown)


MvvmCross/Navigation/EventArguments/NavigateEventArgs.cs
```diff
@@ -13,11 +14,13 @@ public NavigateEventArgs()
         {
         }
 
-        public NavigateEventArgs(IMvxViewModel viewModel)
+        public NavigateEventArgs(IMvxViewModel viewModel, CancellationToken cancellationToken = default(CancellationToken))
```
nickrandolph: Why are we doing it this way - shouldn't we just make NavigateEventArgs inherit from CancelEventArgs?

MvvmCross/Navigation/MvxNavigationService.cs
```diff
@@ -256,37 +256,53 @@ public virtual Task<bool> CanNavigate(Type viewModelType)
...
         protected virtual async Task Navigate(MvxViewModelRequest request, IMvxViewModel viewModel, IMvxBundle presentationBundle = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            var args = new NavigateEventArgs(viewModel);
+            bool hasNavigated = false;
+            if (cancellationToken != default(CancellationToken))
+            {
+                cancellationToken.Register(async () =>
+                {
+                    if(hasNavigated)
+                        await Close(viewModel);
```
nickrandolph: This is dangerous. We have full async support within the navigation, so this is no longer required. If navigation has completed, it should be the responsibility of the destination page to close and return to previous page, rather than some built in mechanism. This is just going to lead to some nasty race conditions

nmilcoff: Good point @nickrandolph. Can you elaborate a bit more on why was this added @martijn00?

nmilcoff: Good point @nickrandolph. Can you elaborate a bit more on why was this added @martijn00?

nickrandolph: @nmilcoff I think this block of code (repeated in the Navigate method that returns a result) shouldn't be here. It should only be there for Navigations that return a result, so that the caller cancel the navigation. Think of the scenario where you prompt the user for some information by navigating to a page that will return a result - you might then want to cancel this after a period of time. In the case of the Navigate method that returns a result, we should be checking to see if the result has been returned (task completion source has completed) and not navigate if the result has been returned.

martijn00:
@thongdoan Can you open a new issue for this?

thongdoan:
yes, let me do it

PR #3013, opened 2018-07-16
[WIP] UWP Suspension handling

### LINKED ISSUE #2881: Xamarin Forms UWP Playground application crash on resume

1. Compile & run MvvmCross Xamarin Forms Playground application

2.Suspend application

3. Resume application

Application should not crash

### Actual behavior :bug:
Application crash with exception:
"Element is already the child of another element."

Stack trace:
   at System.Runtime.InteropServices.WindowsRuntime.IVector`1.Append(T value)
   at System.Runtime.InteropServices.WindowsRuntime.VectorToCollectionAdapter.Add[T](T item)
   at Xamarin.Forms.Platform.UWP.Platform.<SetCurrent>d__52.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.AsyncMethodBuilderCore.<>c.<ThrowAsync>b__6_0(Object state)
   at System.Threading.WinRTSynchronizationContextBase.Invoker.InvokeCore()

Call stack:
>	Playground.Forms.Uwp.exe!Playground.Forms.Uwp.App.InitializeComponent.AnonymousMethod__2_0(object sender, Windows.UI.Xaml.UnhandledExceptionEventArgs e) Line 53	C#
 	[Native to Managed Transition]	
 	[Managed to Native Transition]	
 	System.Private.CoreLib.dll!System.Runtime.InteropServices.WindowsRuntime.WindowsRuntimeMarshal.RoReportUnhandledError(System.Runtime.InteropServices.WindowsRuntime.IRestrictedErrorInfo error)	Unknown
 	System.Private.CoreLib.dll!System.Runtime.InteropServices.WindowsRuntime.WindowsRuntimeMarshal.ReportUnhandledError(System.Exception e)	Unknown
 	System.Runtime.WindowsRuntime.dll!System.Threading.WinRTSynchronizationContextBase.Invoker.InvokeCore()	Unknown
 	System.Runtime.WindowsRuntime.dll!System.Threading.WinRTSynchronizationContextBase.Invoker.InvokeInContext(object thisObj)	Unknown
 	System.Private.CoreLib.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state)	Unknown
 	System.Runtime.WindowsRuntime.dll!System.Threading.WinRTSynchronizationContextBase.Invoker.Invoke()	Unknown

### Configuration :wrench:

**Version:** 6.0.1

**Platform:** 
- [ ] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [X] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [X] :monkey: Xamarin.Forms

### LINKED ISSUE #3012: Deadlock after configuration change in Xamarin.Forms
## 🐛 Bug Report

I have discovered a deadlock issue that prevents MvvmCross.Forms from handling configuration changes on Android (but could prove problematic in other places as well, actually even in non-Forms projects).

### Expected behavior

When I put MvvmCross app in split view mode or freeform mode on Android, the app should reinitialize and work as expected.

### Actual behavior

When app is put in split view, initialization starts, but ends with a deadlock in `MvxAppStart` on [this line](https://github.com/MvvmCross/MvvmCross/blob/8c787b2ee6993476948ac3467f4ae839a6f8ccc1/MvvmCross/ViewModels/MvxAppStart.cs#L32).

### Reproduction steps

1. Run the Playground.Forms.Droid project
2. Open Task view on Android, and drag and drop the app into either the splitview mode or freeform mode
3. Observe that execution gets stuck after a few moments
4. Pause debugging and see that it waits on `StartAsync(hint).GetAwaiter().GetResult();` in `MvxAppStart`

### Investigation results

First and foremost - I have used `Forms` and splitview configuration change as an example, but I am quite confident that this issue can appear unexpectedly on other platforms as well.

Using `StartAsync(hint).GetAwaiter().GetResult();` is problematic because as soon as the execution hits an "actual" `await` which runs on different thread, UI thread enters `GetResult()` and waits for the result in blocking fashion. When the other thread finishes running and tries to return to the UI thread, it cannot do so, because `GetResult()` is already blocking it and hence creates a deadlock.

I have stepped through the code several times to pinpoint the actual location where this problem occurs and I have probably found it. In the first run we are okay, we get to [this line](https://github.com/MvvmCross/MvvmCross/blob/c6ef1953d022585df9fb0643d786c7f510c11b15/MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs#L705)and because the `navigationRootPage` is `null`, it is created and the page is synchronously loaded.

The real trouble begins when configuration changes, app is reinitialized and execution again gets to [this line](https://github.com/MvvmCross/MvvmCross/blob/c6ef1953d022585df9fb0643d786c7f510c11b15/MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs#L705). This time the situation is different, as `navigationRootPage` already exists, so we go to [this line](https://github.com/MvvmCross/MvvmCross/blob/c6ef1953d022585df9fb0643d786c7f510c11b15/MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs#L719) and execute `await navigationRootPage.PushAsync(page, attribute.Animated);`. Now if you look into the Xamarin.Forms repo and find the implementation of this method, you can see that it will lead to `NavigationRenderer` and specifically [here where an animation on another thread is started](https://github.com/xamarin/Xamarin.Forms/blob/0b1400a5d852f2d54e0f9bec3ef3a01f5061ad02/Xamarin.Forms.Platform.Android/Renderers/NavigationRenderer.cs#L250) this is the actual cause of the deadlock in our Playground project.

If you want to cause the deadlock much more easily, you can just put `await Task.Delay(1000)` in the `StartAsync` method of `MvxAppStart`. The same situation will happen.

### Possible solutions

Instead of `GetAwaiter().GetResult()` use something like:

```
try
{
    await StartAsync(hint);
}
catch (Exception ex)
{
    MvxLog.Instance.ErrorException(ex.Message, ex);
}
```

The disadvantage here is that the execution of `Start` will no longer wait for the `StartAsync` to complete which can potentially cause issues.

The best solution would probably be to remove the synchronous `Start` method altogether, but that would require to ensure correct functionality on platforms like iOS which expect a `ViewController` to be present immediately after startup.

### Configuration

**Version:** 6.x

**Platform:** 
- [ ] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [x] :monkey: Xamarin.Forms

MartinZikmund:

Goal is to properly implement suspend/restore in UWP.

This will fix #2881 .

Suspension is handled correctly, but suspended state is then restored while app is resumed, which is not correct, as when `OnResuming` is called, it means the app is actually fully loaded in memory and therefore there is no reason to restore state (including navigation).

The old behavior has been removed, but **restore is not yet implemented** (hence WIP :-) ). [UWP documentation](https://docs.microsoft.com/en-us/windows/uwp/launch-resume/app-lifecycle#app-launch) recommends to restore when the previous execution state is `Terminated`, but this cannot be done before `IMvxAppStart` is finished, as `MvvmCross` is not initialized and restoring state from suspension manager would fail.

The only solution I currently see is to let the user implement the restore from `IMvxSuspensionManager` in `MvxAppStart` if they so choose, because if we force it upon the user, the suspended navigation state would replace viewmodel navigated to in `MvxAppStart`. However, I am afraid that doing `async` work there will cause a deadlock the same way as in #3012 . So the goal might rather be to resolve #3012 first. Anyway, this is open for discussion and I welcome any input and ideas :-) .

Yes, resuming no longer does navigation stack restore, which should, however, avoid issues like #2881 and ideally should not cause problems.

The `Resume` method no longer takes an `IMvxSuspensionManager` as it should not be used during `Resuming`.

-

- Default suspension implementation in UWP samples - [here](https://github.com/Microsoft/Windows-universal-samples/blob/master/Samples/BasicSuspension/cs/App.xaml.cs)
- UWP app lifecycle description - [here](https://docs.microsoft.com/en-us/windows/uwp/launch-resume/app-lifecycle)

- [x] All projects build

### DIFF
--- a/MvvmCross/Platforms/Uap/Views/MvxApplication.cs
+++ b/MvvmCross/Platforms/Uap/Views/MvxApplication.cs
@@ -94,11 +94,6 @@ namespace MvvmCross.Platforms.Uap.Views
                 Window.Current.Content = rootFrame;
             }
 
-            if (activationArgs.PreviousExecutionState == ApplicationExecutionState.Terminated)
-            {
-                OnResumeFromTerminateState();
-            }
-
             RootFrame = rootFrame;
 
             return rootFrame;
@@ -107,10 +102,6 @@ namespace MvvmCross.Platforms.Uap.Views
         protected virtual Frame CreateFrame()
         {
             return new Frame();
-        }
-
-        protected virtual void OnResumeFromTerminateState()
-        {
         }
 
         protected virtual void OnNavigationFailed(object sender, NavigationFailedEventArgs e)
@@ -133,16 +124,13 @@ namespace MvvmCross.Platforms.Uap.Views
             return Task.CompletedTask;
         }
 
-        private async void OnResuming(object sender, object e)
+        private void OnResuming(object sender, object e)
         {
-            var suspension = Mvx.IoCProvider.GetSingleton<IMvxSuspensionManager>();
-            await Resume(suspension);
-            await suspension.RestoreAsync();
+            Resume(e);
         }
 
-        protected virtual Task Resume(IMvxSuspensionManager suspensionManager)
-        {
-            return Task.CompletedTask;
+        protected virtual void Resume(object e)
+        {           
         }
 
         protected virtual void RegisterSetup()


MartinZikmund:
@nickrandolph May I ask for your opinion on how this should be approached?

nickrandolph:
I don't believe the issue here is related to resume events. It appears to be that the UWP host page isn't being reused despite setting NavigationCacheMode to Required. I've shifted setting this property into the base constructor, instead of the Loaded event. This seems to fix the issue as the page doesn't get recreated if it's still in memory.
See https://github.com/MvvmCross/MvvmCross/pull/3023

MartinZikmund:
@nickrandolph Although it does fix the issue, the main fact is that suspension manager shouldn't do anything on the `Resuming` event. Suspension manager is there only for the event when app gets terminated because of low memory and its state needs to be restored - which is a case which cannot be done with MvvmCross with current implementation, unless you implement it the right way as I mentioned in the description. See [this](https://docs.microsoft.com/en-us/windows/uwp/launch-resume/app-lifecycle#app-resume):


> When an app is resumed from the Suspended state, it enters the Running in background state and the system restores the app where it left off so that it appears to the user as if it has been running all along. No app data stored in memory is lost. Therefore, most apps don't need to restore state when they are resumed though they should reacquire any file or device handles that they released when they were suspended, and restore any state that was explicitly released when the app was suspended.

> If the suspended app was terminated, there is no Resuming event and instead OnLaunched() is called with an ApplicationExecutionState of Terminated. Because you saved your state when the app was suspended, you can restore that state during OnLaunched() so that your app appears to the user as it was when they switched away from it.

Also see [the official sample for handling suspension](https://github.com/Microsoft/Windows-universal-samples/blob/master/Samples/BasicSuspension/cs/App.xaml.cs) - the `Suspending` event is handled, but `Resuming` is not - again, because when `Resuming`, the app is still completely in memory and there is no reason to do any restore.

PR #3023, opened 2018-07-22
Fixing Forms reload issue

nickrandolph:
Bug fix

Forms UWP application crashes on resume

Forms UWP application no longer crashes on resume

No

Launch Playground UWP Forms application
Suspend (using VS lifecycle events, or minimizing UWP application)
Resume (using VS lifecycle events, or restoring UWP application)

https://github.com/MvvmCross/MvvmCross/issues/2881

- [ X ] All projects build
- [ X ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contribute/mvvmcross-docs-style-guide))

### DIFF
--- a/MvvmCross.Forms/Platforms/Uap/Views/MvxFormsWindowsPage.cs
+++ b/MvvmCross.Forms/Platforms/Uap/Views/MvxFormsWindowsPage.cs
@@ -1,8 +1,7 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MS-PL license.
 // See the LICENSE file in the project root for more information.
 
-using System.Threading.Tasks;
 using MvvmCross.Forms.Platforms.Uap.Presenters;
 using MvvmCross.Forms.Presenters;
 using MvvmCross.ViewModels;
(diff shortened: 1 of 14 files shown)


MartinZikmund:
@nickrandolph Although it does fix the issue, the main fact is that suspension manager shouldn't do anything on the `Resuming` event. Suspension manager is there only for the event when app gets terminated because of low memory and its state needs to be restored - which is a case which cannot be done with MvvmCross with the current implementation, unless you implement it the right way as I mentioned in the description of #3013. See [this](https://docs.microsoft.com/en-us/windows/uwp/launch-resume/app-lifecycle#app-resume):

> If the suspended app was terminated, there is no Resuming event and instead OnLaunched() is called with an ApplicationExecutionState of Terminated. Because you saved your state when the app was suspended, you can restore that state during OnLaunched() so that your app appears to the user as it was when they switched away from it.

Also see [the official sample for handling suspension](https://github.com/Microsoft/Windows-universal-samples/blob/master/Samples/BasicSuspension/cs/App.xaml.cs) - the `Suspending` event is handled, but `Resuming` is not - again, because when `Resuming`, the app is still completely in memory and there is no reason to do any restore.

MartinZikmund:
Ok, I will raise it separately, I think it will be useful to handle the terminated state properly 👍

PR #3025, opened 2018-07-23
Add IoCProvider property to MvxViewModel

borbmizzet:

Adds `IoCProvider` property to `MvxViewModel`, so it can be accessed like `NavigationService` and `Log`.

`IMvxIoCProvider` must be passed in view model constructor and stored or use `Mvx.IoCProvider` to reference `IoCProvider` from view models.

Allows access to `IoCProvider` from view models through base class property.

No.

Since MvxViewModel.NavigationService is now set by using IoCProvider property, just open a playground project and navigate to another page

- [x] All projects build

### DIFF
--- a/MvvmCross/ViewModels/IMvxIoCViewModel.cs
+++ b/MvvmCross/ViewModels/IMvxIoCViewModel.cs
@@ -1,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MS-PL license.
+// See the LICENSE file in the project root for more information.
+
+using MvvmCross.IoC;
+
+namespace MvvmCross.ViewModels
+{
+    public interface IMvxIoCViewModel : IMvxViewModel
+    {
+        IMvxIoCProvider IoCProvider { get; set; }
+    }
+}
--- a/MvvmCross/ViewModels/MvxViewModel.cs
+++ b/MvvmCross/ViewModels/MvxViewModel.cs
@@ -3,14 +3,16 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Threading.Tasks;
+using MvvmCross.IoC;
 using MvvmCross.Logging;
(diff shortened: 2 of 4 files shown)


martijn00:
I'm thinking we should maybe bring all those extras into 1 interface instead of having a new one for every feature we add.

borbmizzet:
@martijn00 Should that be done in a new PR after merging this one in?

borbmizzet:
seems that second commit broke the tests somehow.  fixing it now

borbmizzet:
The tests still run fine on my machine.  Probably a race condition in the tests.  The changes in my PR should have absolutely no effect on the test that failed.

nickrandolph:
I'm not opposed to having the IMvxIoCViewModel interface but I don't believe the MvxViewModel should expose this as a property - this encourages a very bad coding practice of using the IoCProvider directly to resolve an interface. This is something that should really be avoided.

We currently have code in the properties in MvxViewModel that uses IoCProvider as a fallback, just in case the property isn't set. This code really shouldn't be there - from what I recall, it was added to get the test cases to pass again.

To your point @martijn00, the reason that there are separate interfaces for each of these properties is that they are non-essential and that should someone decide they want to create their own IMvxViewModel, they should be able to do so without have to implement all these additional properties.

In summary:
- I don't believe we should move this forward
- We should work out a better solution, rather than calling Resolve in the properties to enable the test cases to work
- We should remove the fallback calls to Resolve.

Cheesebaron:
I agree with @nickrandolph here. This seems very much like an antipattern. Just like having NavigationService and LogProvider resolved behind your back.

I would much rather prefer to have these as explicit dependencies, rather than implicit and unclear behavior from a public API point of view.

nickrandolph:
@borbmizzet going to close this PR for the moment. We'll revisit how we make it easier for devs to take advantage of core plumbing (eg IoCProvider) going forward

PR #3027, opened 2018-07-23
Fix MvxAutoCompleteTextView not returning results

### LINKED ISSUE #3008: MvxAutoCompleteTextView PartialText never changes after initial setting
### Expected behavior
PartialText's binding should change when the user types

### Reproduction steps
Create MvxAutoCompleteTextView and bind PartialText.
Type some stuff.

### Configuration

**Version:** 6.x

**Platform:** 
- [ ] :iphone: iOS
- [X] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

Cheesebaron:
Bug fix

MvxAutoCompleteTextView doesn't return any results at all when typing in it

Returns matching results

No

1. Add a MvxAutoCompleteTextView
2. Bind `ItemsSource` to list of items that implement `ToString()` for filtering
3. Type stuff that the items contain

This now also allows someone to provide different filtering logic, rather than just matching on `ToString()` by setting `FilterPredicate` on the `MvxFilteringAdapter`.

Fixes #3008 

- [x] All projects build

### DIFF
--- a/MvvmCross/Binding/Extensions/MvxEnumerableExtensions.cs
+++ b/MvvmCross/Binding/Extensions/MvxEnumerableExtensions.cs
@@ -92,13 +92,12 @@ namespace MvvmCross.Binding.Extensions
             if (items == null)
                 return null;
 
-            var enumerator = items.GetEnumerator();
             var matchList = new List<object>();
-            while(enumerator.MoveNext())
+            foreach (var item in items)
             {
-                var match = predicate(enumerator.Current);
+                var match = predicate(item);
                 if (match)
-                    matchList.Add(enumerator.Current);
+                    matchList.Add(item);
             }
 
             if (matchList.Count == 0)
--- a/MvvmCross/Binding/Extensions/MvxEnumerableExtensions.cs
+++ b/MvvmCross/Binding/Extensions/MvxEnumerableExtensions.cs
@@ -2,7 +2,9 @@
 // The .NET Foundation licenses this file to you under the MS-PL license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Collections;
+using System.Collections.Generic;
 
 namespace MvvmCross.Binding.Extensions
 {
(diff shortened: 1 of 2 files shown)


MvvmCross/Binding/Extensions/MvxEnumerableExtensions.cs
```diff
@@ -84,5 +86,25 @@ public static object ElementAt(this IEnumerable items, int position)
 
             return enumerator.Current;
         }
+
+        public static IEnumerable Filter(this IEnumerable items, Func<object, bool> predicate)
+        {
+            if (items == null)
+                return null;
```
nickrandolph: Would it be better to return empty array instead of null - it'll cause few breaks for developers that fail to do a null check on the returned result

MvvmCross/Binding/Extensions/MvxEnumerableExtensions.cs
```diff
@@ -84,5 +86,25 @@ public static object ElementAt(this IEnumerable items, int position)
...
+
+            var enumerator = items.GetEnumerator();
```
nickrandolph: Whilst perhaps not as efficient from a performance perspective, a simple Where linq clause would make this easier to follow?

MvvmCross/Platforms/Android/Binding/Views/MvxFilteringAdapter.cs
```diff
@@ -148,6 +156,35 @@ public override Object GetItem(int position)
...
+                var element = FilteredItemsSource?.ElementAt(position);
+                return element;
+            }
+        }
+
+        public override int GetPosition(object item)
+        {
+            lock (_syncLock)
+            {
+                var pos = FilteredItemsSource?.GetPosition(item) ?? 0;
```
nickrandolph: Should default be 0 or -1

MvvmCross/Binding/Extensions/MvxEnumerableExtensions.cs
Cheesebaron: Linq queries do not work on non-generic IEnumerable, that is why we have these extensions in the first place.

nickrandolph: doh! yeh of course, it requires generics. We should be able to at least use a basic for loop here, right?

MvvmCross/Platforms/Android/Binding/Views/MvxFilteringAdapter.cs
Cheesebaron: Hmm, this is used in `GetItem` and `GetRawItem` to get the element. Not sure what the appropriate default value should be here. If it is 0 and the ItemsSource contains 0 items you will get a Index error, the same with -1. However, if it contains items 0 could return a wrong item and -1 would still throw a Index error. Not sure what is preferred here.

PR #3030, opened 2018-07-24
[WIP] Fix stack overflow in MvxFormsPagePresenter.BuildPageTree

bruzkovsky:
Bug fix

Stack overflow

Assume the following setup:

A `NavigationPage` presenting a modal `MasterDetailPage` (via `MvxModalPresentationAttribute`), which in turn has a `Detail` set to a NavigationPage. The second NP contains the MDP in its `ModalStack`, for whatever reason.

Now the current implementation recurses through the navigation stack, finds a `MasterDetailPage`, recurses through the `Detail` stack, finds a `NavigationPage`, checks the modal stack, finds a `MasterDetailPage`, recurses through the `Detail` stack... and oops, stack overflow. So I added another check to see if the modal is a MDP and is already containing the current `rootPage` in which case we just continue recursing happily.

As this is just adding 3 lines and removing 2, I thought to make a PR like this, if you need anything more please tell me, I am new to contributing to MvvmCross...

No

Occurred on Android when showing a MasterDetailPage modally as described above, then press the hardware back key

None

- [x] All projects build

### DIFF
--- a/MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs
+++ b/MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs
@@ -751,7 +751,11 @@ namespace MvvmCross.Forms.Presenters
             {
                 // Check for modals
                 var topMostModal = navigationPage?.Navigation?.ModalStack?.LastOrDefault();
-                if (topMostModal != null && topMostModal != navigationPage)
+				var modalNotRootPage = topMostModal != null && topMostModal != navigationPage;
+				var modalNotContainingRootPage = topMostModal is MasterDetailPage md &&
+				                                 md.Master != navigationPage &&
+				                                 md.Detail != navigationPage;
+                if (modalNotRootPage && modalNotContainingRootPage)
                 {
                     var currentModalNav = BuildPageTree(topMostModal);
                     if (currentModalNav != null) return (List<Page> list) => currentModalNav(pageListBuilder(list));


nickrandolph:
This will fix the particular scenario listed but there's a more generic problem with the modal stack: There's only one modal stack which is returned by all navigationpages. This means that if you have a navigationpage within the modal stack, there's the potential for a stack overflow.
I'll need some more time in order to review this and see what other scenarios we need to handle

bruzkovsky:
@nickrandolph yes I thought so too, after investigating further... There would be a problem with `MultiPage` containing a `NavigationPage` too!

Additionally, there is a problem with navigating in this nested `NavigationPage`, because the current implementation closes all Modals before navigating:
```c#
protected virtual async Task<Page> CloseAndCreatePage(Type view,
            MvxViewModelRequest request,
            MvxPagePresentationAttribute attribute,
            bool closeModal = true,
            bool closePlatformViews = true,
            bool showPlatformViews = true)
        {
            // always true except when modal is shown
            if (closeModal)
                await CloseAllModals();
```
I would open an issue that concerns this problem as well the initial stack overflow for modal navigation generically, like "Add support for stacked NavigationPages", if that's ok?

nickrandolph:
Yes, definitely raise the new issue. I'm going to reopen this PR as I want to fix the BuildPageTree method and I will use this PR as the basis

Issue #2969, opened 2018-06-20
MvxAndroidSetupSingleton.EnsureSingletonAvailable(ApplicationContext) throws null reference


vatsalyagoel:
Hello,
I am having an issue when sending push notifications while the app is closed
## 🐛 Bug Report

```
Exception android.runtime.JavaProxyThrowable: System.ArgumentNullException: Value cannot be null. Parameter name: source at System.Linq.Enumerable.TryGetFirst[TSource] (System.Collections.Generic.IEnumerable`1[T] source, System.Boolean& found)
[0x0000d] in <1ea41f79845a40618089713a1d9403c4>:0 at System.Linq.Enumerable.FirstOrDefault[TSource] (System.Collections.Generic.IEnumerable`1[T] source)
[0x00000] in <1ea41f79845a40618089713a1d9403c4>:0 at MvvmCross.Forms.Platforms.Android.Core.MvxFormsAndroidSetup.get_FormsApplication ()
[0x00036] in <f9ca34d669a54607bed0e9bd0795a0d4>:0 at MvvmCross.Forms.Platforms.Android.Core.MvxFormsAndroidSetup.CreateViewPresenter ()
[0x00006] in <f9ca34d669a54607bed0e9bd0795a0d4>:0 at MvvmCross.Platforms.Android.Core.MvxAndroidSetup.get_Presenter ()
[0x00000] in <17df0d0bdae848b7a8a12b58d710f763>:0 at MvvmCross.Platforms.Android.Core.MvxAndroidSetup.RegisterPresenter ()
[0x00000] in <17df0d0bdae848b7a8a12b58d710f763>:0 at MvvmCross.Platforms.Android.Core.MvxAndroidSetup.InitializePlatformServices ()
[0x0000c] in <17df0d0bdae848b7a8a12b58d710f763>:0 at MvvmCross.Core.MvxSetup.InitializePrimary ()
[0x00052] in <17df0d0bdae848b7a8a12b58d710f763>:0 at MvvmCross.Core.MvxSetupSingleton.StartSetupInitialization ()
[0x0000a] in <17df0d0bdae848b7a8a12b58d710f763>:0 at MvvmCross.Core.MvxSetupSingleton.EnsureInitialized ()
[0x00017] in <17df0d0bdae848b7a8a12b58d710f763>:0 at MyCairns.Droid.Services.PushNotifications.CRCFirebaseMessagingService.OnCreate ()
[...]

PR #3032, opened 2018-07-25
Fix crash when switching back to the app after Permission change

vatsalyagoel:
when changing permission and going back to the application, ViewAssemblies is null thus causes the app to crash during setup 

Bugfix

When Android forms application is running and you change app permissions when going into Android settings, once you switch back to the app, the app crashes as ViewAssemblies is null. s

I saw that we have a get view assemblies method which we probably should use instead of referenceing ViewAsemblies directly

No

Launch App
Go to app Info without closing the app
Enable all permissions available
Go back to app in task switcher
Switch back to permissions
Disable a permission
Switch back to app
App Crashes with error
```
System.Linq.Enumerable.TryGetFirst[TSource] (System.Collections.Generic.IEnumerable`1[T] source, System.Boolean& found) [0x0000d] in <908c8d3ddc3a406497c942eef2cea13b>:0 
at System.Linq.Enumerable.FirstOrDefault[TSource] (System.Collections.Generic.IEnumerable`1[T] source) [0x00000] in <908c8d3ddc3a406497c942eef2cea13b>:0 
at MvvmCross.Forms.Platforms.Android.Core.MvxFormsAndroidSetup.get_FormsApplication () [0x00036] in <f9ca34d669a54607bed0e9bd0795a0d4>:0 
at MvvmCross.Forms.Platforms.Android.Core.MvxFormsAndroidSetup.CreateViewPresenter () [0x00006] in <f9ca34d669a54607bed0e9bd0795a0d4>:0 
at MvvmCross.Platforms.Android.Core.MvxAndroidSetup.get_Presenter () [0x00000] in <17df0d0bdae848b7a8a12b58d710f763>:0 
at MvvmCross.Platforms.Android.Core.MvxAndroidSetup.RegisterPresenter () [0x00000] in <17df0d0bdae848b7a8a12b58d710f763>:0 
at MvvmCross.Platforms.Android.Core.MvxAndroidSetup.InitializePlatformServices () [0x0000c] in <17df0d0bdae848b7a8a12b58d710f763>:0 at MvvmCross.Core.MvxSetup.InitializePrimary () [0x00052] in <17df0d0bdae848b7a8a12b58d710f763>:0 
at MvvmCross.Core.MvxSetupSingleton.StartSetupInitialization () [0x0000a] in <17df0d0bdae848b7a8a12b58d710f763>:0 at MvvmCross.Core.MvxSetupSingleton.EnsureInitialized () [0x00017] in <17df0d0bdae848b7a8a12b58d710f763>:0
 at MvvmCross.Forms.Platforms.Android.Views.MvxFormsAppCompatActivity.OnCreate (Android.OS.Bundle bundle) [0x0000c] in <f9ca34d669a54607bed0e9bd0795a0d4>:0
```

#2969 
Refer https://stackoverflow.com/questions/33488589/android-marshmallow-dynamic-permission-change-kills-all-application-processe
This also fixes the issue when launching the app from MvxIntentService

- [x] All projects build

### DIFF
--- a/MvvmCross.Forms/Platforms/Android/Core/MvxFormsAndroidSetup.cs
+++ b/MvvmCross.Forms/Platforms/Android/Core/MvxFormsAndroidSetup.cs
@@ -59,7 +59,7 @@ namespace MvvmCross.Forms.Platforms.Android.Core
                 {
                     var activity = Mvx.IoCProvider.Resolve<IMvxAndroidCurrentTopActivity>()?.Activity ?? ApplicationContext;
                     var asmb = activity.GetType().Assembly;
-                    Xamarin.Forms.Forms.Init(activity, null, ViewAssemblies.FirstOrDefault() ?? asmb);
+                    Xamarin.Forms.Forms.Init(activity, null, GetViewAssemblies()?.FirstOrDefault() ?? asmb);
                 }
                 if (_formsApplication == null)
                 {
--- a/MvvmCross.Forms/Platforms/Android/Core/MvxFormsAndroidSetup.cs
+++ b/MvvmCross.Forms/Platforms/Android/Core/MvxFormsAndroidSetup.cs
@@ -59,7 +59,7 @@ namespace MvvmCross.Forms.Platforms.Android.Core
                 {
                     var activity = Mvx.IoCProvider.Resolve<IMvxAndroidCurrentTopActivity>()?.Activity ?? ApplicationContext;
                     var asmb = activity.GetType().Assembly;
-                    Xamarin.Forms.Forms.Init(activity, null, GetViewAssemblies()?.FirstOrDefault() ?? asmb);
+                    Xamarin.Forms.Forms.Init(activity, null, ExecutableAssembly ?? asmb);
                 }
                 if (_formsApplication == null)
                 {
--- a/MvvmCross/Platforms/Android/Core/MvxAndroidSetup.cs
+++ b/MvvmCross/Platforms/Android/Core/MvxAndroidSetup.cs
@@ -38,7 +38,7 @@ namespace MvvmCross.Platforms.Android.Core
             _applicationContext = applicationContext;
         }
 
-        public virtual Assembly ExecutableAssembly => ViewAssemblies.FirstOrDefault() ?? GetType().Assembly;
+        public virtual Assembly ExecutableAssembly => ViewAssemblies?.FirstOrDefault() ?? GetType().Assembly;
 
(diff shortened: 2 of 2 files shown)


nickrandolph:
Please can we address the issue of why ViewAssemblies is null - this is there as an optimisation so that that GetViewAssemblies (which could be an expensive call) doesn't have to be called repeatedly.

nickrandolph:
I know this code was already there but the fallback to GetType().Assembly concerns me - it's less likely that this will resolve to the correct assembly now. It's no longer required to override Setup, which means that GetType().Assembly will resolve to the MVX assembly, rather than the application assembly

vatsalyagoel:
I do have another suggestion which may require a bit of architecture change. How about making the splash screen just a splash screen with no setup needed and moving the setup to AppCompatActivity for which the code already exists.

nickrandolph:
No architecture change required. If you want you can use a vanilla splash screen today with mvx and just load mvx when you reach the first activity. But that defeats the whole point - which is to have splash shown whilst you setup mvx

nickrandolph:
@vatsalyagoel do the repro steps in the description work with the Playground sample app? I'm concerned that ViewAssemblies is null as this shouldn't really happen

vatsalyagoel:
Hi @nickrandolph I tested the issue in playground and it does happen there as well

martijn00:
I tested this in one of my own apps, and can confirm that i both see the same issue, and that this fixes it. I'm not 100% sure this should be the longer term fix, but for now it works.

JelleDamen:
I've seen this in a project of mine, but in a different scenario (direct on startup)
I've worked around it by implementing the 'ExecutableAssembly' property as follows:

`public override Assembly ExecutableAssembly => ViewAssemblies?.FirstOrDefault() ?? GetType().Assembly;`

instead of

`public override Assembly ExecutableAssembly => ViewAssemblies.FirstOrDefault() ?? GetType().Assembly;`

because during setup the ViewAssemblies are not loaded yet.

vatsalyagoel:
@JelleDamen That's what we ended up doing in this PR, However I agree with @martijn00 and @nickrandolph This is not a good long term fix as we don't exactly know what gettype().Assembly will give us. It may give us the Mvx or even Mono Droid assembly.

JelleDamen:
@vatsalyagoel Ah I see. Any clue why the ViewAssemblies are not loaded, I've not looking into it yet

martijn00:
Let's merge this, and look further into the problem for 6.3.

PR #3044, opened 2018-07-31
Adding MvxNavigationViewModel to remove injected NavigationService and LogProvider

nickrandolph:
Feature

LogProvider and NavigationService are injected into MvxViewModel

MvxViewModel no longer has NavigationService or LogProvider
New MvxNavigationViewModel has NavigationService or LogProvider and requires them as part of the constructor (declare dependencies)

Yes - MvxViewModel no longer has NavigationService or LogProvider

Run Playground samples

- [ ] All projects build


nickrandolph:
@Cheesebaron I just realised there was an issue with file naming - this has been fixed if you want to take another look.

MvvmCross/ViewModels/MvxNavigationViewModel.cs
```diff
@@ -0,0 +1,65 @@
...
+        private IMvxLogProvider _logProvider;
+        private IMvxLog _log;
+
+        protected MvxNavigationViewModel(IMvxLogProvider logProvider, IMvxNavigationService navigationService) : base()
+        {
+            LogProvider = logProvider;
+            NavigationService = navigationService;
+        }
+
+        public virtual IMvxNavigationService NavigationService { get; }
```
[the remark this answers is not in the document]
nickrandolph: done

SamuelDebruyn:
How is this a feature? This is another breaking change for a breaking change introduced in 6.1. You can't do much with a ViewModel without navigation, so why not leave them be?

nickrandolph:
@SamuelDebruyn we can reclassify it as a bug fix if you'd prefer - essentially the change we made previously was done for convenience for developers, so they didn't need to add NavigationService and LogProvider to every view model. However, by adding them as properties, they become hidden dependencies and this broke the unit tests too. 
There was a thread on Slack discussing this and a few ideas were proposed. Introducing a second VM with these properties but having them set via the constructor as a bit of a middle ground.
There are some scenarios where VMs are used that don't require navigation, which is why we didn't change the constructor (and properties) on MvxViewModel

programmation:
@SamuelDebruyn I have always split my ViewModels into "navigating" (e.g. modelling a whole page) and "non-navigating" (e.g. modelling list items or subviews in a page). Non-navigating ViewModels still need to be ViewModels in that they require `INotifyPropertyChanged` and have `ICommands`,  but they will often inform their container about actions/commands being invoked rather than respond to the command directly themselves.

@nickrandolph Is there any way Mvx could provide the `INotifyPropertyChanged` and so on as "mix-in" rather than requiring my classes to inherit the Mvx base classes? When you add in the generic support there are actually 3 Mvx base classes that have essentially the same code and the same properties but which have to be inherited individually: `MvxViewModel`, `MvxViewModel<T>` and `MvxViewModel<T, R>`. Now we might need "navigating" and "non-navigating" versions as well, so 6 all up? It's not the end of the world, but I was wondering if there's a better way - or are we at a language/paradigm limitation?

programmation:
@nickrandolph I had a hare-brained idea that it might be possible to use interfaces and extension methods to enable users to get Mvx functionality without having to inherit from Mvx base classes.

To illustrate, I made a short example here: https://gist.github.com/programmation/acd140048f7a3641d5609005868173b2.

I don't want to hijack this thread (any further) - perhaps we could continue on Slack?

martijn00:
I would love to use default interfaces for this. One problem is that this is first coming on C#8 which is not released yet.

MvvmCross/ViewModels/MvxNavigationViewModel.cs
```diff
@@ -0,0 +1,65 @@
...
+// The .NET Foundation licenses this file to you under the MS-PL license.
+// See the LICENSE file in the project root for more information.
+
+using System.Threading.Tasks;
+using MvvmCross.Logging;
+using MvvmCross.Navigation;
+
+namespace MvvmCross.ViewModels
+{
+    public abstract class MvxNavigationViewModel
```
martijn00: Shouldn't we name it differently? `MvxNavigationViewModel` sounds like it is only for Navigation. What about `MvxBaseViewModel` or something?

MvvmCross/ViewModels/MvxNavigationViewModel.cs
```diff
@@ -0,0 +1,65 @@
...
+        : MvxViewModel
+    {
+        private IMvxLogProvider _logProvider;
```
martijn00: I think we should also provide IoC in here.

MvvmCross/ViewModels/MvxNavigationViewModel.cs
```diff
@@ -0,0 +1,65 @@
...
+        private IMvxLog _log;
+
+        protected MvxNavigationViewModel(IMvxLogProvider logProvider, IMvxNavigationService navigationService) : base()
```
martijn00: Should be public?

MvvmCross/ViewModels/MvxNavigationViewModel.cs
nmilcoff: `IoCProvider`? Mhmm I think we want to encourage users to use dependency injection in most cases 🤔

MvvmCross/ViewModels/MvxNavigationViewModel.cs
nickrandolph: MvxBaseViewModel sounds like it should be the base class of MvxViewModel, not the other way around. I think there are typically two types of viewmodels that developers create - those that are used as part of navigation (ie would inherit from MvxNavigationViewModel) and those that are used to populate lists (or other non-navigation related scenarios, and would inherit from MvxViewModel). This is the reason I set the name as MvxNavigationViewModel 

Having said that, happy to consider other names. The other alternative is that we switch the names - this would be MvxViewModel, and the class that is currently MvxViewModel would become MvxBaseViewModel. This would also limit the breaking changes - developers would still need to add the parameters to their constructor but they wouldn't need to change their base class.

programmation:
A bit late in the day but I wonder if calling the ViewModel that does navigation `MvxNavigatingViewModel` might help?

MvvmCross/ViewModels/MvxNavigationViewModel.cs
martijn00: How would you be able to create a new instance in a unit test?

Cheesebaron: You can mock it

nickrandolph: Done - I've added a basic unit test to validate that protected properties are indeed set during construction.

MvvmCross/ViewModels/MvxNavigationViewModel.cs
martijn00: Alright, but let's make `IMvxNavigationService` the first ctor parameter then, so that it is easier to refactor later if we ever decide to remove `IMvxLogProvider` from it.

PR #3046, opened 2018-08-01
Adding cancel support through navigation

nickrandolph:
Feature

Cancellation of navigation is only supported via Cancellation token (ie can't cancel within navigation lifecycle)

Navigation can still be cancelled via cancellation token (ie by code that triggers navigation)
Navigation can also be cancelled throughout the navigation lifecycle

Yes - added support for MvxNavigationEventArgs

- [ ] All projects build

### DIFF
--- a/MvvmCross/Navigation/EventArguments/NavigateEventArgs.cs
+++ b/MvvmCross/Navigation/EventArguments/NavigateEventArgs.cs
@@ -22,7 +22,7 @@ namespace MvvmCross.Navigation.EventArguments
             Mode = mode;
         }
 
-        public MvxNavigateEventArgs(NavigationMode mode, IMvxViewModel viewModel, CancellationToken cancellationToken = default) : this(mode, cancellationToken)
+        public MvxNavigateEventArgs(IMvxViewModel viewModel, NavigationMode mode, CancellationToken cancellationToken = default) : this(mode, cancellationToken)
         {
             ViewModel = viewModel;
         }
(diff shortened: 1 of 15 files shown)


MvvmCross/Navigation/EventArguments/NavigateEventArgs.cs
```diff
@@ -2,25 +2,32 @@
...
+
+    public class MvxNavigateEventArgs : MvxCancelEventArgs, IMvxNavigateEventArgs
+    {
+        public MvxNavigateEventArgs(NavigationMode mode, CancellationToken cancellationToken = default) : base(cancellationToken)
         {
+            Mode = mode;
         }
 
-        public NavigateEventArgs(IMvxViewModel viewModel, CancellationToken cancellationToken = default(CancellationToken))
+        public MvxNavigateEventArgs(NavigationMode mode, IMvxViewModel viewModel, CancellationToken cancellationToken = default) : this(mode, cancellationToken)
```
martijn00: I would put the `NavigationMode` after viewmodel, to keep it more like it already was, and to better match the other ctor.

nickrandolph:
@martijn00 I've fixed up the ordering of the parameters and done a rebase. I need to fix the build errors when I get a chance

markuspalme:
Hi @nickrandolph,

I want to update the navigation documentation as part of https://github.com/MvvmCross/MvvmCross/issues/3106. Is my understanding correct that...

- the newly introduced model allows cancelletion within the navigation lifecycle for the outside world (for example by subscribing to an event handler such as `BeforeNavigate`)
- previously, only the caller of a method like `mvxNavigationService.Navigate` could cancel the navigation using the cancellation token.

Any other important concepts that should be covered?

Thanks in advance!

Issue #2903, opened 2018-05-23
Exceptions are swallowed during Android setup


tbalcom:
### Steps to reproduce :scroll:

1. Associate more than one view with a view model.
2. Start the app.
3. The Android splash screen never closes. There are no exceptions thrown or messages logged. Previous versions of MvvmCross would trigger an exception making it easier to debug these kind of setup problems.

### Expected behavior :thinking:
Exceptions thrown during setup should not be swallowed on Android.

### Actual behavior :bug:
Exceptions during setup are swallowed and nothing is logged.

### Configuration :wrench:

**Version:** 6.0.1

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

sergeyyurkov:
I believe it also applies at least to WPF. I didn't add one of the nuget packages to WPF project so I got a System.Reflection.ReflectionTypeLoadException somewhere in mscorelib that appeared only if I unselect 'Just my code' in the debugger options. With the option enabled (which is a default) the WPF app end up running without any window and absolutely error-clean log.

nickrandolph:
If you enable all exceptions in the exceptions window in visual studio do any get thrown? If so, can you provide a PR that shows how to reproduce this using the playground sample please

sergeyyurkov:
@nickrandolph Unfortunately I can't. Cloned the repo and got a bunch of errors trying to build Playground.WPF. Started with `Error occurred while restoring NuGet packages: The operation failed as details for project MvvmCross.Forms could not be loaded` and continued with dozens of `...\File\obj\project.assets.json' not found. Run a NuGet package restore to generate this file.`
Did I miss something obvious?

nickrandolph:
Can you try the 6.1 branch please - we've had to fix a bunch of build related issues, so please try this branch instead. 
If it doesn't build properly, please do a solution clean, restart VS and try building again (reach me on mvvmcross.slack.com if you need assistance)

tbalcom:
I was able to see a `System.ArgumentException` when I enabled all exceptions.

> Exception:
> 
> System.ArgumentException: An item with the same key has already been added. Key: Playground.Core.ViewModels.RootViewModel occurred
>

[...]

PR #3051, opened 2018-08-03
Propagating exceptions out of setup

nickrandolph:
Bug fix

If an exception is raised during the second half of initialization of setup class (ie initializeSecondard) it is swallowed and the application never loads the first view of the application

If an exception is raised during the second half of initialization, initialization completes and the exception is propagated to the host application

No (although it does change the IMvxSetupMonitor interface

1. Associate more than one view with a view model.
2. Start the app.
3. Previously:The Android splash screen never closes, leaving a white screen displayed.
Now: An exception is bubbled up allowing the error to be detected

https://github.com/MvvmCross/MvvmCross/issues/2903

- [ ] All projects build

### DIFF
--- a/MvvmCross/Core/MvxSetupSingleton.cs
+++ b/MvvmCross/Core/MvxSetupSingleton.cs
@@ -7,6 +7,7 @@ using System.Threading.Tasks;
 using MvvmCross.Base;
 using MvvmCross.Exceptions;
 using MvvmCross.Logging;
+using MvvmCross.ViewModels;
 
 namespace MvvmCross.Core
 {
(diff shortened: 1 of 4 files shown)


nickrandolph:
The change to InitializationComplete was actually a carry over from my first attempt. I left it in because it provides a point where the exception could be inspected, before it crashes the app (this can make debugging particularly on Android easier).
It's actually not essential as the core change actually allows the Exception to propagate all the way up, eventually crashing the app. This is a good thing and intentional behaviours as it means that a developer can diagnose what's causing the exception, rather than it being swallowed by mvx.
To see this in action, try overriding one of the methods in Setup and throw an exception. You should see it crash the app.

martijn00:
@nickrandolph i would agree on not changing InitializationComplete for now, but keep the other change. That will be sufficient for now i think.

nickrandolph:
Done

MvvmCross/Core/MvxSetupSingleton.cs
```diff
@@ -169,17 +174,32 @@ protected override void Dispose(bool isDisposing)
...
             Task.Run(async () =>
             {
-                _setup.InitializeSecondary();
+                try
+                {
+                    _setup.InitializeSecondary();
+                }
+                catch(Exception ex)
+                {
+                    SetupException = ex;
```
Cheesebaron: If this exception would be rethrown at a later point, it could be worth using `ExceptionDispatchInfo.Capture(ex)` to give a more meaningful stack trace.

Issue #3029, opened 2018-07-24
NavigationService Close doesn't respect Viewmodel Parameter in forms and always pop latest page


abdomh:
## 🐛 Bug Report

Calling close in navigation service with ViewModel instance always pops the latest page

I did some digging and found out that every call to `ClosePage` in `MvxFormsPagePresenter` sets the expected page to close to null instead of finding the page, which triggers `PopAsync`
[https://github.com/MvvmCross/MvvmCross/blob/6.1.2/MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs#L647](https://github.com/MvvmCross/MvvmCross/blob/6.1.2/MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs#L647)

I could be wrong but this is what I think where the issue is.

### Expected behavior
Close the passed ViewModel instance only or return false if not possible

### Reproduction steps
Add two view models
set ViewModelA to be the default page
Navigate to ViewModelB
Close ViewModelA (will close ViewModelB insted)

### Configuration

**Version:** 6.x
6.1.2

**Platform:** 
- [ ] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [x] :monkey: Xamarin.Forms

nickrandolph:
@AbdelrahmanGIT are you able to propose a fix as a PR for review?

AbdelrahmanGIT:
I will take a look at it

PR #3063, opened 2018-08-10
3029 fix close ViewModel behavior

AbdelrahmanGIT:
Bug fix

`IMvxNavigationService Close` Ignores `viewModel` parameter and will always pop the latest page

`IMvxNavigationService Close` Closes the view for `viewModel` instance where possible

Yes, if the user depends on the old behavior where it pops the latest page.
to pop the latest page instead you can use `MvxPopRecursivePresentationHint` with 1 level

[https://github.com/MvvmCross/MvvmCross/issues/3029](https://github.com/MvvmCross/MvvmCross/issues/3029)

- [x] All projects build


MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs
```diff
@@ -463,7 +473,15 @@ public override MvxBasePresentationAttribute CreatePresentationAttribute(Type vi
 
         public virtual Task<bool> CloseContentPage(IMvxViewModel viewModel, MvxContentPagePresentationAttribute attribute)
         {
-            return ClosePage(FormsApplication.MainPage, null, attribute);
+            var view = FindViewFromViewModel(viewModel);
```
nickrandolph: Please refactor this block of code so it's not repeated three times.

AbdelrahmanGIT: I updated the code but it shows a failing test in `MvvmCross.UnitTest.Navigation.NavigationServiceTests.Test_NavigateCallbacks` but in my machine, it passes the test not sure why

AbdelrahmanGIT:
I updated the code but it shows a failing test in MvvmCross.UnitTest.Navigation.NavigationServiceTests.Test_NavigateCallbacks but in my machine, it passes the test not sure why

MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs
```diff
@@ -530,7 +532,7 @@ public virtual Task<bool> CloseMasterDetailPage(IMvxViewModel viewModel, MvxMast
                 case MasterDetailPosition.Master:
                     return ClosePage(masterDetailHost.Master, null, attribute);
```
nickrandolph: Shouldn't we do the same for the master? In theory it could have navigation too

MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs
```diff
@@ -958,5 +962,27 @@ public virtual void ReplacePageRoot(Page existingPage, Page page, MvxPagePresent
                 throw new MvxException(ex, "Cannot replace MainPage root");
             }
         }
+
+        protected virtual Task<bool> FindAndCloseViewFromViewModel(IMvxViewModel mvxViewModel, Page rootPage, MvxPagePresentationAttribute attribute)
+        {
+            var root = TopNavigationPage();
+            Page pageToClose = null;
+
+            if (root?.Navigation?.NavigationStack != null)
```
nickrandolph: Do we need this null check - can't we propagate the null through to the next line?

Projects/Playground/Playground.Forms.UI/Playground.Forms.UI.csproj
```diff
@@ -19,6 +19,9 @@
     <EmbeddedResource Update="Pages\ConvertersPage.xaml">
       <Generator>MSBuild:UpdateDesignTimeXaml</Generator>
     </EmbeddedResource>
+    <EmbeddedResource Update="Pages\NavigationClosePage.xaml">
```
nickrandolph: Do the xaml pages need to be defined in the csproj? I thought the default behaviour is that they're included without need for them to be explicitly listed here.

MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs
AbdelrahmanGIT: I'm not sure since the current behavior for `MasterDetailPosition.Root` and `MasterDetailPosition.Master` is to send a root page which is then ignored by `ClosePage` method and will try to `PopAsync` from the navigation stack but `master detail` as far as I know it doesn't push itself into the navigation stack, knowing that I don't think we can do `popAsync or removePage` from the navigation stack?

AbdelrahmanGIT: Might it require a different implementation?

martijn00:
@AbdelrahmanGIT Can you do a rebase on this?

MvvmCross.Forms/Presenters/MvxFormsPagePresenter.cs
```diff
@@ -958,5 +962,24 @@ public virtual void ReplacePageRoot(Page existingPage, Page page, MvxPagePresent
...
+
+        protected virtual Task<bool> FindAndCloseViewFromViewModel(IMvxViewModel mvxViewModel, Page rootPage, MvxPagePresentationAttribute attribute)
+        {
+            var root = TopNavigationPage();
+            Page pageToClose = null;
+
+            // finding the view from viewmodel in navigation stack
+            pageToClose = root?.Navigation?.NavigationStack?
+                .OfType<IMvxPage>()
+                .FirstOrDefault(x => x.ViewModel == mvxViewModel) as Page;
```
nmilcoff: Wouldn't it make more sense to use`LastOrDefault` here?

AbdelrahmanGIT: `LastOrDefault` will try to find the last occurrence of an item and in most cases will start from the beginning of a list which means it will always do `O(n)` since it needs to know if there is another items that matches the condition. while `FirstOrDefault` will stop once the condition is satisfied.

PR #3073, opened 2018-08-18
Add ApplyWithClearBindingKey to MvxFluentBindingDescriptionSet

Plac3hold3r:

Feature

In order to add clear binding key currently, you have to add one to each individual binding description.

A developer can now make use of `ApplyWithClearBindingKey` on a `MvxFluentBindingDescriptionSet` which will apply the key to the entire set.

No

See "Show Fluent Binding Example" option in the Android playground.

n/a

- [X] All projects build

### DIFF
--- a/MvvmCross/Binding/BindingContext/MvxBaseFluentBindingDescription.cs
+++ b/MvvmCross/Binding/BindingContext/MvxBaseFluentBindingDescription.cs
@@ -17,7 +17,7 @@ using MvvmCross.Exceptions;
 namespace MvvmCross.Binding.BindingContext
 {
     public class MvxBaseFluentBindingDescription<TTarget>
-        : MvxApplicableTo<TTarget>
+        : MvxApplicableTo<TTarget>, IMvxBaseFluentBindingDescription
         where TTarget : class
     {
         private readonly TTarget _target;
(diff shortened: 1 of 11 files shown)


MvvmCross/Binding/BindingContext/MvxBaseFluentBindingDescription.cs
```diff
@@ -163,6 +163,12 @@ public static class SourceSpecHelpers
 
         protected object ClearBindingKey { get; set; }
 
+        object IMvxBaseFluentBindingDescription.ClearBindingKey
```
nmilcoff: This should be public

Plac3hold3r: @nmilcoff  I believe the default access modifier of an explicit interface implementation is public by virtue of the interface contract require it to be public. As it can only be public C# does not require an access modifier to be specified ([Explicit Interface Implementation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation)).

nmilcoff: Oh, right! I didn't see it was implemented explicitly 🙂 👍

Plac3hold3r: No worries 🙂

Issue #3083, opened 2018-08-31
Playground.Droid can't navigate to RootViewModel


tbalcom:
## 🐛 Bug Report
`MvxDefaultViewModelLocator.RunViewModelLifecycle(IMvxViewModel viewModel, IMvxBundle parameterValues, IMvxBundle savedState, IMvxNavigateEventArgs navigationArgs)` is throwing a NullReferenceException because `navigationArgs` is null.

> MvvmCross.Exceptions.MvxException: Problem navigating to ViewModel RootViewModel ---> MvvmCross.Exceptions.MvxException: Failed to construct and initialize ViewModel for type Playground.Core.ViewModels.RootViewModel from locator MvxDefaultViewModelLocator - check InnerException for more information ---> MvvmCross.Exceptions.MvxException: Problem running viewModel lifecycle of type RootViewModel ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at MvvmCross.ViewModels.MvxDefaultViewModelLocator.RunViewModelLifecycle (MvvmCross.ViewModels.IMvxViewModel viewModel, MvvmCross.ViewModels.IMvxBundle parameterValues, MvvmCross.ViewModels.IMvxBundle savedState, MvvmCross.Navigation.EventArguments.IMvxNavigateEventArgs navigationArgs) [0x0001d] in E:\src\MvvmCross\MvvmCross\ViewModels\MvxDefaultViewModelLocator.cs:103 
    --- End of inner exception stack trace ---
   at MvvmCross.ViewModels.MvxDefaultViewModelLocator.RunViewModelLifecycle (MvvmCross.ViewModels.IMvxViewModel viewModel, MvvmCross.ViewModels.IMvxBundle parameterValues, MvvmCross.ViewModels.IMvxBundle savedState, MvvmCross.Navigation.EventArguments.IMvxNavigateEventArgs navigationArgs) [0x000af] in E:\src\MvvmCross\MvvmCross\ViewModels\MvxDefaultViewModelLocator.cs:123 
[...]

PR #3084, opened 2018-08-31
Fix bug preventing Playground.Droid from starting

tbalcom:
Bug fix

Playground.Droid throws a NullReferenceException while attempting to navigate to RootViewModel.

Playground.Droid can successfully navigate to RootViewModel.

No

Build & run Playground.Droid

#3083

- [x] All projects build

### DIFF
--- a/MvvmCross/ViewModels/MvxDefaultViewModelLocator.cs
+++ b/MvvmCross/ViewModels/MvxDefaultViewModelLocator.cs
@@ -100,20 +100,20 @@ namespace MvvmCross.ViewModels
             try
             {
                 CallCustomInitMethods(viewModel, parameterValues);
-                if (navigationArgs.Cancel)
+                if (navigationArgs?.Cancel == true)
                     return;
                 if (savedState != null)
                 {
                     CallReloadStateMethods(viewModel, savedState);
-                    if (navigationArgs.Cancel)
+                    if (navigationArgs?.Cancel == true)
                         return;
                 }
                 viewModel.Start();
-                if (navigationArgs.Cancel)
+                if (navigationArgs?.Cancel == true)
                     return;
 
                 viewModel.Prepare();
-                if (navigationArgs.Cancel)
+                if (navigationArgs?.Cancel == true)
                     return;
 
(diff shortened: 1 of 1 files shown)


MvvmCross/ViewModels/MvxDefaultViewModelLocator.cs
```diff
@@ -100,20 +100,20 @@ protected void RunViewModelLifecycle(IMvxViewModel viewModel, IMvxBundle paramet
             try
             {
                 CallCustomInitMethods(viewModel, parameterValues);
-                if (navigationArgs.Cancel)
+                if (navigationArgs?.Cancel == true)
```
Cheesebaron: I don't think the `navigationArgs` will all of the sudden be null after this point. So it should be enough just to check for null one place.

tbalcom: That doesn't appear to be the case, at least on Android. Stepping through Playground.Droid in the debugger I see the `RunViewModelLifecycle(IMvxViewModel viewModel, IMvxBundle parameterValues, IMvxBundle savedState, IMvxNavigateEventArgs navigationArgs)` is called first and `navigationArgs` is null. Then `RunViewModelLifecycle<TParameter>(IMvxViewModel<TParameter> viewModel, TParameter param, IMvxBundle parameterValues, IMvxBundle savedState, IMvxNavigateEventArgs navigationArgs)` is called later and `navigationArgs` is null there as well. `navigationArgs` is currently always null during both of these methods. Removing any of these null checks I've added in this PR just moves where the NullReferenceException originates.

There seems to be a bug in PR #3046 where `navigationArgs` are not being supplied in all scenarios so the default value of null is sometimes used. There were no null checks against `navigationArgs` even though it defaults to null. This PR addresses the lack of null checks and allows Playground to run again. There is still an underlying problem introduced by #3046 if `navigationArgs` isn't expected to be null.

Cheesebaron: What I mean is not that the argument cannot be null. But rather inside of `RunViewModelLifecycle` there is nothing that modifies `navigationArgs` and it should be fine checking it only once.

PR #3095, opened 2018-09-12
Fix Playground.Forms showing a black screen after splash

Cheesebaron:
AppStart Start isn't called.

Bug fix

Black screen on Android after Splash Screen

App actually runs

Nope

Launch Playground.Forms.Droid

- [x] All projects build

### DIFF
--- a/Projects/Playground/Playground.Forms.Droid/SplashScreen.cs
+++ b/Projects/Playground/Playground.Forms.Droid/SplashScreen.cs
@@ -29,7 +29,7 @@ namespace Playground.Forms.Droid
         protected override Task RunAppStartAsync(Bundle bundle)
         {
             StartActivity(typeof(MainActivity));
-            return Task.CompletedTask;
+            return base.RunAppStartAsync(bundle);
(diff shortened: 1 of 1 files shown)


Cheesebaron:
Double navigation? I dunno, but it is a whole lot better than the App doesn't work at all.

nickrandolph:
Just concerned because this has been working, so don't know what would have changed to have broken this. I had in the past noticed a blank screen between splash and the first page but I think that was just down to my emulator running super slow

nmilcoff:
I just tested this on latest develop and it's working fine on my end 🤔

Cheesebaron:
@nmilcoff just to confirm, you are deploying and launching the Playground.Forms.Droid project, right?

For me on my device it just shows a black screen after the splash has been shown and stays there forever, unless I call the base method, which kicks in AppStart.

Cheesebaron:
OK. So I think VS is messing with me. If I deploy using Rider instead it works as expected. I have no idea what the heck is going on.

PR #3101, opened 2018-09-13
Use EnteredBackground to trigger suspension logic

### LINKED ISSUE #3090: Proper implementation of UWP suspension
## 🐛 Bug Report

MvvmCross currently does not handle UWP suspension app lifecycle properly.

When app goes into suspension,[ `MvxSuspensionManager` is used in `MvxApplication`](https://github.com/MvvmCross/MvvmCross/blob/bdaa09299714d94cf3f2c548a465d994c20d52f0/MvvmCross/Platforms/Uap/Views/MvxApplication.cs#L121) to store the current state of the app, including navigation:

```
private async void OnSuspending(object sender, SuspendingEventArgs e)
{
    var deferral = e.SuspendingOperation.GetDeferral();

    var suspension = Mvx.IoCProvider.GetSingleton<IMvxSuspensionManager>();
    await Suspend(suspension);
    await suspension.SaveAsync();
    deferral.Complete();
}
```

This behavior was the recommended one before Windows 10 version 1607, but currently the recommendation is to put suspension logic in `EnteredBackground` event handler, [as per docs](https://docs.microsoft.com/en-us/windows/uwp/launch-resume/app-lifecycle#app-suspend
):

> When an app is suspended, it invokes the Application.Suspending event. Visual Studio’s UWP project templates provide a handler for this event called OnSuspending in App.xaml.cs. Prior to Windows 10, version 1607, you would put the code to save your state here. Now the recommendation is to save your state when you enter the background state, as described above.

App run suspension in one of two ways - either user closes the app, or the app is either minimized or lock screen is on. In case of app close the app goes through suspension and then is immediately removed from memory by the system. In case of minimize, the documentation says the following:

> When the user minimizes an app Windows waits a few seconds to see whether the user will switch back to it. If they do not switch back within this time window, and no extended execution, background task, or activity sponsored execution is active, Windows suspends the app. An app is also suspended when the lock screen appears as long as no extended execution session, etc. is active in that app.

However, it is key to note, that suspension **does not mean** the app will necessarily be removed from memory. If the user switches back to a minimized app, it will continue running as it was and does not require any restore except for a possible refresh of the UI if the developer so chooses.

Here MvvmCross uses the wrong path - the suspended state is restored **only** when the [`Resuming` event is called in `MvxApplication`](https://github.com/MvvmCross/MvvmCross/blob/bdaa09299714d94cf3f2c548a465d994c20d52f0/MvvmCross/Platforms/Uap/Views/MvxApplication.cs#L136):

```
private async void OnResuming(object sender, object e)
{
    var suspension = Mvx.IoCProvider.GetSingleton<IMvxSuspensionManager>();
    await Resume(suspension);
    await suspension.RestoreAsync();
}
```

This means when the user switches back to the minimized app, it first restores the state. But that is not necessary at all, because the full app state is still in memory and there is no additional work required. Basically, suspension can be seen as a precautionary measure in case the system runs out of resources and needs to kill the app to free some up.

When the system kills a suspended app, when the user launches it or switches back to it, it will now go through the full launch lifetime, which you can [confirm here in docs:](https://docs.microsoft.com/en-us/windows/uwp/launch-resume/app-lifecycle#app-launch)

ApplicationExecutionState | Explanation | Action to take
-- | -- | --
Terminated | The app was previously suspended but was then shutdown at some point because the system needed to reclaim memory. | Restore the state that the app was in when the user switched away from it.
ClosedByUser | The user closed the app with the close gesture in tablet mode, or with Alt+F4. When the user closes the app, it is first suspended and then terminated. | Because the app has essentially gone through the same steps that lead to the Terminated state, handle this the same way you would the Terminated state.

In both **Terminated** and **ClosedByUser** execution states it is appropriate to restore the app state from suspended state because user expects to continue where he left off last time.

To fix the current behavior, I suggest the following:

1. **Move the suspension logic to the `EnteredBackground`** event handler to match the recommended approach
2. **Remove the `Resuming` handler** completely as it should be used only in case the developer needs to do some refresh of data or reacquire file handles
3. _Optionally_ implement restore in `OnLaunched` and `OnActivated`. This is optional because the better solution may just be to leave this up to the developer to decide if she wants to restore the state saved in suspension or not. This would give better flexibility to the developers using MvvmCross and would also preserve the current behavior which is that state is not restored.

### Expected behavior

When `EnteredBackground` event is called, `MvxSuspensionManager` is used to save current app state. When user switches to minimized app, no state is restored as the app is already fully in memory. If the app is terminated by the system in the meantime or if the user closes the app, app is restored from suspension manager if the app developer so chooses in `OnLaunched` and `OnActivated`.

For another example refer to [this UWP sample](https://github.com/Microsoft/Windows-universal-samples/blob/master/Samples/BasicSuspension/cs/App.xaml.cs), which deals with suspension. Note that this is an older sample and does not use `EnteredBackground` yet.

### Reproduction steps

Create a basic MvvmCross UWP app and test different execution states in Visual Studio.

### Configuration

**Version:** 6.2, affects all versions

**Platform:** 
- [ ] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [x] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

mmangold7:
…Resuming handler to fix suspension/resume crash bug

Bug fix

app crashes after resuming on UWP (such as from lock screen)

app now longer crashes after resuming

no

run a UWP project based on MvvmCross. lock the device. log in to the device. app should no longer crash upon resume/login.

Fixes #3090

- [x] All projects build


Cheesebaron:
Can you please rebase on develop or merge develop into your branch.

mmangold7:
@Cheesebaron I think I properly merged from develop. I checked out develop in my repo and merged from upstream/develop (upstream is this repo). Then I checked out the bugfix in my repo and merged from develop. I pushed the resulting commits.

MvvmCross/Platforms/Uap/Views/MvxApplication.cs
```diff
@@ -118,28 +117,18 @@ protected virtual void OnNavigationFailed(object sender, NavigationFailedEventAr
             throw new Exception("Failed to load Page " + e.SourcePageType.FullName);
         }
 
-        private async void OnSuspending(object sender, SuspendingEventArgs e)
+        private async void OnEnteredBackground(object sender, object e)
```
nickrandolph: This isn't sufficient - the point of deferrals in the OnSuspending is that if there is substantial amount of work to be done the app gets a chance to finish it before the app is suspended. I agree that we should move the Suspend logic to entering background but we need to make sure that if the app gets suspended whilst the Suspend logic is being run, we capture the deferral and only release it once the Suspend logic is complete.
Also, we can't just ignore the Resume logic - what happens if the app does in fact get suspended. As we just going to ignore resuming the application?

mmangold7: @nickrandolph  Thanks for the feedback. I didn't realize you could get the deferral off the EnteredBackgroundEventArgs. Silly of me to leave it out.

As for the Resuming handler, do we actually have to resume the suspension? Per what @MartinZikmund wrote in the original issue:

>This means when the user switches back to the minimized app, it first restores the state. But that >is not necessary at all, because the full app state is still in memory and there is no additional >work required. Basically, suspension can be seen as a precautionary measure in case the system >runs out of resources and needs to kill the app to free some up.
>
>When the system kills a suspended app, when the user launches it or switches back to it, it will >now go through the full launch lifetime

Calling "Resume(suspension)" when the app was suspended but not terminated is what causes the original issue. As Martin mentioned, in the case of termination, we could call resume from OnLaunch/OnActivated. His opinion on this was:

>the better solution may just be to leave this up to the developer to decide if she wants to restore >the state saved in suspension or not. This would give better flexibility to the developers using >MvvmCross and would also preserve the current behavior which is that state is not restored.

At the moment, I'm not sure which I prefer. But what I have changed will allow the app to be suspended and resumed as long as it is not terminated by windows. If it is, it will not resume from the suspension point, which is currently does not since it is broken. Developers could handle the relevant events (Launch/Activated) themselves if they want to resume from suspension after termination.

nickrandolph:
My view is that we should handle the resume from background, even if we don't do anything (since app is already in memory). I would suggest we handle the resume, grab the deferral (....actually there might not be one for resume) and call an overridable protected method to "handle" the resume. The default handling would be to do nothing but at least developers know that if they do want to handle it, they just need to override the Mvx method

MartinZikmund:
@nickrandolph I think that is fine. Resume does not have deferral as far as I remember. The only thing is that we should not restore from suspension manager inside the resume, but otherwise it is a good thing to achieve consistency of providing all lifetime methods.

mmangold7:
@nickrandolph @MartinZikmund what you guys have said makes sense to me. I think my latest commit complies with your reasoning. The handler is back but it does not restore by default.

nickrandolph:
Great but thinking about it, should we use the opposite method from OnEnteringBackground instead of Resuming?

MartinZikmund:
Good question... Docs talk about the suspension only, so it is hard to say. In any case, if the user wants he can add additional handler for `Resuming` or `LeavingBackground` manually in `MvxApplication`, so maybe we can just leave it up to the developer if they want to use any of them and not provide a empty implementation after all.

MvvmCross/Platforms/Uap/Views/MvxApplication.cs
```diff
@@ -118,9 +118,9 @@ protected virtual void OnNavigationFailed(object sender, NavigationFailedEventAr
...
         }
 
-        private async void OnSuspending(object sender, SuspendingEventArgs e)
+        private async void OnEnteredBackground(object sender, EnteredBackgroundEventArgs e)
         {
-            var deferral = e.SuspendingOperation.GetDeferral();
+            var deferral = e.GetDeferral();
 
             var suspension = Mvx.IoCProvider.GetSingleton<IMvxSuspensionManager>();
             await Suspend(suspension);
```
nickrandolph: Before we accept this, we should tidy up the methods a bit. We're no longer calling "Suspend" from within the OnSuspending - this could be a breaking change for some developers.
I suggest we keep the OnSuspending handler so that we can grab the deferral and call Suspend method.
Have an additional OnEnteredBackground where we grab the deferral invoke an EnteredBackground method.
I also don't think we should be calling SaveAsync in the OnEnteredBackground given that we don't call RestoreAsync - this feels like we're breaking this pattern

martijn00:
@mmangold7 can you make the last requested changes so we can merge this?

Cheesebaron:
Any news here?

Cheesebaron:
Can someone in @MvvmCross/core please grab this and address the review comments?

nickrandolph:
@mmangold7 / @Cheesebaron / @martijn00 take a look at how I've refactored the suspend/resume methods and let me know what you think. I've aimed to make it easy for developers to hook into enter/leave as well as suspend/resume events. The only anomaly is that we save session on entering background, as well as suspending but we do not load session information by default. It's easy enough for developers to now add their own logic but as was pointed out in the comments, in most cases the application stays memory resident so there's no work required to resume the state of the app.

PR #3128, opened 2018-09-26
Add ability to use tags from attributes

Cheesebaron:
Bugfix/feature

When customizing the AppCompat presenters on Android, it is unclear how tags are set on ViewPager fragments, since there is a default fallback to title. So this PR removes the overload constructors which do the fallback, so the presenter does this more explicitly now.

This PR also adds the ability to set a tag in the presentation attributes on Android, such that it eases reuse of fragment classes that use different viewmodels. Because fallback tag before was simply the title.

Default tag if not set in the attribute is now the java class name for the fragment.

Not really, adds a tag argument on the presentation attribute on fragments.

Navigate between fragments.

Test change presentation i.e. for navigating between tabs

- [x] All projects build

### DIFF
--- a/Directory.build.props
+++ b/Directory.build.props
@@ -22,7 +22,7 @@
     <GeneratePackageOnBuild Condition=" '$(Configuration)' == 'Release' and '$(IsTestProject)' != 'true'">true</GeneratePackageOnBuild>
     
     <Platform>AnyCPU</Platform>
-    <DebugType>embedded</DebugType>
+    <DebugType>portable</DebugType>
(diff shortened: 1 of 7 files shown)


Directory.build.props
```diff
@@ -22,7 +22,7 @@
     <GeneratePackageOnBuild Condition=" '$(Configuration)' == 'Release' and '$(IsTestProject)' != 'true'">true</GeneratePackageOnBuild>
     
     <Platform>AnyCPU</Platform>
-    <DebugType>embedded</DebugType>
+    <DebugType>portable</DebugType>
```
martijn00: Should this be changed?

Cheesebaron: Yes, read the commit message.

Issue #3106, opened 2018-09-15
Docs: Navigation is outdated


nmilcoff:
## 🐛 Bug Report

#3046 introduced quite a few changes and no docs were updated. [This document](https://www.mvvmcross.com/documentation/fundamentals/navigation) is therefore outdated.

Things I've noticed:
- It does not explain how cancellation works.
- The interface `IMvxNavigationService` has changed and the listed methods do not match anymore.

Cheesebaron:
To be honest, I don't think we should list the API there at all, since it is prone to change a lot over time.

However, adding the new changes introduced in 6.2 in the doc is 👍

PR #3131, opened 2018-09-27
#3106 - Updated API docs for IMvxNavigationService

markuspalme:
There is discussion about removing the API signature completely in #3106 - but I think it has some value if up to date.

The PR does not address the issue of missing docs for cancellation (yet).


martijn00:
Thanks!

PR #3132, opened 2018-09-27
Updated DI documentation to use new API

markuspalme:
The IoC related methods on the Mvx class are marked obsolete, the documentation now uses the methods on Mvx.IoCProvider instead (following up on https://github.com/MvvmCross/MvvmCross/pull/3122).


martijn00:
Thanks!

PR #3139, opened 2018-09-30
Fix circular references for UITableView and UICollectionView

nmilcoff:
Bug fix

Both CollectionViewSource and TableViewSource have strong references to UI widgets (UICollectionView, UITableView). This creates circular references [described here](https://docs.microsoft.com/en-us/xamarin/ios/deploy-test/performance#avoid-strong-circular-references).

This is something we were missing and although nobody reported this, I remember having read some complains in the slack channel.

Above is fixed. Proposed solution is to use weak references through WeakAttribute, [described here](https://docs.microsoft.com/en-us/xamarin/ios/deploy-test/performance#avoid-strong-circular-references) and also in this [Xamarin University video](https://www.youtube.com/watch?v=RTfrCwO3aPI).

Not really. But this change can potentially unmask issues related to other subscriptions. For example, if the subscription for collection changes of the ItemsSource property lives longer than it should, we will now see NRE because TableView/CollectionView will be null.

1. Open any MvvmCross app with a ViewController containing a UITableView / CollectionView.
2. Open that ViewController, override `Dispose` and place a log call there.
3. Run the app, navigate to that ViewController, and then pop it.
4. Nothing happens!
5. Apply the same change as I did, this time `Dispose` will be called and your log will appear.

Note: If you think it's necessary to have a test scenario on the Playground project I can add it, I just didn't do it this time because my VS / VS4Mac are out of sync.

-

- [X] All projects build

### DIFF
--- a/MvvmCross/Platforms/Ios/Binding/Views/MvxBaseCollectionViewSource.cs
+++ b/MvvmCross/Platforms/Ios/Binding/Views/MvxBaseCollectionViewSource.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MS-PL license.
 // See the LICENSE file in the project root for more information.
 
@@ -17,7 +17,7 @@ namespace MvvmCross.Platforms.Ios.Binding.Views
         public static readonly NSString UnknownCellIdentifier = null;
 
         private readonly NSString _cellIdentifier;
-        private readonly UICollectionView _collectionView;
+        [Weak] private UICollectionView _collectionView;
 
         protected virtual NSString DefaultCellIdentifier => _cellIdentifier;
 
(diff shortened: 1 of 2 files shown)


nmilcoff:
Interesting, does this happen the first time the ViewController is displayed, or when you navigate back to it? Is it possible for you to provide us with a repro sample? I can take a look at that.

ivmirx:
@nmilcoff on the first time. I will try to create a sample but not sure if I'll succeed because only one table controller behaves like this. I double-checked its setup code against two other table controllers in my app but couldn't find the difference (except that it has significantly more rows).

nmilcoff:
It's actually super strange. At some point the GC decides the Xamarin handle for the `TableView` is not necessary anymore. So it releases the handle and `TableView` becomes null. 

When you add anything like a logging sentence that causes the `TableView` handle to be re-created (important here is that only the handle is re-created, not the actual object). As a result, the Source.TableView will have a value.

Because the reference in our `MvxBaseTableViewSource` is weak, using the local member `TableView` there won't re-create the handle.

PR #3140, opened 2018-10-01
Updated documentation to use Mvx.IoCProvider instead of Mvx.

markuspalme:
Updated documentation to use Mvx.IoCProvider instead of Mvx.


martijn00:
It seems like this needs a rebase? It would also be good to create new branches for new PR's so your develop branch doesn't get messy.

nmilcoff:
@markuspalme any update? We would love to have this merged

PR #3172, opened 2018-10-23
Update color.md

xufeipyxis:
Add an example to introduce MvxIosColor's usage.


Cheesebaron:
I think it is more appropriate to show the usage of `ToNativeColor` instead, this extension works on every platform.

PR #3207, opened 2018-11-16
Incorrectly using TableView reference rather than tableView parameter…

AlanYost:
… reference.

Bug Fix

Sometimes the TableView property is null.  The method takes a tableView parameter that the method should be using as it does elsewhere in the code.

The result is a null reference exception.

The method now uses the tableView parameter passed into the method and no longer fails with a null pointer exception.

Should not be a breaking change 

Hard to reproduce in playground but occurs 100% of times in internal application.

- [ X] All projects build
- [ X] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contributing/mvvmcross-docs-style-guide))
- [ X] Rebased onto current develop


nickrandolph:
@Cheesebaron @martijn00 are you able to verify that this fix is safe? Was there a reason why this method was referencing the Property or field instead of the method parameter?

PR #3214, opened 2018-11-26
Remove calls to AppStart from normal Activities

nmilcoff:
### :sparkles: What kind of change does this PR introduce? (Bug fix, feature, docs update...)
Bug fix

### :arrow_heading_down: What is the current behavior?

As described in #3196, when an Android app does not use a SplashScreen, the first ViewModel is created twice:
- One time through the MvxAppStart mechanism
- One time in the Activity's `base.OnCreate` method

Also `Mvx.IoCProvider.Resolve` is used in SplashScreen classes and it's causing some crashes as described in #2810.

### :new: What is the new behavior (if this is a feature change)?

AppStart code is removed from normal activities. As a result, when an Android app does not use a SplashScreen the AppStart mechanism won't run, but this is the best experience we can achieve right now.

### :boom: Does this PR introduce a breaking change?

Probably, but it's for the greater good if fixing the issue (also I don't think this will have a major impact).

### :bug: Recommendations for testing

Comment out the SplashScreen in Playground.Droid and set `RootView` as `MainLauncher`. Set a breakpoint in `RootViewModel` - it should be called only once.

### :memo: Links to relevant issues/docs
Fixes #3196.
Fixes #3177 (no more double initial navigation)
Closes #3014.
Fixes #2810.

### :thinking: Checklist before submitting

- [X] All projects build
- [X] Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))
- [ ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contributing/mvvmcross-docs-style-guide))
- [X] Rebased onto current develop

PR #3246, opened 2019-01-08
Make it safe to try call appstart

martijn00:

Before we do the proposed change in https://github.com/MvvmCross/MvvmCross/pull/3214 we should merge this, and then think a bit more about changing the actual behavior.

Sometimes the app crashes when appstart is not available.

It doesn't crash on the call to resolve appstart

No

https://github.com/MvvmCross/MvvmCross/pull/3214

- [ ] All projects build


nickrandolph:
I like the intent but are there scenarios where it's safe for MvxAppStart not to resolve? At minimum we should work out how to log a message when AppStart doesn't resolve

nmilcoff:
I know I'm quite late here but I pretty much agree with @nickrandolph, it seems to me we are now masking some errors and not throwing exceptions in cases we should

PR #3269, opened 2019-01-23
Enhancing SetProperty with Action parameter

dogukandemir:
Feature

SetProperty doesn't have overloaded method with action parameter to run extra lines of codes easily.

SetProperty has new overloaded method with new parameter which is afterAction. This parameter makes multi binding easier than before. Check last commit on [data-binding.md](https://github.com/dogukandemir/MvvmCross/commit/3a68afd8598013bb94ef32cc0f70ea6a800f9a9a).

Before:
```
private string _foo;
public string Foo
{
    get => _foo;
    set
    {
        if (SetProperty(ref _foo, value))
        {
            Debug.WriteLine("some action");
            // more action
        }
    }
}
```

After:
```
private string _foo;
public string Foo
{
    get => _foo;
    set => SetProperty(ref _foo, value, () => { Debug.WriteLine("some action"); /* more action */ }); }
}
```

or
```
public void MyAction()
{
    Debug.WriteLine("some action"); 
    // more action
}

private string _foo;
public string Foo
{
    get => _foo;
    set => SetProperty(ref _foo, value, MyAction); }
}
```
This PR is not a breaking change. It only has new overloaded method.

If you assign any action to SetProperty method, check your method is invoked or not.

- [x] All projects build

### DIFF
--- a/MvvmCross/ViewModels/MvxNotifyPropertyChanged.cs
+++ b/MvvmCross/ViewModels/MvxNotifyPropertyChanged.cs
@@ -142,6 +142,18 @@ namespace MvvmCross.ViewModels
         }
 
         [NotifyPropertyChangedInvocator]
+        protected virtual bool SetProperty<T>(ref T storage, T value, Action afterAction, [CallerMemberName] string propertyName = null)
+        {
+            if (!SetProperty(ref storage, value, propertyName))
+            {
+                return false;
+            }
+
+            afterAction?.Invoke();
+            return true;
+        }
+
+        [NotifyPropertyChangedInvocator]
         protected virtual bool SetProperty<T>(ref T storage, T value, [CallerMemberName] string propertyName = null)
         {
             if (EqualityComparer<T>.Default.Equals(storage, value))
@@ -160,6 +172,7 @@ namespace MvvmCross.ViewModels
             RaisePropertyChanged(propertyName);
             return true;
         }
+
         protected virtual MvxInpcInterceptionResult InterceptRaisePropertyChanged(PropertyChangedEventArgs changedArgs)
         {
             if (MvxSingletonCache.Instance != null)
--- a/MvvmCross/ViewModels/MvxNotifyPropertyChanged.cs
+++ b/MvvmCross/ViewModels/MvxNotifyPropertyChanged.cs
@@ -142,15 +142,21 @@ namespace MvvmCross.ViewModels
         }
 
         [NotifyPropertyChangedInvocator]
+        protected virtual void SetProperty<T>(ref T storage, T value, Action<bool> action, [CallerMemberName] string propertyName = null)
+        {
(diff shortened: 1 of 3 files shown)


MvvmCross/ViewModels/MvxNotifyPropertyChanged.cs
```diff
@@ -141,6 +141,18 @@ void raiseChange()
...
 
+        [NotifyPropertyChangedInvocator]
+        protected virtual bool SetProperty<T>(ref T storage, T value, Action afterAction, [CallerMemberName] string propertyName = null)
+        {
+            if (!SetProperty(ref storage, value, propertyName))
+            {
+                return false;
+            }
+
+            afterAction?.Invoke();
```
martijn00: So this will only be executed when it is true. I'm not sure that makes it always usable. Why not make the afterAction, `Action<bool>` and always call it, with the result as parameter.

dogukandemir: I propose having both ideas at the same time. Initial idea was to get rid of if condition when you want to do something after set property (especially if it's only one line of code). 

```
[NotifyPropertyChangedInvocator]
protected virtual void SetProperty<T>(ref T storage, T value, Action<bool> action, [CallerMemberName] string propertyName = null)
{
    action?.Invoke(SetProperty(ref storage, value, propertyName));
}

[NotifyPropertyChangedInvocator]
protected virtual bool SetProperty<T>(ref T storage, T value, Action afterAction, [CallerMemberName] string propertyName = null)
{
    if (SetProperty(ref storage, value, propertyName))
    {
        afterAction?.Invoke();
        return true;
    }

    return false;
}
```


If you agree, I will edit Multi-binding part of data-binding documentation like this (to show two alternative usages).

**Multi-binding**

In Swiss binding, each binding can only reference a single ViewModel property path.

This meant that if a ViewModel had 2 properties like `FirstName` and `LastName`, then the main way to create a display of the ful name was to create a new ViewModel property - e.g.:

```c#
private string _firstName;
public string FirstName
{
    get => _firstName;
    set => SetProperty(ref _firstName, value, () => RaisePropertyChanged(() => FullName));
}

private string _lastName;
public string LastName
{
    get => _lastName;
    set => SetProperty(ref _lastName, value, (setPropertyResult) => 
    {
        if (setPropertyResult)
        {
            RaisePropertyChanged(() => FullName);
        }
    });
}

public string FullName => _firstName + " " + _lastName;
```

nickrandolph: I really like the option of having Action<bool> and Action. Are you able to commit your code and let's get this feature merged.

Cheesebaron:
Since this is essentially a breaking change, I would wait with merging this to after 6.2.3 has been released. So probably for 6.3.0

nickrandolph:
Ummm I thought in current format it's not breaking any more?

Issue #3289, opened 2019-02-17
NavigationService doesn't wait for the view to appear in ios


AbdelrahmanGIT:
## 🐛 Bug Report

in ios When navigating through navigation service to another view model the navigation will run and return before the navigation is completed, will also fire `AfterNavigation` Event before navigation has completed.

this happens because the code responsible for showing view model in `MvxIosViewDispatcher` doesn't know about the task being executed as it's action is stored in `Action` instead of `Func<Task>` which in turn results on an asynchronous execution

### Expected behavior
When awaited it should not return until the navigation is completed

### Reproduction steps
Navigate from ViewModelA to ViewModelB and do something after navigation it will execute before the navigation is completed in ios only

### Configuration

**Version:** 6.x

**Platform:** 
- [x] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [x] :monkey: Xamarin.Forms

PR #3291, opened 2019-02-18
Return the task to properly await showing view model

AbdelrahmanGIT:
Bug fix

NavigationService doesn't wait for the view to appear in ios

NavigationService is now aware of the showing task and will properly wait for it to complete before returning

No

#3289 

- [x] All projects build


martijn00:
Seems valid, but shouldn't this be changed on all the platforms?

AbdelrahmanGIT:
> Seems valid, but shouldn't this be changed on all the platforms?

mmm I was testing initially in Android and Ios, but after some digging the issue seems to be present in `mac` `tizen` and `tvos` as well

martijn00:
@AbdelrahmanGIT so this doesn't need to be changed on UWP or Android?

AbdelrahmanGIT:
@martijn00 It seems so, since `Android` and `UWP` are returning the task through inline lambda `=> result`

Issue #3158, opened 2018-10-11
ViewDestroy lifecycle method is broken


arkadiuszokon:
## 🔙 Regression

commit which breaks this: https://github.com/MvvmCross/MvvmCross/pull/3006/commits/32b6868e84740e6b7c35a56822736702989b325e

How we should use ViewDestroy now? It's pointless - it's called every time when application goes to background with "viewFinishing" flag set by default to true. 

### Old (and correct) behavior
ViewDestroy was called with corresponding native methods of view lifecycle

### Current behavior
ViewDestroy is called with OnDisappearing method in mvx view classes.

### Configuration

**Version:** 6.2

**Platform:** 
- [ ] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [x] :monkey: Xamarin.Forms

piotrekbigu:
Hi,
When we can expect a version where this issue will be fixed?

ViewDestroy method is very important in view lifecycle (releasing resources, unsubscribing, etc) and fact that is called whenever view disappears is really critical bug.

nmilcoff:
I think ViewDestroy was not being called before, since obviously classes like `ContentPage` do not extend any MvvmCross class.

I'm not sure removing the call to `ViewDestroy` for all Xamarin Forms views is the best solution here.

Cheesebaron:
@piotrekbigu please don't ask when it is going to be fixed. Instead, if it is such a pressing matter for you, open a PR fixing this and it will most likely get out next time we release.

keannan5390:
This could can be fixed by creating custom renderers for all of the MvvmCross pages in each platform and making to call to `ViewDestory` from the `Dispose` method of the renderer. I've verified that this works on Android & iOS. I'm would assume the same would be the case for other platforms.

```
protected override void Dispose(bool disposing)
{
    if (disposing)
    {
        var viewModel = this.Element?.BindingContext as IMvxViewModel;
        viewModel?.ViewDestroy();
    }

    base.Dispose(disposing);
}
```

I'm not setup to build/develop on the mvvmcross solution on my machine (i don't have a windows vm). Otherwise I'd look at submitting a PR for this.

keannan5390:
[...]

PR #3292, opened 2019-02-18
3158 ViewDestroy lifecycle method is broken

abdomh:
Bug fix

ViewDestroy is called when view disappears and doesn't reflect an actual view destroy.

ViewDestroy is called when the view is popped from a `MvxNavigationPage` or from a application wide modal pop.

I have implemented this for pages not really sure about `MvxEntryCell , MvxImageCell ... etc` as they still call `ViewDestroy` in `Disappear`

Yes (Different behavior)

One way to notice this is by showing a viewmodel that returns a result and the go to another app (or load another view model) the view model will issue a cancel result because it thinks it got destroyed.

#3158 

- [x] All projects build
- [x] Implementing something similar  for `IMvxCell` if it's possible
- [x] Implementing `ViewDestroy` for `MainPage`

### DIFF
--- a/MvvmCross.Forms/Core/MvxFormsApplication.cs
+++ b/MvvmCross.Forms/Core/MvxFormsApplication.cs
@@ -3,6 +3,8 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using MvvmCross.Forms.Views;
+using MvvmCross.ViewModels;
 using Xamarin.Forms;
 
 namespace MvvmCross.Forms.Core
@@ -11,6 +13,7 @@ namespace MvvmCross.Forms.Core
     {
         protected MvxFormsApplication()
         {
+            this.ModalPopped += OnModalPopped;
         }
 
         public event EventHandler Start;
@@ -32,6 +35,14 @@ namespace MvvmCross.Forms.Core
         protected override void OnResume()
         {
             Resume?.Invoke(this, EventArgs.Empty);
+        }
+
+        private void OnModalPopped(object sender, ModalPoppedEventArgs e)
+        {
+            if (e.Modal is IMvxPage mvxPage && mvxPage.ViewModel is IMvxViewModel modalViewModel)
+            {
+                modalViewModel.ViewDestroy(true);
+            }
         }
     }
 }
--- a/MvvmCross.Forms/Core/MvxFormsApplication.cs
+++ b/MvvmCross.Forms/Core/MvxFormsApplication.cs
@@ -13,7 +13,7 @@ namespace MvvmCross.Forms.Core
     {
         protected MvxFormsApplication()
         {
-            this.ModalPopped += OnModalPopped;
+            ModalPopped += OnModalPopped;
         }
 
         public event EventHandler Start;
@@ -37,12 +37,20 @@ namespace MvvmCross.Forms.Core
             Resume?.Invoke(this, EventArgs.Empty);
         }
 
+        protected override void OnPropertyChanging(string propertyName = null)
+        {
+            if (propertyName == nameof(MainPage)
+                && MainPage is IMvxPage mvxPage
+                && mvxPage.ViewModel is IMvxViewModel mvxViewModel)
+                mvxViewModel.ViewDestroy(true);
+
+            base.OnPropertyChanging(propertyName);
+        }
+
         private void OnModalPopped(object sender, ModalPoppedEventArgs e)
         {
             if (e.Modal is IMvxPage mvxPage && mvxPage.ViewModel is IMvxViewModel modalViewModel)
-            {
                 modalViewModel.ViewDestroy(true);
-            }
         }
     }
 }
--- a/MvvmCross.Forms/Views/MvxCarouselPage.cs
+++ b/MvvmCross.Forms/Views/MvxCarouselPage.cs
@@ -89,7 +89,6 @@ namespace MvvmCross.Forms.Views
             base.OnDisappearing();
             ViewModel?.ViewDisappearing();
             ViewModel?.ViewDisappeared();
-            ViewModel?.ViewDestroy();
         }
     }
 
--- a/MvvmCross.Forms/Views/MvxContentPage.cs
+++ b/MvvmCross.Forms/Views/MvxContentPage.cs
@@ -90,7 +90,6 @@ namespace MvvmCross.Forms.Views
             base.OnDisappearing();
             ViewModel?.ViewDisappearing();
             ViewModel?.ViewDisappeared();
-            ViewModel?.ViewDestroy();
         }
     }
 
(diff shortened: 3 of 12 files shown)


nmilcoff:
I like this approach - at least makes the method usable for most scenarios. Should we also consider MainPage changes?

AbdelrahmanGIT:
> I like this approach - at least makes the method usable for most scenarios. Should we also consider MainPage changes?

mmm Maybe listening to `PropertyChanged` on `Application` for `MainPage` will be sufficient? i might do that today and see

also do you have any idea about how to we can detect `IMvxCell`  ViewDestroy? since this still uses `Cell` on disappearing event.

nmilcoff:
> mmm Maybe listening to PropertyChanging on Application for MainPage will be sufficient? i might do that today and see

This sounds reasonable, yes.

> also do you have any idea about how we can detect IMvxCell ViewDestroy? since this still uses Cell on disappearing event.

I just went through the classes in X.F and I don't think there's a clean/easy way to do this. But probably removing the call to ViewDestroy makes sense.

martijn00:
I don't think this is a good approach. We don't force people to use MvxFormsApplication so there is no way to know if it is used inside a page. Any existing apps that use a close navigation will break this way.

I don't understand why this is a problem for now? The close cancelation token needs to be triggered when a page is closed, so if that isn't the behavior we want, we should change that instead.

AbdelrahmanGIT:
> I don't understand why this is a problem for now? The close cancelation token needs to be triggered when a page is closed, so if that isn't the behavior we want, we should change that instead.

As were it stands now `ViewDestroy` doesn't reflect it's meaning and will fire wither you are closing the view or you just changing for another app to get something(or opening another view) and return back, which is different from `MvvmCross Xamarin Native` behavior.

 it gives the view-model the impression that it's being closed while in reality it's only being disappeared 
 which in turn results in `ViewDestroy` can't be used for cleanup ex disposing or closing an active connection

>The close cancelation token needs to be triggered when a page is closed, so if that isn't the behavior we want, we should change that instead

that's where the problem is most visible right now closing or opening a view-model over a result-view-model will cause the view model to return a cancellation result since there is miscommunication between the view and the view model and the user will still be able interact with result-view-model and trigger a result without the app knowing about it which hurts the user experience greatly.

MysterDru:
I think i found a more reliable way to properly call the OnDestroy lifecycle of the view models that doesn't rely on developers extending from the MvxFormsApplication or an MvxRenderer. Some variation of the following code should work.

Whenever a renderer is disposed at the platform layer, the attached `Renderer`  property for the element is set to null. This is a non public value, so there is certainly some risk with doing it this way since it relies on reflection, but it seems like it might be the most reliable.

The following could be added to each of the Mvx page implementations:

```
protected override void OnPropertyChanged([CallerMemberName] string propertyName = null)
{
    base.OnPropertyChanged(propertyName);

    // when Dispose is called on the renderers, the body of this if statement will be set executed
    // the Dispose code of the renderers will set the attached bindable property for "Renderer" on the 
   // forms platform class. So if Renderer property changed, and its value is null, we can assume that the view was destroyed
    if(propertyName == "Renderer" && this.GetValue(this.GetRendererProperty()) == null)
    {
        this.ViewModel.ViewDestroy(true);
    }
}
```

And the implementation for GetRendererProperty is this:

```
public static BindableProperty GetRendererProperty(this Page formsPage)
{
    // borrowed from James Monetemagno's PullToRefresh layout control
    // https://github.com/jamesmontemagno/Xamarin.Forms-PullToRefreshLayout/blob/ad63e21a5010dd04b570a89eca5c5287ac36e0ed/PullToRefresh/PullToRefresh.Standard/PullToRefreshLayoutRenderer.android.cs#L132

    var type = Type.GetType("Xamarin.Forms.Platform.Android.Platform, Xamarin.Forms.Platform.Android");
    var prop = type.GetField("RendererProperty", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static);
    var val = prop.GetValue(null);
    var rendererProperty = val as BindableProperty;

    return rendererProperty;
}
```

Hackmodford:
This issue just built me. I tried @MysterDru solution and it seemed to work better. Should I create a PR that just implements that?

henda79:
Is this project dead ?

Hackmodford:
No ?

henda79:
You don't seem so sure.

Issues are reported, people ask for a fix, they are told to create a PR and fix themselves. A fix for this issue was presented 3 years ago, a PR was created several months ago, by you if I'm not mistaken and now it just sits there.

Hackmodford:
True, but I think the big problem is everyone is in the middle of a transition to .net 6. Once .NET MAUI is released, I'd hope these issues could pick up speed. :)

henda79:
I'd like to think that, but I can see MAUI been the same as Xamarin, no one wants to adopt early as it will be buggy and by the time it starts to get stable like Xamarin 5, it gets ditched, just to start the same thing over again.

Anyway, enough rant. I'll try to compile a version myself with the bug fixes as I can't see the NuGet getting updated soon.

henda79:
@Hackmodford , so I've spent a day getting the environment sorted to build and debug MVVMCross, as its bear impossible to find issues otherwise. I've noticed that this PR only will only call ViewDestroyed *if* the page is a modal. The ViewDestroyed method is never called for other pages such as MvxContentPage.

Was this intended ? I would have expected it to be called regardless of the presentation method.

Hackmodford:
@henda79 PR sounds like it was trying to fix that. See this comment for another possible fix. https://github.com/MvvmCross/MvvmCross/pull/3292#issuecomment-482335537

henda79:
@Hackmodford , after a bit of thought, I've decided the following.

IMHO: ViewDestroy should be called when closing the page (modal or not) and at the moment as of 8.0.2, it get called when navigating forwards and backwards.

My solution, since I only use MvxNavigationService.Close, is to call ViewDestroy within OnWillClose (MvxNavigationService), or change the Close method to call it before OnDidClose.

The only other thing which is required is to remove all the call to ViewDestroy in the Views as per this PR. See the code changes.

Hackmodford:
Destroy should be called when navigating forward as well.

Navigating from A -> B means A's `Destroy` should be called.
Navigating from A -> modal B should not call A's destroy.

henda79:
@Hackmodford , thats seems to be the cause of the problem.

If you are going A -> B (non modal) -> C (non modal)

but A is awaiting on B as it returns a Result. As soon as you navigate to C, ViewDestroyed is called on B and no result is returned.

You can see the issue ?

Hackmodford:
True... though bear in mind I think awaiting results from viewModels is being removed from this library as it's unreliable on Android.

henda79:
@Hackmodford , it that's the case then maybe I need to refactor the little work already done, rather than waiting until the project is completed.

I've not noticed any issues up to now waiting for results from VM's other than the one I've mentioned.

Is there a roadmap to remove this feature or other reports I should consider ?

DuncanMT:
This issue: 955975

henda79:
Thanks, I think i'll refactor and remove that function.

PR #3338, opened 2019-03-25
Use license expression for nuget package

martijn00:

Feature

No

- [ ] All projects build

### DIFF
--- a/Directory.build.props
+++ b/Directory.build.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <Copyright>Copyright (c) .NET Foundation and Contributors</Copyright>
-    <PackageLicenseUrl>https://opensource.org/licenses/ms-pl</PackageLicenseUrl>
+    <PackageLicenseExpression>MS-PL</PackageLicenseExpression>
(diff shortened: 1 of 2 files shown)


Directory.build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <Copyright>Copyright (c) .NET Foundation and Contributors</Copyright>
-    <PackageLicenseUrl>https://opensource.org/licenses/ms-pl</PackageLicenseUrl>
+    <PackageLicenseExpression>MS-PL</PackageLicenseExpression>
```
Cheesebaron: I think in addition to using this new property we should add the license to the NuGet package too:

```msbuild
<ItemGroup>
    <None Include="licenses\LICENSE.txt" Pack="true" PackagePath=""/>
</ItemGroup>
```

martijn00: What would that help?

Cheesebaron: You wouldn't have to go somewhere else to get the actual license for the library. Also for the license to actually cover the distributed code, it must be accompanied with it.

PR #3356, opened 2019-04-05
Constructor overload for object creation during dependency injection

nickrandolph:
### :sparkles: What kind of change does this PR introduce? (Bug fix, feature, docs update...)
Perf improvement

### :arrow_heading_down: What is the current behavior?
Types need to be constructed using reflection or via an Func delegate

### :new: What is the new behavior (if this is a feature change)?
Can supply a Func that takes type resolved arguments and a call to a constructor, eliminating the need for reflection to inspect the type being created

### :boom: Does this PR introduce a breaking change?
No

### :bug: Recommendations for testing
Run all unit tests

### :thinking: Checklist before submitting

- [X] All projects build
- [X] Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))
- [ ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contributing/mvvmcross-docs-style-guide))
- [X] Rebased onto current develop

PR #3359, opened 2019-04-08
Adding parameter to Register methods to determine whether existing registration should be overwritten

nickrandolph:
Feature

With successive calls to RegisterType, each call replaces the registered resolver. This means that if a developer registers a resolver prior to startup, mvvmcross will overwrite the resolver with the default for that type.

The overrideIfExists parameter determines whether the existing resolver will be overwritten.
This allows for the core of mvvmcross to be updated to only register resolvers where they haven't been already registered, giving more flexibility to developers

No

Blocked waiting on PR: https://github.com/MvvmCross/MvvmCross/pull/3356

- [ X ] All projects build

### DIFF
--- a/MvvmCross/IoC/IMvxIoCProvider.cs
+++ b/MvvmCross/IoC/IMvxIoCProvider.cs
@@ -34,26 +34,26 @@ namespace MvvmCross.IoC
 
         object GetSingleton(Type type);
 
-        void RegisterType<TFrom, TTo>()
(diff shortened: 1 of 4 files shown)


nickrandolph:
@martijn00 @Cheesebaron this has been rebased. I'll try to get around to adding some test cases. The point of this is that it gives us more control when registering types to prevent overriding prior registrations. This is more of an enabler than specifically adding functionality.

nickrandolph:
Withdrawing this PR as don't think it's needed right now

PR #3365, opened 2019-04-09
Fix log format crashing for MvxBindingLog

### LINKED ISSUE #3337: MvxBindingLog should not format messages + fix
## 🐛 Bug Report

In BindingLog.cs:

```
        private static void Trace(MvxLogLevel level, string message, params object[] args)
        {
            if (level < TraceBindingLevel) return;

            Instance.Log(level, () => string.Format(message, args));
        }
```

will crash with a FormatException if the formatting fails.The original message is lost.

### Expected behavior

Does not crash

### Reproduction steps

Too hard, but happens!!

### Solution

Replace ```Instance.Log(level, () => string.Format(message, args));```  
With ```Instance.Log(level, () => message, null, args);```

The log provider will then catch the error. If a custom log provider is used, it will be able to display more informations, including why the format failed. This information is currently not available.

### Configuration

**Version:** 6.2.3

**Platform:** 
All (MvxBindingLog is common)

- [X] :iphone: iOS
- [X] :robot: Android
- [X] :checkered_flag: WPF
- [X] :earth_americas: UWP
- [X] :apple: MacOS
- [X] :tv: tvOS
- [X] :monkey: Xamarin.Forms

softlion:
See #3337

See #3337

When a message containing '{some number}' is formatted by this class, it crashes because it has no guard against it.
This crash is reported by the log, the original crash captured by this class is lost.

See #3337

Delegates the formatting of the error message to MvxLog classes, instead of formatting it there.

No

- [X] All projects build

### DIFF
--- a/MvvmCross/Binding/MvxBindingLog.cs
+++ b/MvvmCross/Binding/MvxBindingLog.cs
@@ -16,7 +16,7 @@ namespace MvvmCross.Binding
         {
             if (level < TraceBindingLevel) return;
 
-            Instance.Log(level, () => string.Format(message, args));
+            Instance.Log(level, () => message, null, args);
         }
 
         public static void Trace(string message, params object[] args)


nickrandolph:
Please create a test case that validates this behaviour. Ideally the test case should fail prior to the change, and pass with the change made.

softlion:
The test case would involve creating triggering a bug in Mvvmcross. The fix is easy enough to understand to not need any test case. I've tons of logs from my yesterday coding session where this happens. And it is really a pain as the real crashing lines are not available because of this bug. @martijn00  plz.

softlion:
A more friendly answer

The test case would involve a lot of efforts to trigger a bug in Mvvmcross where the binding would fail with a message containing a {10} which would then be incorrectly formatted  (it would crash in fact).

No class calling into MvxLog should format message with non secured arguments themselves, because 1) it requires safety and 2) the MvxLog classes already do it better.

I've tons of logs from my yesterday coding session where this happens because of a 'random' bug in the android's RecyclerView that is now fixed. 
And it is really a pain as the real crashing lines are not available because of this bug.

nickrandolph:
Unless I'm missing something, the test case can call the public static Trace method with a message/args combination that will cause string.format to fail. 
I'm not against this change, and to be honest I'm not sure why the stirng.format call is there at all since the base logger does that anyhow.

softlion:
It would be a test caseof the "String.Format" method which is a nonsense.

softlion:
If you don't want the pr i remove them all.

nickrandolph:
@softlion I'm not sure why you're getting annoyed by this, I'm not asking you to write a test case to verify string.format. I'm asking you to write a test that would break string.format and that would pass after you changes are approved. 
What you don't appreciate is that by adding this test case we'll then see that we have no protection for when the same formatting string breaks whatever log provider is being selected. 
I agree with your fix, but it doesn't actually fix the issue that we pass the message (including formatting tags eg {0} ) through to the log provider along with the args. This could potentially cause issues if the log provider throws an error itself.
Again I didn't mean any offence by this but don't just assume that because your PR is good code, that it's sufficient to get accepted without further work.

softlion:
You can get unit test for String.Format in https://github.com/dotnet/corefx/tree/master/src/Common/tests/System

I don't have time for this.

softlion:
"the original crash captured by this class is lost."

PR #3366, opened 2019-04-09
Fix crash and improve MvxRecyclerViewAdapter and MvxRecyclerViewHolder

### LINKED ISSUE #3295: MvxRecycler view must not post NotifyDataSetChanged on MainLooper
## 🔙 Regression

Introduced by the fix for #2144

There may be a delay between the change of the source collection and the calls to NotifyItemXXX methods of the adapter, resulting in desynchronizations of the recycler's view and its datasource, thus generating random crashes when the recyclerview try to load a removed item.

### Old (and correct) behavior

When an ObservableCollection is structurally updated in a viewmodel, and this collection is bound to the ItemsSource of a RecyclerView's adapter, the adapter subscribes to INotifyCollectionChanged and dispatch these events  to NotifyDataSetChanged().

If the viewmodel's collection is changed on a worker thread, the app will crash. But there will be no delay between the change of the collection and the recycler's view notifications.

### Current behavior

To prevent crashes when a viewmodel's collection is changed on a worker thread, the adapter posts the change on the MainLooper. This introduces the delay explained at the start of this bug report and leads to random crashes.

```
            if (Looper.MainLooper == Looper.MyLooper())
            {
                NotifyDataSetChanged(e);
            }
            else
            {
                var h = new Handler(Looper.MainLooper);
                h.Post(() => NotifyDataSetChanged(e));
            }
```

### Reproduction steps

1. Connect an ObservableCollection to an MvxRecyclerView, add lots of items.
2. Remove an item every second in your viewmodel from a worker thread (timer)
3. scroll the list in the app, it will crash at some point.

### Configuration

**Version:** 6.x

**Platform:** 
- [ ] :iphone: iOS
- [X] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

softlion:
Bug fix + refactor of MvxRecyclerViewAdapter and MvxRecyclerViewHolder

- Potential crash when updating ItemsSource from worker thread.
- Inconsistencies when changing Adapter, making ItemClick and ItemLongClick not work until items are recycled and shown again

Fixes the above crash and inconsistencies
Allows extensibility through new virtual methods:
- OnItemViewClick
- OnItemViewLongClick
- ExecuteCommandOnItem

Also improved cleaning up of resources when Dispose is called from the Java side

No

Fixes #3295

- [X] All projects build


MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
```diff
@@ -241,15 +241,10 @@ protected virtual void SetItemsSource(IEnumerable value)
...
-            {
-                NotifyDataSetChanged(e);
-            }
-            else
-            {
-                var h = new Handler(Looper.MainLooper);
-                h.Post(() => NotifyDataSetChanged(e));
-            }
+            if (Looper.MainLooper != Looper.MyLooper())
+                MvxBindingLog.Error("All collection changes ");
```
Cheesebaron: Is there something missing from the message here?

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerView.cs
```diff
@@ -33,6 +28,19 @@ public MvxRecyclerView(Context context, IAttributeSet attrs, int defStyle)
...
         }
 
+        [Preserve(Conditional = true)]
+        protected MvxRecyclerView(IntPtr javaReference, JniHandleOwnership transfer)
+            : base(javaReference, transfer)
+        {
+        }
+
+        [Preserve(Conditional = true)]
+        protected static void LinkerPleaseInclude(MvxRecyclerView list)
```
Cheesebaron: I don't think this should go here, but rather in a LinkerPleaseInclude file.

martijn00:
@softlion can you address the comments?

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerView.cs
softlion: I do not agree. The "Conditional" property indicates to the linker that if the MvxRecyclerView is used, then it should preserve the method. If not, it should optimize it away.

There are 3 advantages to have it here. 1) you don't have to add it elsewhere. 2) If something else should not be optimized away, or if in the future something is not required anymore, it will be in the same source file. 3) If RecyclerView is not used, it will be optimized away and not kept.

softlion:
I've fixed things + added other fixes. Creating a new PR.

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
softlion: fixed

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
```diff
@@ -27,70 +26,52 @@ namespace MvvmCross.Droid.Support.V7.RecyclerView
...
 
-        protected IMvxAndroidBindingContext BindingContext => _bindingContext;
-
-        public MvxRecyclerAdapter() : this(MvxAndroidBindingContextHelpers.Current())
-        {
-        }
+        protected IMvxAndroidBindingContext BindingContext { get; }
 
-        public MvxRecyclerAdapter(IMvxAndroidBindingContext bindingContext)
+        public MvxRecyclerAdapter(IMvxAndroidBindingContext bindingContext = null)
```
martijn00: Doesn't Android need an empty ctor? Maybe just leave it as it was?

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
```diff
@@ -215,54 +191,47 @@ protected virtual int GetItemsSourcePosition(int viewPosition)
 
         protected virtual void SetItemsSource(IEnumerable value)
         {
+            if (Looper.MainLooper != Looper.MyLooper())
+                Mvx.IoCProvider.Resolve<IMvxLog>().Error("ItemsSource property set on a worker thread. This leads to crash in the RecyclerView. It must be set only from the main thread.");
```
martijn00: Use MvxLog.Instance.

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
```diff
@@ -215,54 +191,47 @@ protected virtual int GetItemsSourcePosition(int viewPosition)
...
             if (Looper.MainLooper == Looper.MyLooper())
-            {
                 NotifyDataSetChanged(e);
-            }
             else
-            {
-                var h = new Handler(Looper.MainLooper);
-                h.Post(() => NotifyDataSetChanged(e));
-            }
+                Mvx.IoCProvider.Resolve<IMvxLog>().Error("ItemsSource collection content changed on a worker thread. This leads to crash in the RecyclerView as it will not be aware of changes immediatly and may get a deleted item or update an item with a bad item template. All changes must be synchronized on the main thread.");
```
martijn00: Use MvxLog.Instance.

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
```diff
@@ -290,5 +252,25 @@ protected virtual void OnMvxViewHolderBound(MvxViewHolderBoundEventArgs obj)
         {
             MvxViewHolderBound?.Invoke(obj);
         }
+
+        /// <summary>
+        /// Always called with disposing = false, as it is only disposed from java
+        /// </summary>
+        protected override void Dispose(bool disposing)
+        {
+            //Mvx.IoCProvider.Resolve<IMvxLog>().Trace("MvxRecyclerAdapter Dispose");
```
martijn00: Cleanup

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
```diff
@@ -290,5 +252,25 @@ protected virtual void OnMvxViewHolderBound(MvxViewHolderBoundEventArgs obj)
...
+
+            //if (disposing)
```
martijn00: //if (disposing) ? Cleanup please.

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerView.cs
martijn00: @softlion i see your point, but if we do it this way we should probably change it all over the code base. Maybe you can remove this for now, and add a new issue to pick this up separately?

softlion:
> i see your point, but if we do it this way we should probably change it all over the code base. Maybe you can remove this for now, and add a new issue to pick this up separately?

If you don't start one day, you will never start.

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerView.cs
softlion: We should start at some point. Otherwise you will never start.

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
softlion: ok

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
softlion: MvxLog is internal and not available is this assembly.

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
softlion: ok

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
softlion: ko

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
softlion: MvxLog is internal and not available is this assembly.

martijn00:
@softlion we use rebase instead of merge to get changes from develop into the branch you make changes in. I did one on your branch, but that means that you have to delete your local branch and check it out again because the history is overwritten.

If you want i can do it again to get rid of the merge commit.

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
martijn00: https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross.Android.Support/V7.RecyclerView/MvxAndroidLog.cs#L11

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
martijn00: https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross.Android.Support/V7.RecyclerView/MvxAndroidLog.cs#L11

softlion:
> @softlion we use rebase instead of merge to get changes from develop into the branch you make changes in. I did one on your branch, but that means that you have to delete your local branch and check it out again because the history is overwritten.
> 
> If you want i can do it again to get rid of the merge commit.

I'm sorry i try to always use rebase, and it should always rebase. But as soon as something is wrong i am unable to fix it. I even don't know what is wrong. So i'll be unable to fix this.

MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
softlion: Its not MvxLog, it's MvxAndroidLog.

Cheesebaron:
I've rebased it for you

softlion:
> I've rebased it for you. Can you please check the File Changed tab and verify that those are the correct changes? @softlion

You can't rebase my repository, you don't have access to it. I don't know what you mean.

Cheesebaron:
@softlion actually I do have access to it. As soon as you make a Pull Request, I have access to push to that branch. Unless you remove the check mark in the checkbox on the PR, which allows/disallows others to push to it.

Cheesebaron:
This looks good to me. Also tested it on a sample app I made, which tested both, ItemClick, ItemLongClick, adding, removing items, adding items in bulk. Everything seems OK.

fadaEntrepidus:
If I had my ViewHolder like this: `public override RecyclerView.ViewHolder 

`OnCreateViewHolder(ViewGroup parent, int viewType){
           var bc = new MvxAndroidBindingContext(parent.Context, BindingContext.LayoutInflaterHolder);
            var v = InflateViewForHolder(parent, viewType, bc);
            return new ViewHolder(v, bc)
            {
                Click = ItemClick,
                LongClick = ItemLongClick
            };
            return new ViewHolder(v, bc); 
}` 

`public override void OnAttachedToWindow()
            {
                try
                {
                    base.OnAttachedToWindow();
                    ...
                }
                catch (Exception ex)
                {
                    Mvx.IoCProvider.Resolve<ILogger>().LogException("TAG", ex, nameof(View.Click));
                }
            }`


How should I be with the current changes? Because it gives me this error: 

> MonoDroid: UNHANDLED EXCEPTION: 

> MonoDroid: System.ArgumentNullException: missing source event info in MvxWeakEventSubscription 

> MonoDroid: Parameter name: sourceEventInfo

> MonoDroid:   at MvvmCross.WeakSubscription.MvxWeakEventSubscription`2[TSource,TEventArgs]..ctor (Android.Views.View source, System.Reflection.EventInfo sourceEventInfo, System.EventHandler`1[TEventArgs] targetEventHandler) [0x0001d] in <d4f0f9c16e25440aaf673bb2434bc101>:0

tbalcom:
You may be running into the issue described in #3411. Check there for a solution.

fadaEntrepidus:
> You may be running into the issue described in #3411. Check there for a solution.

Yesss!! Work. Thanks @tbalcom

SamuelDebruyn:
This did introduce a breaking change by adding events to the IMvxRecyclerViewHolder. Maybe some documentation should be added.

anne-k:
This change removes public and protected items from the MvxRecyclerViewHolder, which does in fact make it breaking imo. Sure, none of them were in the documentation, but if they're accessible you can expect they will be used.

To be more specific, what this breaks in my app is some of my custom ViewHolders/Adapters. That is a usage not covered at all in the documentation but I don't consider it a strange thing to do. The MvxRecyclerView can cover a lot of use cases with axml and TemplateSelectors only, but not all of them :)

softlion:
@anne-k you are right it may change the behavior of some custom changes made by advanced users, but they are very few and have the capacity of finding the problem, sometime with passion, sometime with cold blood.

This change is for the good, and has been adapoted a long time ago by all big android libs providing custom recycler views.

The goal of this change is to make sure that a cell's id is unique throughout all the recyclerviews of an app, so the hidden "recycler pools" work as expected. Recycler view pools cache and reuse cell views that are in the same activity. As most devs now use fragments, they have only one activity, and one recycler view pool. With this change the unicity is now implicit as it is "forced".

anne-k:
@softlion Ah, I didn't know that! That's interesting, that makes it make more sense. It's a bit unfortunate that it doesn't match with the TemplateSelector's viewtype anymore but changing that would break stuff for a lot more people, I get that you don't want to do that.

Thank you for replying!

softlion:
@anne-k Ha i understand what you mean. I'm not using the mvvmcross template selector, that's why i was not aware of the impact. I'm using my own which knows both the id in the selector and the matching layout id.

PR #3372, opened 2019-04-10
Tidying up and making startup more consistent

nickrandolph:
Improved consistency of startup

Overriding startup actions isn't consistent

Startup actions are more consistent

Yes

TBD

- [ X ] All projects build

### DIFF
--- a/MvvmCross.Forms/Platforms/Android/Core/MvxFormsAndroidSetup.cs
+++ b/MvvmCross.Forms/Platforms/Android/Core/MvxFormsAndroidSetup.cs
@@ -21,6 +21,7 @@ using MvvmCross.Forms.Platforms.Android.Presenters;
 using Xamarin.Forms;
 using MvvmCross.Droid.Support.V7.AppCompat;
 using MvvmCross.Core;
+using MvvmCross.IoC;
 
 namespace MvvmCross.Forms.Platforms.Android.Core
 {
@@ -39,10 +40,11 @@ namespace MvvmCross.Forms.Platforms.Android.Core
             return _viewAssemblies;
         }
 
-        protected override void InitializeIoC()
+        protected override IMvxIoCProvider InitializeIoC()
         {
-            base.InitializeIoC();
-            Mvx.IoCProvider.RegisterSingleton<IMvxFormsSetup>(this);
+            var provider = base.InitializeIoC();
+            provider.RegisterSingleton<IMvxFormsSetup>(this);
+            return provider;
         }
 
         protected override void InitializeApp(IMvxPluginManager pluginManager, IMvxApplication app)
--- a/MvvmCross.Forms/Platforms/Ios/Core/MvxFormsIosSetup.cs
+++ b/MvvmCross.Forms/Platforms/Ios/Core/MvxFormsIosSetup.cs
@@ -2,22 +2,22 @@
 // The .NET Foundation licenses this file to you under the MS-PL license.
 // See the LICENSE file in the project root for more information.
 
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
 using MvvmCross.Binding;
 using MvvmCross.Binding.Bindings.Target.Construction;
-using MvvmCross.Forms.Presenters;
-using MvvmCross.Localization;
-using System.Collections.Generic;
-using System.Reflection;
 using MvvmCross.Forms.Core;
 using MvvmCross.Forms.Platforms.Ios.Bindings;
+using MvvmCross.Forms.Platforms.Ios.Presenters;
+using MvvmCross.Forms.Presenters;
+using MvvmCross.IoC;
+using MvvmCross.Localization;
 using MvvmCross.Platforms.Ios.Core;
 using MvvmCross.Platforms.Ios.Presenters;
 using MvvmCross.Plugin;
 using MvvmCross.ViewModels;
-using UIKit;
-using MvvmCross.Forms.Platforms.Ios.Presenters;
 using Xamarin.Forms;
-using System.Linq;
 
 namespace MvvmCross.Forms.Platforms.Ios.Core
 {
@@ -36,10 +36,11 @@ namespace MvvmCross.Forms.Platforms.Ios.Core
             return _viewAssemblies;
         }
 
-        protected override void InitializeIoC()
+        protected override IMvxIoCProvider InitializeIoC()
         {
-            base.InitializeIoC();
-            Mvx.IoCProvider.RegisterSingleton<IMvxFormsSetup>(this);
+            var provider = base.InitializeIoC();
+            provider.RegisterSingleton<IMvxFormsSetup>(this);
+            return provider;
         }
 
         protected override void InitializeApp(IMvxPluginManager pluginManager, IMvxApplication app)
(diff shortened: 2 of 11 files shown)


nickrandolph:
@martijn00 @Cheesebaron this PR shouldn't change actual startup behaviour. It does rely on the IoCProvider more to create instances of objects but this should make it easier for developers to override implementation of entities created by setup (ie they just register their own types)

nickrandolph:
@Cheesebaron / @martijn00 is there a way to get AppVeyor to run the build again without me committing a change. I think it's refused to build because the PR was draft. It's not draft any more and I think it just needs to run the build again.

Cheesebaron:
@nickrandolph yes, you can go to AppVeyor and trigger it in there. You should already be a part of the AppVeyor team, contact me if you have issues or further questions. I've triggered a new build.

As for this PR. It seems good to me. I've run a couple of different projects with the changes in and checked that stuff is registered in the IoC Container. Doesn't seem to break anything else than some minor naming changes. I'd guess most people wont be affected.

Cheesebaron:
```
Platforms\Mac\Core\MvxFormsMacSetup.cs(39,33): error CS0508: 'MvxFormsMacSetup.InitializeIoC()': return type must be 'IMvxIoCProvider' to match overridden member 'MvxSetup.InitializeIoC()' [C:\projects\mvvmcross\MvvmCross.Forms\MvvmCross.Forms.csproj]

Platforms\Ios\Core\MvxFormsIosSetup.cs(39,33): error CS0508: 'MvxFormsIosSetup.InitializeIoC()': return type must be 'IMvxIoCProvider' to match overridden member 'MvxSetup.InitializeIoC()' [C:\projects\mvvmcross\MvvmCross.Forms\MvvmCross.Forms.csproj]

Platforms\Tizen\Core\MvxFormsTizenSetup.cs(38,33): error CS0508: 'MvxFormsTizenSetup.InitializeIoC()': return type must be 'IMvxIoCProvider' to match overridden member 'MvxSetup.InitializeIoC()' [C:\projects\mvvmcross\MvvmCross.Forms\MvvmCross.Forms.csproj]

Platforms\Uap\Core\MvxFormsWindowsSetup.cs(56,33): error CS0508: 'MvxFormsWindowsSetup.InitializeIoC()': return type must be 'IMvxIoCProvider' to match overridden member 'MvxSetup.InitializeIoC()' [C:\projects\mvvmcross\MvvmCross.Forms\MvvmCross.Forms.csproj]

Navigation\RoutingServiceTests.cs(41,13): error CS0120: An object reference is required for the non-static field, method, or property 'MvxNavigationService.LoadRoutes(IEnumerable<Assembly>)' [C:\projects\mvvmcross\UnitTests\MvvmCross.UnitTest\MvvmCross.UnitTest.csproj]

Platforms\Android\Core\MvxFormsAndroidSetup.cs(42,33): error CS0508: 'MvxFormsAndroidSetup.InitializeIoC()': return type must be 'IMvxIoCProvider' to match overridden member 'MvxSetup.InitializeIoC()' [C:\projects\mvvmcross\MvvmCross.Forms\MvvmCross.Forms.csproj]
```

Build fails with these errors. @nickrandolph can you please take a look?

softlion:
RegisterSetupDependencies  should not be static.

nickrandolph:
@Cheesebaron - apologies, missed your comment on this. I'll try to get these resolved asap

@softlion thanks for the feedback - RegisterSetupDependencies needs to be static as the intent is that developers can get away from having to override the Setup class just so that they can register their dependencies. There are opportunities later in the lifecycle for the registration of services and other dependencies but if you want to override some of the core dependencies, it needs to be one of the first operations that's done after the container is created. The point of the static method (and I agree, I'm usually not a big fan of global static methods like this) is that it can be registered from anywhere but the intent is that it's probably registered in the platform specific head project (or perhaps the UI project in the case of a XF project).

softlion:
@nickrandolph And where would the developer add these static dependencies on Android ? In a required  Application class ? I don't mean to be rude, but I don't see what's good in this. Also on iOS the appdelegate class is required and has a hidden non static setup class. 

I fear on Android that this will lead to developers to initialize this in the SplashScreen class, which won't work with other android statup shemes (service, alternate activity, ...).

nickrandolph:
@softlion that's again for the feedback - don't worry I don't take offence to your feedback.

The static method is designed more of a "get out of jail" option, rather than something I envisage every developer using. Due to the way setup is run, I don't see a graceful alternative way to call back to the platform head projects in order to do override core services. As part of fixing the issues @Cheesebaron pointed out I'll take another look and see whether there is a better alternative. 

Feel free to offer alternatives if you see any.

nickrandolph:
@Cheesebaron  this has been rebased and I've fixed up the breaks, including the tests that were failing. Should be good to merge if you're still ok with the changes

Cheesebaron:
@nickrandolph will review this later today. Thanks!

Cheesebaron:
I've tested this now and seems to work nicely. I am impartial about that `RegisterSetupDependencies` static. Maybe it could be an event instead? There is nothing in `InitializePrimary` that prevents it from being an event, right? In either case it serves the same purpose?

softlion:
I love when you don't take into account my comments.

ElteHupkes:
FYI Since most types in `RegisterDefaultSetupDependencies` were previously constructed explicitly but are now constructed through reflection their constructors are going to be stripped out for people (like me) with full linking enabled. Should there be a core `LinkerPleaseInclude.cs` content file created so people know what to include? I can create a PR but I'm not sure what the correct approach is.

Cheesebaron:
We can't include content in NuGet packages. That is why LinkerPleaseInclude isn't automatically added to your project anymore when adding the MvvmCross NuGet.

softlion:
+ considering that reflection is much slower. Compare a mvvmcross app and a pure Xamarin Forms app. The Xamarin Forms app opens immediatly, as it uses nearly no reflection, where the mvvmcross app needs 1 second.

ElteHupkes:
> We can't include content in NuGet packages. That is why LinkerPleaseInclude isn't automatically added to your project anymore when adding the MvvmCross NuGet.

I know that, but if I'm not mistaken there are content files with reasonable defaults for this so at least people don't have to guess. I was just wondering what the general idea for those is so I can update those, because as far as I can tell from the diff that wasn't done as part of this issue.

@softlion The reflection thing is something I'm somewhat worried about too, but I haven't benchmarked this change so I don't know quite how much this matters.

softlion:
I'm just jealous of the startup time of a pure xamarin forms app ...

For full link, there is a solution which is to add [Preserve(Conditional=true)] in strategic places of mvvmcross files. And i agree with you analysis of course. I am too always using full link enable.

nickrandolph:
The point of the changes in this PR (and not sure why we're continuing the discussion on a merged PR since we need to move forward) is to make overriding parts of Mvx more consistent. 
The reflection concern is valid but actually I think the additional startup is more to do with the overhead of Mvx as a whole, rather than reflection specifically.
If you look at some of the registration overloads added to the IoC provider you'll see you can register a factory. When I worked on this I did some benchmarks and it seemed to me that the one time hit on reflection for the dozen or so classes that are registered is almost insignificant in the scheme of things. Switching to registering based on a factory had little or no benefit.
Anyhow, as always, if you want to suggest ways to improve perf, feel free to PR away :-)

Cheesebaron:
We are happy to discuss this. Please open a new issue where this can happen rather than a merged PR.

PR #3376, opened 2019-04-13
Android: remove dependency on SplashScreen. Fix startup issues.

softlion:
Remove dependency on SplashScreen on Android. 
Also remove startup lifecycle  from within the base activity.
Also fix all android lifecycle problems.
Also fix static android shortcuts that were not working.
Also fix the usage of AppCenter Distribute, where the first launch could crash the app after coming back from appcenter screen.

An Android app must have a SplashScreenActivity derived from MvxSplashScreenAppCompatActivity.
Also all Activities implements the startup code.

SplashScreenActivity can be of any kind. Especially it can have no layout at all, when its layout is being defined by it's style, to get the fastest possible startup time.

Yes (impact: low): the handling of activity startup is done inside the StartupLifecycleCallback class only, and no more on all the activities that derived from MvxActivity.

Yes (impact: moderate): the app must have an Application class, which results in all static variables and variables linked to the application class behind held in memory, saved and restored. If memory management is not carefully crafted, it will quickly lead to out of memory exceptions.

Android playground updated

- [X] All projects build
- [-] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contributing/mvvmcross-docs-style-guide))


nickrandolph:
Hey @softlion does this assume now that all Android applications will have an application class? I think the current implementation allows for the scenario where there is no application class, and thus the splash screen initiates the Mvx startup.

softlion:
Yes. In fact all apps does have a hidden application class anyway.

Also, it permits to use a real 0 code instant splash screen on Android like described [here](https://android.jlelse.eu/right-way-to-create-splash-screen-on-android-e7f1709ba154)

I use myself the Application class to add a global error handler to catch all crashes hidden (mostly by mvvmcross) in the startup codes.

nickrandolph:
Sorry I wasn't very clear with my previous comment - I meant that this would mean that all developers would have to explicitly include an application class in their Android project which is a change in behaviour for Mvx. I think this might make things simpler and definitely like the implementation you've added.
@martijn00 any thoughts on the impact of these changes?

MvvmCross.Android.Support/V7.AppCompat/StartupLifecycleCallback.cs
```diff
@@ -0,0 +1,262 @@
...
+    /// Mark all your root activities with this interface.
+    /// </summary>
+    /// <remarks>
+    /// A root activity shuts down the mvvmcross application to free resources when it is destroyed.
+    /// </remarks>
+    public interface IShutdownActivity
+    {
+    }
+
+    public class StartupLifecycleCallback : Java.Lang.Object, Application.IActivityLifecycleCallbacks, IMvxSetupMonitor, IMvxAndroidCurrentTopActivity
```
martijn00: Needs `Mvx` naming convention.

martijn00:
Wouldn't it be possible to detect the splashscreen based on the `[Activity]` attribute with `MainLauncher = true`? Same would go for Intents with Attribute i guess. Also doesn't this do some similair stuff as https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Platforms/Android/Presenters/MvxAndroidViewPresenter.cs#L76 ?

softlion:
> Wouldn't it be possible to detect the splashscreen based on the [Activity] attribute with MainLauncher = true

No. There is no more such "splash screen activity". Any activity / background service can be an entry point, like with push notifications which can start activities with intent filters. The mainlauncher is for when there is no such intent filter designating no other entry point.

MvvmCross.Android.Support/V7.AppCompat/StartupLifecycleCallback.cs
softlion: You mean prefix with Mvx ?

martijn00: Yes. This would count for any custom class.

softlion:
I've fixed the naming and included another improvement to IMvxAndroidCurrentTopActivity. I discovered in one of my project that the default IMvxAndroidCurrentTopActivity is not working in some situations involving the new Google and Facebook logins. And there is no point using another one as IMvxAndroidCurrentTopActivity is already doing the work.

softlion:
I found an issue with the playground and the way the tabs are inserted in the view. Also the presenter uses tasks to add tabs. This is not good in a real production app as you must be on the UI thread to add tabs. So I lightly modified the android view TabRootView of the playground so it calls ShowInitialViewModelsCommand after OnResume instead of OnCreate. Look at code on my fork branch android-nospashscreen.

Ok @martijn00  ?

Cheesebaron:
This PR is actually pretty cool. I think we should get it into MvvmCross 7.0.0. Sorry for dragging this out @softlion

softlion:
I have since used this architecture in 3 existing customer apps and tweaked this code to be more universal. A "problem" with that code is that is requires an application class, which have these consequences:
- the app is kept in memory, so it must correctly handles the background and restore events, and disable/reenable it's services, and correctly free all memory when closing a view, otherwise memory leaks stack.

I'll update the pr with the tweaks.

softlion:
PR updated.

softlion:
I will update the PR to include native handling of android shortcuts

DavidMarquezF:
Are there any plans to migrate this to android x and merge? It would help a lot since now we're forced to use an Activity for initialization (in our case, we follow the single activity pattern so we end up needing two activities: the splash and the one where all the navigation happens)

Hackmodford:
I branched off of develop and hand copied the changes @softlion made. MvvmCross doesn't build in macOS so I don't have a way to test if the changes carried over. Maybe you'd like to take a look at what I did and force push? I'd like to see this effort continue!

https://github.com/Hackmodford/MvvmCross/tree/android-no-splashscreen-copy

Hackmodford:
@softlion How does this work with no splash screen? I tried modifying the playground to not use a splash screen and I seem to get two instances of the root activity.

Cheesebaron:
@Hackmodford the approach here with having to create Application class is fine. However, coming in MAUI and I've also been thinking about it, we would need to provide an AppHost.

So this means that taking a dependency on Microsoft.Extensions.DependencyInjection and creating our own App host would very much obsolete this work and also obsolete how MvvmCross starts up and registers all its dependencies currently.

So while this code might work, I was thinking about changing everything. Just need to find some time to spend on this.

softlion:
The approach here does work perfectly in at least 2 commercial apps built in 2019. With the version of mvvmcross that was shipped at this time. But the release "speed" of mvvmcross (call it "innovation") was too slow for me, so i started using another framework. I thank you and wish you good luck !

Hackmodford:
@Cheesebaron Does the AppHost comment apply if we were building a native app instead of a MAUI (Forms) type app?

Paul-N:
Recently I've exploring some MAUI code and I think we can have similar IStartup approach in Mvx. What we need is dependency of these three libs: MAUI.Core, MAUI.Essentials and Microsoft.Maui.Graphics. If we will add them we can stay closer to modern .NET developing. Btw those libs can be built against 'old' pre-6 Xamarin targets and netStd.

Issue #3383, opened 2019-04-30
Mvx. namespace shortcut not working in VS2019 16.1 preview2 with D8 and aapt2 enabled


softlion:
## 🔙 Regression

The "Mvx" namespace shortcut is not working in android projects compiled in VS2019 preview in debug mode when  Dex compiler is set to D8, code shrinker is set to ProGuard and aapt2/multidex are checked (not sure if multidex is an issue here).

For example, this android xml layout code which is working using older xamarin versions does not work anymore. MvxGridView is not found.

```xml
<Mvx.MvxGridView
            android:layout_width="match_parent"
            android:layout_height="match_parent"
            app:MvxBind="ItemsSource Items;ItemClick ItemClickCommand"
 />
```

The cause is the property `ViewNamespaceAbbreviations` in MvxAndroidSetup which is using upper chars in the namespace. Using only lower chars fix this issue.

```csharp
        protected override IDictionary<string, string> ViewNamespaceAbbreviations
        {
            get
            {
                return new Dictionary<string, string>
                {
                    {"Mvx", "mvvmcross.platforms.android.binding.views"}    
                };
            }
        }
```

### Configuration

**Version:** 6.2.3
**Platform:** 
- [ ] :iphone: iOS
- [X] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

martijn00:
Looks like a valid fix. It should not break anyone, right?

It would make sense to also rename `Mvx` to lowercase because all of Android has this already. Problem is that it would break everybodies apps.

softlion:
You could create 2 properties, mvxBind and MvxBind ;)
Note that i don't know if it works without D8.

martijn00:
@softlion can you check my PR?

PR #3393, opened 2019-05-06
Use lowercase names on android

martijn00:

Fixes https://github.com/MvvmCross/MvvmCross/issues/3383

- [ ] All projects build

### DIFF
--- a/MvvmCross/Platforms/Android/Core/MvxAndroidSetup.cs
+++ b/MvvmCross/Platforms/Android/Core/MvxAndroidSetup.cs
@@ -231,8 +231,6 @@ namespace MvvmCross.Platforms.Android.Core
         protected virtual IDictionary<string, string> ViewNamespaceAbbreviations => new Dictionary<string, string>
         {
             { "Mvx", "mvvmcross.platforms.android.views"
-            },
-            { "Mvx", "mvvmcross.platforms.android.binding.views"
             }
         };
 
@@ -242,7 +240,7 @@ namespace MvvmCross.Platforms.Android.Core
             "Android.Widget",
             "Android.Webkit",
             "MvvmCross.Platforms.Android.Views",
-            "MvvmCross.Platforms.Android.Binding.Views",
+            "MvvmCross.Platforms.Android.Binding.Views"
         };
 
         protected virtual IEnumerable<Assembly> AndroidViewAssemblies => new List<Assembly>()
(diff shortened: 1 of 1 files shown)


MvvmCross/Platforms/Android/Core/MvxAndroidSetup.cs
```diff
@@ -230,7 +230,10 @@ protected virtual IEnumerable<Assembly> ValueConverterAssemblies
 
         protected virtual IDictionary<string, string> ViewNamespaceAbbreviations => new Dictionary<string, string>
         {
-            { "Mvx", "MvvmCross.Platforms.Android.Views" }
+            { "Mvx", "mvvmcross.platforms.android.views"
+            },
+            { "Mvx", "mvvmcross.platforms.android.binding.views"
```
softlion: Having 2 items with the same key (Mvx) won't work in a dictionary

martijn00: Aah true, but don't you need both namespaces? And wouldn't `mvvmcross.platforms.android.binding.views` be more important to keep?

nmilcoff: Changing the namespace to be `binding.views` will result in a breaking change, won't it? 🤔

martijn00: Well, i think it might not be working at all right now. I always use full names, maybe someone can give this a quick test?

MvvmCross/Platforms/Android/Core/MvxAndroidSetup.cs
```diff
@@ -231,8 +231,6 @@ protected virtual IEnumerable<Assembly> ValueConverterAssemblies
         protected virtual IDictionary<string, string> ViewNamespaceAbbreviations => new Dictionary<string, string>
         {
             { "Mvx", "mvvmcross.platforms.android.views"
-            },
-            { "Mvx", "mvvmcross.platforms.android.binding.views"
             }
         };
```
softlion: mvvmcross.platforms.android.binding.views is the original and should be the one to keep. Otherwise breaking changes occurs.

MvvmCross/Platforms/Android/Core/MvxAndroidSetup.cs
```diff
@@ -239,6 +240,7 @@ protected virtual IEnumerable<Assembly> ValueConverterAssemblies
             "Android.Widget",
             "Android.Webkit",
             "MvvmCross.Platforms.Android.Views",
```
nickrandolph: Should this line still exist - it was removed out of the ViewNamespaceAbbreviations dictionary?

Issue #3108, opened 2018-09-17
Documentation: .NET Standard Library instead of PCL?


markuspalme:
The documentation recommends PCLs for shared code across platforms:
https://www.mvvmcross.com/documentation/getting-started/portable-class-library-pcl

I read the page a couple of times and did not really understand what it's trying to convey. For example, the statement `it is possible to write the MvvM either as a PCL or as a shared project` does not make a lot of sense to me.

Should we a) get rid of the page as it is or b) replace it with a page that recommends .NET Standard libraries for shared code?

As I see it, the current documentation on the subject does recommend an obsolete technology and does not help any newcomers, quite the opposite.

nickrandolph:
We should definitely update this to recommend using .NET Standard library, ideally with multi-targeting now.
See https://nicksnettravels.builttoroam.com/post/2018/08/19/Getting-Started-with-XamarinForms-and-Multi-Targeting-in-Visual-Studio.aspx for setting up a csproj for multi-targeting

PR #3396, opened 2019-05-07
Issue 3108: Removed/updated documention: PCL -> .NET Standard

markuspalme:
This PR addresses #3108.

MvvmCross is no longer shipped as PCL librarires - thus `netstandard.md` is no longer relevant.

After updating `portable-class-library-pcl.md` to .net standard there was nothing left of value to the document. The stuff about the `Device.*` methods has nothing to do with PCL or .NET standard.

Finally, some plugins are no longer shipped as PCLs but as .NET standard libraries (verified using Fuget.org), the documentation has been updated here to reflect that.


martijn00:
Thanks! This is very helpful as we don't have much time for things like this! More docs contributions are always appreciated!

Issue #3198, opened 2018-11-12
Switch MvxColor to System.Drawing


Cheesebaron:
## 🚀 Feature Requests

Currently we use our own custom MvxColor implementation for cross platform colors. .NET Standard allows us to use built in System.Drawing namespace, which can eliminate our custom implementation.

This will require the following things:

- Adjust `ToNative()` implementation
- Move over extensions like FromHSV and other goodies
- Move over custom MvxColors if necessary

PR #3456, opened 2019-07-04
Replace MvxColor with System.Drawing.Color

Strifex:

This PR replaces the MvxColor class in favor of leveraging the Color struct in System.Drawing as referenced in #3198 .  This PR also updates associated unit tests, UI Converters, and purges the static MvxColors as they are all available in System.Drawing.Color.

The current behavior is that MvxColor must be used in viewmodels instead of the .net system library's Color struct in System.Drawing.

The new behavior is that System.Drawing.Color must be used instead of MvxColor and now there is less to maintain in mvvmcross itself.  Unfortunately, I couldn't find good equivalents in .net standard for parsing hex strings into color values so for now I moved that code from MvxColor into a new MvxHexParser class - if anyone has a better way or sees more opportunity for reduction here let me know.

Yes, MvxColor and MvxColors will no longer exist and System.Drawing must be used in viewmodels instead.

The unit tests appear to be passing - but for any additional testing just setup a viewmodel with a Color property or utilize the existing RGG, ARGB, etc. converters to check that you get the expected result.

NOTE: Apparently System.Drawing is referenced from a very odd version of OpenTK in Xamarin.iOS and Xamarin.Mac so I had to include that reference in the project file in order for System.Drawing to be able to be seen.  Supposedly this is "normal" for Xamarin.iOS/Mac platforms......

- [x] All projects build


Cheesebaron:
You removed a big test class which goes through some of the parsing of different values that you pass into what was previously MvxColor. I think it would be nice if we could add that back and use it on the new Hex Parser.

Strifex:
@Cheesebaron The MvxColorTests class that I removed only tested against the constructors of MvxColorTests and providing integer or byte values for argb, rgb, or individual rgb values.  None of that code ever tested the methods that previously were used to parse hex strings.  However, there is a unit tests class called MvxRgbaValueConverterTests in the plugins unit test that tests all of the UI converters which directly call the hex parsing logic.  So essentially, no test coverage has been lost and technically the hex parsing logic is being tested the same way it was before (being covered by the higher-level tests against the rgba value converters).

FinHorsley:
@Strifex Awesome, big thanks! We are using [Acr UserDialogs](https://github.com/aritchie/userdialogs/tree/master/src/Acr.UserDialogs), and all the config classes use `System.Drawing.Color`, which means we have to jump through some hoops at the moment to convert from `MvxColour`

Cheesebaron:
I haven't forgotten about this. Will review this soon.

Cheesebaron:
We may want to move the ToNative methods out of the Color plugin and into MvvmCross directly. Will create a separate issue for that.

Thanks for the PR!

MoeHamdan:
so I am wondering, TextColor used to work with MvxColor now since you have changed it to System.Drawing.Color it is not working, may you please fix MvxTextViewTextColorBinding ?

Cheesebaron:
@MoeHamdan please open a new issue, we will have a look at fixing it.

PR #3461, opened 2019-07-05
Breaking up logic for creating view lookup

nickrandolph:
Refactor - the motivation here is actually to support XF Shell where we want to be able to intercept and build routes based on the association between view and viewmodels

Two steps of startup process are in the same method

Break steps into two virtual methods so that they can be intercepted.

Only if the InitializeViewLookup method has been overridden - it the override method just needs to be extended to include the lookup dictionary.

Run the playground app

None

- [ X ] All projects build
- [ X ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contributing/mvvmcross-docs-style-guide))

### DIFF
--- a/MvvmCross/Core/MvxSetup.cs
+++ b/MvvmCross/Core/MvxSetup.cs
@@ -116,8 +116,10 @@ namespace MvvmCross.Core
             InitializeViewModelTypeFinder();
             SetupLog.Trace("Setup: ViewsContainer start");
             InitializeViewsContainer();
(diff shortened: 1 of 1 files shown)


UnreachableCode:
Is it a requirement to register all view lookups now? Since replacing InitializeViewLookup with InitializeLookupDictionary, I've had a lot of ViewModel KeyNotFoundExceptions, leading me to believe what was once done automatically, is now setup manually.

PR #3464, opened 2019-07-06
Improving TFM switching for developers

nickrandolph:
Feature - improve developer support for switching between TFMs.

Previously developers had to manually edit the Directory.build.props file to switch between All, Android, iOS, Uap based on what platform they wanted to develop for. There was only the .sln file which meant all projects were being loaded (or it would remember the developers last loaded state).

Added solution filters (.slnf) for All, Android, iOS and Uap to make it quicker to load solution for developing for each platform.
Updated Directory.build.props to have a TargetsToBuildDeveloperOverride property that can be set to the platform the developer wants to target (All, Android, iOS or Uap)
Added LaunchVS.XXX.bat files that can be used to switch the TargetsToBuildDeveloperOverride and then launch the appropriate .slnf file.

No

Try it out by invoking the .bat files
Note: I think the .bat will only work on Windows. Can someone on MacOS offer an alternative that would work for MacOS

N/A

- [ X ] All projects build
- [ N/A ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contributing/mvvmcross-docs-style-guide))


Cheesebaron:
Something like this should work with bash

Then you can run it with `sh script.sh tfm`

```bash
#!/bin/bash
tfm="$1"
sed_args="s/<TargetsToBuildDeveloperOverride>[a-zA-Z]*<\/TargetsToBuildDeveloperOverride>/<TargetsToBuildDeveloperOverride>$tfm<\/TargetsToBuildDeveloperOverride>/g"
sed -i -e $sed_args Directory.build.props
```

nickrandolph:
@Cheesebaron sure, but I'm not going to open bash just to launch VS. Feel free to modify/add the scripts to try to get it to work for both. At this stage since multi-targeting still doesn't work well on mac, I'm not sure of the value prop.....

Cheesebaron:
We could probably make automator shortcuts for running the bash scripts. For now I am fine with just having Windows support in this PR as multi-target still doesn't work on macOS, unless you use Rider.

I'll take a look at reviewing this soon 👍

Cheesebaron:
Not needed anymore. After switching to net6.0.

PR #3484, opened 2019-08-02
Prepare MvxIoCTest to allow other IoC providers

SamuelDebruyn:
* Move the references to the specific implementation of `MvxIoCProvider` to a single method
* Cleanup the class for consistency: make classes private where appropriate, use var where type is clear, remove the dependency on another fixture
* Use built-in xUnit mechanisms for setup and cleanup
* Add a separate unit test to verify if the IoCProvider is correctly registered
* Make the `CreateIoCProvider` virtual so that the tests can be used for other IoCProviders

- [x] All projects build
- [x] Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))
- [x] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contributing/mvvmcross-docs-style-guide))
- [x] Rebased onto current develop

Issue #2069, opened 2017-07-26
Improve the IoC


martijn00:
We should identify the current performance and the features supported in MvvmCross. After that we can work on improving it.

- [ ] Singleton (Mvx.RegisterSingleton)
- [ ] Transient
- [ ] Combined
- [ ] Complex
- [ ] Property
- [ ] Generics
- [ ] IEnumerable
- [ ] Conditional
- [ ] Child Container
- [ ] Asp Net Core
- [ ] Interception With Proxy
- [ ] Prepare And Register
- [ ] Prepare And Register And Simple Resolve

Reference:
http://www.palmmedia.de/blog/2011/8/30/ioc-container-benchmark-performance-comparison

willsb:
Somewhat related to #534

Is this a list of features that are supported? Because if it is, [Property Injection](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Platform/Platform/IoC/MvxPropertyInjector.cs) is supported.

martijn00:
This is a list of features we could support. I was planning to add the current Mvx implementation to them if it is already supported and then check them off. That way we can track which already are there.

nmilcoff:
We could also add an unregister feature, it is impossible to `Unregister` anything currently.

fedemkr:
I've researched a little about how other IoC libraries work and given that there are a lot of them that are already optimized, which would be the benefit on adding the features to the Mvx-IoC instead of using one of them e.g. https://github.com/seesharper/LightInject ?
By the way, Complex and Generics (if "generic dependency" means that the object has Generics type parameters) are also supported by the current IoC.

fedemkr:
I'd like to be assigned to this issue. I'm adding Mvx's IoC to the benchmarks' project so that we can have a comparison to other IoCs and seeing what each feature exactly means.
About the Generics, it is not supported by the current IoC AFAIK, because it refers to a generic without type specification, e.g. IMyGenericInterface<>.

fedemkr:
I've forked the [IoCPerformance repository](https://github.com/danielpalme/IocPerformance) and added MvvmCross to it: [fedemkr/IoCPerformance](https://github.com/fedemkr/IocPerformance).

[...]

PR #3491, opened 2019-08-08
IocConstruct constructor lookup & constructor injection

SamuelDebruyn:
**Feature**: IocConstruct now tries harder to look for the best-matching constructor instead of the .FirstOrDefault() that it sometimes fell back to. It can now also do constructor injection mixed with provided arguments. Added 1 unit test for this specific scenario.

Note that I built this PR on top of #3484 so make sure to either merge or decline that PR first. The refactored unit tests for IoC were a lot easier to work with while I was working on this.

You can't do this now. It throws an exception saying that it couldn't find a matching constructor.

MvvmCross can now do `IocConstruct` with constructor injection and supplied arguments when you're using the overload with the array.

It used to throw an exception where it can now handle this scenario. IMHO, not really a breaking change.

¯\_(ツ)_/¯ Suggest more unit tests?

Maybe #2069 

- [x] All projects build

### DIFF
--- a/MvvmCross/IoC/MvxTypeExtensions.cs
+++ b/MvvmCross/IoC/MvxTypeExtensions.cs
@@ -247,13 +247,21 @@ namespace MvvmCross.IoC
             {
                 return constructors.OrderBy(c => c.GetParameters().Length).FirstOrDefault();
             }
-            
-            var names = arguments.Keys;
 
+            var unusedKeys = new List<string>(arguments.Keys);
+            
             foreach (var constructor in constructors)
             {
                 var parameters = constructor.GetParameters();
-                if (parameters.All(p => names.Contains(p.Name)))
+                foreach (var parameter in parameters)
+                {
+                    if (unusedKeys.Contains(parameter.Name) && parameter.ParameterType.IsInstanceOfType(arguments[parameter.Name]))
+                    {
+                        unusedKeys.Remove(parameter.Name);
+                    }
+                }
+
+                if (unusedKeys.Count == 0)
                 {
                     return constructor;
                 }
--- a/MvvmCross/IoC/MvxTypeExtensions.cs
+++ b/MvvmCross/IoC/MvxTypeExtensions.cs
@@ -239,5 +239,54 @@ namespace MvvmCross.IoC
 
             return Activator.CreateInstance(type);
         }
+        
+        public static ConstructorInfo FindApplicableConstructor(this Type type, IDictionary<string, object> arguments)
+        {
+            var constructors = type.GetConstructors();
+            if (arguments == null || arguments.Count == 0)
+            {
+                return constructors.FirstOrDefault();
+            }
+            
+            var names = arguments.Keys;
+
+            foreach (var constructor in constructors)
+            {
+                var parameters = constructor.GetParameters();
+                if (parameters.All(p => names.Contains(p.Name)))
+                {
+                    return constructor;
+                }
+            }
+
+            return constructors.FirstOrDefault();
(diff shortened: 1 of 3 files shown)


MvvmCross/IoC/MvxTypeExtensions.cs
```diff
@@ -239,5 +239,54 @@ public static object CreateDefault(this Type type)
 
             return Activator.CreateInstance(type);
         }
+        
+        public static ConstructorInfo FindApplicableConstructor(this Type type, IDictionary<string, object> arguments)
+        {
+            var constructors = type.GetConstructors();
+            if (arguments == null || arguments.Count == 0)
+            {
+                return constructors.FirstOrDefault();
```
nickrandolph: This assumes the parameter-less constructor is always first - is this always true??

MvvmCross/IoC/MvxTypeExtensions.cs
```diff
@@ -239,5 +239,54 @@ public static object CreateDefault(this Type type)
...
+            }
+            
+            var names = arguments.Keys;
+
+            foreach (var constructor in constructors)
+            {
+                var parameters = constructor.GetParameters();
```
nickrandolph: Even if we have the names of the arguments, we should be matching on type of arguments first

MvvmCross/IoC/MvxTypeExtensions.cs
```diff
@@ -239,5 +239,54 @@ public static object CreateDefault(this Type type)
...
+        }
+        
+        public static ConstructorInfo FindApplicableConstructor(this Type type, object[] arguments)
+        {
+            var constructors = type.GetConstructors();
+            
+            foreach (var constructor in constructors)
+            {
+                var parameterTypes = constructor.GetParameters().Select(p => p.ParameterType);
+                var unusedArguments = arguments.ToList();
```
nickrandolph: We should check here to see if number of arguments matches the number of parameters to the constructor. However, also need to consider optional parameters

SamuelDebruyn: That's what I wanted to change. Before, `IocConstruct` could (in some cases, not all) only inject the arguments that were supplied. But why would you use it then? With these changes, you can have some params supplied by args and some others injected from registrations. So it's possible to have more parameters than arguments. What this checks is that all the supplied arguments can be used in the constructor, otherwise you probably meant to use a different constructor.

MvvmCross/IoC/MvxTypeExtensions.cs
SamuelDebruyn: No, I should change this to null.

MvvmCross/IoC/MvxTypeExtensions.cs
SamuelDebruyn: Alright, fixed that

MvvmCross/IoC/MvxTypeExtensions.cs
SamuelDebruyn: I was wrong, changed this to the constructor with the least amount of parameters

SamuelDebruyn:
I can't really see what caused the build to fail...

Cheesebaron:
Could I get you to rebase on develop please?

```
git checkout feature/ioc-construct-mixed
git rebase origin/master
```

Where `origin` is mvvmcross upstream.

SamuelDebruyn:
Hi @nickrandolph, anything else I'd have to change? I have some spare time on my hands in the coming week :)

PR #3510, opened 2019-08-27
Add more Android Target bindings

Cheesebaron:
Feature

Added target bindings for

- VideoView Uri
- WebView Uri
- WebView Html Data
- View Focus Change

- [ ] All projects build

### DIFF
--- a/MvvmCross/Platforms/Android/Binding/MvxAndroidBindingBuilder.cs
+++ b/MvvmCross/Platforms/Android/Binding/MvxAndroidBindingBuilder.cs
@@ -5,6 +5,7 @@
 using Android.Graphics;
 using Android.Preferences;
 using Android.Views;
+using Android.Webkit;
 using Android.Widget;
 using MvvmCross.Base;
 using MvvmCross.IoC;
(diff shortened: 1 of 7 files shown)


Cheesebaron:
Ehm... There was a build error...

kmiterror:
I want to add a comment on issue that some people might face because of this PR. Especially that it's time consuming to figure out.
Before this PR there was no Html binding for WebView in Android. 
We (and probably some more devs) created our own bindable property in a class which inherits from WebView. 
After updating to 6.4.0 we discovered that we have a problem with our webview. 
Apparently this is because of naming collision (we also named our property "Html"), and our property was overriden by this PRs implementation.

It is important to note that LoadData method is not enough in many cases, as you might need to provide base url. We use LoadDataWithBaseURL so we renamed our property to 'CustomHtml' so it doesn't collide with the MvvmCross implmentation.
Perhaps it's matter of discussion but maybe it is worth to add 'baseUrl' as a param when using Html binding?

Cheesebaron:
@kmiterror what do you mean with "add `baseUrl` as a param"? Where would you add that?

kmiterror:
I'm not sure entirely how to solve it.
First thing that comes to me is passing an object instead of string with 2 propeties one for Html and the other for BaseUrl. 
So when calling Html binding you would have to pass well known object, something like:
```
MvxWebViewHtmlParams {
   string Html {get;set;}
   string BaseUrl {get;set;}
}
```

Other possibility  might be to introduce property called BaseUrl and "if not empty" call LoadDataWithBaseURL method instead of LoadData

Basically this is an open question.
I rather wanted to point out a problem that we faced (naming collision) and how we solved it.

PR #3533, opened 2019-09-10
MvvmCross 6.4.0 blog post

Cheesebaron:
docs

- [ ] All projects build

### DIFF
--- a/docs/_posts/2019-9-9-mvvmcross-6.4.0-release.md
+++ b/docs/_posts/2019-9-9-mvvmcross-6.4.0-release.md
@@ -1,0 +1,137 @@
+---
+layout: post
+title: MvvmCross 6.4.0
+date:   2019-09-9 10:00:00 -0200
+categories: mvvmcross
+---
+
+A new MvvmCross version is available on [NuGet](https://www.nuget.org/packages/MvvmCross/6.4.0)! You can always find the latest [changelog in the root of the repository](https://github.com/MvvmCross/MvvmCross/blob/develop/CHANGELOG.md) to see what has changed between versions.
+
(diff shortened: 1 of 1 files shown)


Cheesebaron:
Wtf? This is not what I committed...

FinHorsley:
With the changes to using `System.Drawing.Color` in [PR#3456](https://github.com/MvvmCross/MvvmCross/pull/3456), do we have to use the plugin to convert to native colours (e.g. `UIColour`)?

We currently have the following converter, but I don't see how this will work given that `System.Drawing.Color` isn't available from the iOS project (might have missed something here though)
```
public static UIColor ToNativeColour(this MvxColor mvxColor)
{
     return UIColor.FromRGBA(mvxColor.R, mvxColor.G, mvxColor.B, mvxColor.A);
}
```

Cheesebaron:
@FinHorsley yes, that logic is still in the plugin. I think we should move it into the core project in the future.

FinHorsley:
Ok 🙂 Does this mean we have to use the plugin to convert to native?

Ideally I'd rather not have to use the plugin, and just have a custom extension method on `System.Drawing.Color` (if this is possible?)

Cheesebaron:
@FinHorsley depends on where you want to use it. But sure, you have the rudder and here is a captain hat captain hat

Joke aside. The converters are mainly for binding from ViewModel to View. We have a couple of different Binding Targets provided that will use `ToNative()` from the Color plugin.

Strifex:
@FinHorsley hey if you're having issues referencing System.Drawing.Color in iOS......see the notes in the PR I made, essentially System.Drawing is referenced from a very odd version of OpenTK in Xamarin.iOS and Xamarin.Mac so include that reference in the project file in order for System.Drawing to be able to be seen.

PR #3548, opened 2019-09-14
removing code with CachedFragment that causing problem since it is no…

MoeHamdan:
…t assigned in this case.

Bug fix

MvxCachingFragmentStatePagerAdapter causing null reference exception 
MvxCachingFragmentStatePagerAdapter will not cause null reference exception

No

- [ X] All projects build

### DIFF
--- a/MvvmCross.Android.Support/Fragment/MvxCachingFragmentStatePagerAdapter.cs
+++ b/MvvmCross.Android.Support/Fragment/MvxCachingFragmentStatePagerAdapter.cs
@@ -61,8 +61,6 @@ namespace MvvmCross.Droid.Support.V4
 
             mvxFragment.ViewModel = GetViewModel(fragmentInfo);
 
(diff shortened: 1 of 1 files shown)


MvvmCross.Android.Support/Fragment/MvxCachingFragmentStatePagerAdapter.cs
```diff
@@ -92,18 +90,5 @@ private static IMvxViewModel GetViewModel(MvxViewPagerFragmentInfo fragmentInfo)
 
             return viewModelLoader.LoadViewModel(fragmentInfo.Request, null);
         }
-        
-        private static Bundle GetArguments(MvxViewPagerFragmentInfo fragmentInfo)
```
Prin53: Unfortunately, I did not see your PR when I made my #3549. I would like to set `Arguments` as this allows to get `MvxViewModelRequest` inside the fragment.

PR #3587, opened 2019-10-03
Change Tab Aware Behavior

bbenetskyy:

If you need to pass data between tabs in runtime you should define your behavior and attach it.

Now you can use build-in behavior for `MvxTabbedPage`. At ViewModels For that in tabs which should share or get data from others in runtime, we should implement `IMvxChangeTabAware<TParameter>` and add `MvxChangeTabBehavior<TParameter>` with the same `TParameter` to your `MvxTabbedPage`(root page).

New Docs are attached to this PR

- [x] All projects build


Cheesebaron:
This is for Forms only?

bbenetskyy:
@Cheesebaron, in brief - yes. As far as I know, behaviors are mostly used in Xamarin.Forms projects. That's why it was an idea to simplify life just by using the attached behavior to standard `MvxTabbedPage`. 

Is there any easy way to use the attached behaviors for tabbed pages in Xamarin.Android, Xamarin.iOS and other Native projects?

Cheesebaron:
@bbenetskyy this is fine that it is a Forms only thing. I'll have a stab at reviewing it this weekend.

bbenetskyy:
@Cheesebaron thatnks for reviewing! 

The main idea of PR was about passing data not only on first navigation to tab but also at any time when the tab is opened.

I'm not familiar at the moment with `ChangePresentation` and `MvxPagePresentationHint` but suppose that it forces tab to be changed. But is any way to handle that when the user changing the tab by himself?

Issue #3588, opened 2019-10-03
Running Playground.Wpf crashes with Reflection.TypeLoadException


MartinZikmund:
### Old (and correct) behavior

`Playground.Wpf` app starts properly

### Current behavior

Trying to run `Playground.Wpf` crashes with `TypeLoadException` while trying to load plugins (`LoadPlugins`).

### Solution

This problem can be solved using approach mentioned [here](https://stackoverflow.com/a/7889272/732221), which ignores the inaccessible types, but loads the types which are accessible.

### Reproduction steps

Try to run `Playground.Wpf` for debugging

### Configuration

**Version:** 6.x

**Platform:** 
- [ ] :iphone: iOS
- [ ] :robot: Android
- [x] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

MartinZikmund:
If you agree with the mentioned solution, I will implement in a PR

MartinZikmund:
In fact, I found such an extension already exists in the repository, I will use that.

PR #3590, opened 2019-10-03
Getting plugin types safely

MartinZikmund:
### :sparkles: What kind of change does this PR introduce? 

Bug fix for #3588

When a Plugin assembly cannot be loaded from the target, the app startup crashes on `LoadPlugins` with `ReflectionTypeLoadException`.

Using the `ExceptionSafeGetTypes` extension method to ignore types which cannot be loaded.

No, instead of a hard crash, the app now breaks if the debugger is attached or ignores the error otherwise.

Test on the `Playground.Wpf` app where I noticed the problem due to `Plugin.Permissions` which is not available for WPF.

-

- [x] All projects build

### DIFF
--- a/MvvmCross/Core/MvxSetup.cs
+++ b/MvvmCross/Core/MvxSetup.cs
@@ -568,7 +568,7 @@ namespace MvvmCross.Core
                 _state = value;
                 FireStateChange(value);
             }
-        }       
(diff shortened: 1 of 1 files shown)


Cheesebaron:
Thanks!

PR #3599, opened 2019-10-09
Chore(readme): use https

imba-tjd:


Cheesebaron:
Thanks. Can you change the target to master instead?

imba-tjd:
Do I need to close this and open another PR?

global.json
```diff
@@ -1,4 +1,7 @@
 {
+  "sdk": {
+    "version": "3.0.100"
```
onovotny: There's no need to pin a version, only makes things harder

azure-pipelines.yml
```diff
@@ -24,6 +24,12 @@ variables:
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
 
 steps:
+- task: UseDotNet@2
+  displayName: 'Use .NET Core sdk'
+  inputs:
+    packageType: sdk
+    version: 3.0.100
```
onovotny: I'd use '3.0.x' as the version as that ensures you always use the latest patch release (fixes/security updates)

PR #3629, opened 2019-11-07
OnCreateViewHolder should receive the view type

DavidMarquezF:
Bug fix

OnCreateViewHolder should receive the view type, not the LayoutId. This is confusing when extending the MvxRecyclerAdapter.
OnCreateViewHolder will be called with the viewType.

No

[OnCreateViewHolder docs](https://developer.android.com/reference/android/support/v7/widget/RecyclerView.Adapter#oncreateviewholder)

- [ ] All projects build

### DIFF
--- a/MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
+++ b/MvvmCross.Android.Support/V7.RecyclerView/MvxRecyclerAdapter.cs
@@ -123,26 +123,27 @@ namespace MvvmCross.Droid.Support.V7.RecyclerView
         public override int GetItemViewType(int position)
         {
             var itemAtPosition = GetItem(position);
-            var viewTypeIndex = ItemTemplateSelector.GetItemViewType(itemAtPosition);
-            var viewType = ItemTemplateSelector.GetItemLayoutId(viewTypeIndex);
+            var viewType = ItemTemplateSelector.GetItemViewType(itemAtPosition);
+            
             return viewType;
         }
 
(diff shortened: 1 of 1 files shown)


Cheesebaron:
I am pretty sure this was done on purpose. @softlion can probably confirm. It was done in https://github.com/MvvmCross/MvvmCross/pull/3366

softlion:
Yes its not a bug and has a purpose.

The RecyclerView's GetItemViewType should return an index that is unique for a given view holder throughout your whole  activity. This means your whole application if you use only fragments (and one activity).

Why ? Because the RecyclerView has a common cache for the whole activity that can cache viewholders of any RecyclerView. And the cache key is the viewType (returned by GetItemViewType).
This means the cache key should be unique to the whole application to efficiently make use of this common cache.
And an easy way to make this unique is by using layout ids by default. Instead of using a layout index, which obvioulsy will always defeat the purpose of this cache (0, 1, 2) (0,1), resulting in lots of very late Dispose()  - mono does not even call Dispose on those viewholders, only their destructor is called after a very long time - and thus potentially a huge memory consumption leading to app crashes.

you can learn more of this little known feature there https://developer.android.com/reference/android/support/v7/widget/RecyclerView.RecycledViewPool

DavidMarquezF:
Nice to know!

Maybe this should be added to the docs?

Cheesebaron:
@DavidMarquezF you are welcome to help us adding it to the docs :)

PR #3670, opened 2019-12-21
Bug fix - Init() in ViewModel called twice on Android

### LINKED ISSUE #3394: Init() in ViewModel called twice on Android - Deep Linking
## 🐛 Init() Called Twice on Android

When we started to utilize the URI based navigation (deep linking), we noticed that within a ViewModel, if you specify an Init() method to receive your params, the method would get called twice on Android (once on iOS). We also noticed that even if you aren't linking to the page with a URI, Init() is called regardless with any overloads using the default values of your params. 

#### Example nav with ViewModel type
```C#
//From ViewModel1
NavigationService.Navigate<ViewModel2>();

//ViewModel2 class - called with default value for int (0), twice on Android, once on iOS
public void Init(int testid) { }
```

#### Example nav with URI navigation and query params
```C#
//From ViewModel1 
NavigationService.Navigate("mvvmcross-test://deeplink/?testid=123");

//ViewModel2 class - called with testid = 123, twice on Android, once on iOS
//with registered scheme: mvvmcross-test://deeplink/\?testid=(?<testid>[0-9]{1,3})$
public void Init(int testid) { }
```
 The above Init() method in a ViewModel is called every time the ViewModel is navigated to regardless of using a URI deep link or not. On Android, it is called twice. iOS is only called once, but every time as well regardless of having query param data to pass in.

When it's called through a deep link URI with query params, the data is passed correctly. When you're navigating to the viewmodel type directly (not URI) then the method is still invoked, but with just the default values of string and int being passed in.

You can add any number of Init() methods to the class with different overloads and they're all called regardless of using URI nav or ViewModel type nav.

### Expected behavior

1. Android should only call Init() once to match iOS's behavior.
2. Ideally the Init() overloads with params are only called during a URI deep link and not under normal ViewModel nav where it passes default params. I don't know the expected requirements around this though as it's not documented anywhere. Otherwise we end up having to check for default values every time, and if we receive a value that is of the default type (Ex: 0 for integer) that we want to use, then we can't add the ignore default(int) logic.

### Reproduction steps
I have a reproduction project below. I have only added logic to the Droid project to reproduce the double Init() calls. I also added two buttons, one using URI nav with params, one with viewmodel type nav to showcase that Init() with my overloads gets called regardless each time it's navigated to.
[Reproduction Project](https://github.com/justinwojo/XamEssentialsTest)

### Configuration
Visual Studio for Mac (2019 8.0.3)
Xamarin.iOS 12.8.0.0
Xamarin.Android 9.2.0.5
Any Android/iOS device and version. Not dependent upon any OS version from what I can tell.

**Version:** 6.2.3

**Platform:** 
- [?] :iphone: iOS (depending on Expected Behavior above)
- [X] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

pinkysek:
Bug fix.

Problem was in MvxFragmentExtensions in OnCreate() method [HERE](https://github.com/MvvmCross/MvvmCross/blob/3923c6df361c7aa56bcf9b9bd10bdead60e8bd7b/MvvmCross.Android.Support/Fragment/MvxFragmentExtensions.cs#L40) and [HERE](https://github.com/MvvmCross/MvvmCross/blob/3923c6df361c7aa56bcf9b9bd10bdead60e8bd7b/MvvmCross/Platforms/Android/Views/Fragments/MvxFragmentExtensions.cs#L38). There was called RunViewModelLifecycle() [HERE](https://github.com/MvvmCross/MvvmCross/blob/3923c6df361c7aa56bcf9b9bd10bdead60e8bd7b/MvvmCross/Platforms/Android/Views/MvxFragmentExtensions.cs#L69) method which call CallBundleMethods() with parameters "Init" and "ReloadState". But same call is on ViewModel side in MvxDefaultViewModelLocator [HERE](https://github.com/MvvmCross/MvvmCross/blob/3923c6df361c7aa56bcf9b9bd10bdead60e8bd7b/MvvmCross/ViewModels/MvxDefaultViewModelLocator.cs#L13) where method Load() and Reload() call RunViewModelLifecycle() from MvxDefaultViewModelLocator too and there is same implementation for calling CallBundleMethods() with parameters "Init" and "ReloadState". I think there is not necessary call RunViewModelLifecycle() on Fragment side when We have this implementation on ViewModel side. So this is probably why this problem was only on Android platform. 

On Android platform is Init() method from ViewModel called only once (correct behavior same like other platforms)

I don't think so.

I need someone who knows core implementation of Android platform and Navigation for review. Because maybe I'm wrong and We need call RunViewModelLifecycle() on Fragment side.

Fixes #3394
Probably this issue too (there was added call RunViewModelLifecycle() in OnCreate()) - https://github.com/MvvmCross/MvvmCross/pull/1165

- [x] All projects build
- [x]  Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))

### DIFF
--- a/Directory.build.props
+++ b/Directory.build.props
@@ -47,7 +47,7 @@
   </PropertyGroup>
 
   <ItemGroup Condition="'$(IsTestProject)' != 'true' and '$(SourceLinkEnabled)' != 'false' and '$(IsLibraryProject)' == 'true'">
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-19554-01" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
     <None Include="$(MSBuildThisFileDirectory)\LICENSE" Pack="true" PackagePath=""/>
   </ItemGroup>
 
--- a/MvvmCross.Analyzers/CodeAnalysis/MvvmCross.CodeAnalysis.csproj
+++ b/MvvmCross.Analyzers/CodeAnalysis/MvvmCross.CodeAnalysis.csproj
@@ -9,7 +9,7 @@ This package contains the 'Code Analysis' analyzers and code fixes for MvvmCross
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis" Version="3.3.1" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="3.4.0" />
   </ItemGroup>
 
   <ItemGroup>
--- a/MvvmCross.Android.Support/Fragment/MvxFragmentExtensions.cs
+++ b/MvvmCross.Android.Support/Fragment/MvxFragmentExtensions.cs
@@ -34,11 +34,6 @@ namespace MvvmCross.Droid.Support.V4
             {
                 // check if ViewModel instance was cached. If so, clear it and ignore previous instance
                 cache.GetAndClear(fragmentView.ViewModel.GetType(), fragmentView.UniqueImmutableCacheTag);
-
-                //TODO call MvxViewModelLoader.Reload when it's added in MvvmCross, tracked by #1165
-                //until then, we're going to re-run the viewmodel lifecycle here.
-                Platforms.Android.Views.MvxFragmentExtensions.RunViewModelLifecycle(fragmentView.ViewModel, bundle, request);
-
                 return;
             }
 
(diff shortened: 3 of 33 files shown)


pinkysek:
I don't know why build failed.

Cheesebaron:
@pinkysek not your fault. I've reported a bug with Xamarin, waiting for a response.

pinkysek:
@martijn00 I made mistake and commit another issue. I made revert but review from you was dismissed.

PR #3674, opened 2019-12-29
Remove obsolete platform services method

Strifex:

This PR removes the obsolete InitializePlatformServices method from MvxSetup's startup process.  It also updates the WpfSetup to utilize the proper InitializeFirstChance method instead to be consistent with all the other platforms.

Platforms can utilize the InitializePlatformServices method (marked obsolete) instead of InitializeFirstChance and the internals of the wpf setup class also utilize it.

Developers can no longer utilize the obsolete InitializePlatformServices method in MvxSetup and Wpf platform no longer relies on it.

Yes, if a developer overrode InitializePlatformServices in MvxSetup, even though it was obsolete - then they will have to change to InitializeFirstChance method instead.

Run the Wpf playground project and other mvx projects to see that no behavior changed.

N/A

- [X] All projects build


MvvmCross/Platforms/Wpf/Core/MvxWpfSetup.cs
```diff
@@ -97,10 +91,14 @@ protected override IMvxNameMapping CreateViewToViewModelNaming()
...
 
+        protected override void InitializeFirstChance()
+        {
+            RegisterPresenter();
+        }
+
         protected override void InitializeLastChance()
         {
             InitializeBindingBuilder();
-            base.InitializeLastChance();
```
nickrandolph: Why are we removing this call? It's good practice to call the base method, even though currently the base method doesn't do anything.

MvvmCross/Platforms/Wpf/Core/MvxWpfSetup.cs
```diff
@@ -97,10 +91,14 @@ protected override IMvxNameMapping CreateViewToViewModelNaming()
             return new MvxPostfixAwareViewToViewModelNameMapping("View", "Control");
         }
 
+        protected override void InitializeFirstChance()
+        {
```
nickrandolph: See my other comment - we should call the base method ie base.InitializeFirstChance()

Strifex: @nickrandolph my mistake - this was an oversight when I was removing some test code I must have removed the base calls as well - its fixed now.

MvvmCross/Platforms/Wpf/Core/MvxWpfSetup.cs
Strifex: @nickrandolph my mistake - this was an oversight when I was removing some test code I must have removed the base calls as well - its fixed now.

Strifex:
@Cheesebaron looks like build bot failed here but if builds fine on two different machines I have - problem with build machine potentially?

Error in the pipeline looks like the following:

C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets(1669,2): error XA0000: Unhandled exception: System.IO.IOException: The process cannot access the file 'D:\a\1\s\MvvmCross.Plugins\Location.Fused\obj\Release\monoandroid90\android\src\com\google\android\gms\common\R.java' because it is being used by another process.

Cheesebaron:
@Strifex should be fixed now if you either rebase on develop or merge develop into your branch.

Strifex:
@Cheesebaron updated!

PR #3675, opened 2019-12-29
Small Optimizations for Loading Plugins

Strifex:

This micro-optimizes retrieving bootstrapped assemblies by reducing duplicate array copies, and loading of plugins by removing LINQ in favor of traditional loops to reduce iterations. This helps when the project has a large amount of plugins.

The same, but slightly longer.

Savings on my test device varied but on average were around ~40ms on startup for a project with a small amount of plugins, would be higher for projects with more plugins / assemblies. Also, this does slightly reduce memory pressure on startup.

No

Use the android playground project before/after - you'll see a small change, easier to see if you add lots of plugin assemblies.

N/A

- [X] All projects build


Strifex:
@nickrandolph sure...

GetBootstrapOwningAssemblies:

- This method previously allocated a new List<T>() that wasn't needed as we never added anything beyond the ViewAssemblies to that list, therefore duplicating the list needlessly.
- This method also called .ToArray() at the end of its Distinct() call which is unnecessary as Distinct() also returns an Ienumberable just like the method signature - so the extra allocation of a new array was unnecessary.
- These two changes saved memory usage on startup and a very small amount of CPU, especially for projects with lots of assemblies.

LoadPlugins:

- This method previously utilized linq queries (selectMany and where) - previously the behavior was for those to excecute upon the start of the following foreach loop - which meant the query would execute - and then the result of the query would iterate through in order to call pluginManager.EnsurePluginLoaded(someType);  The new behavior changes this to simple for loops, which have less overhead then the LINQ queries, and it also removes the extra iteration in the previous behavior as it loads the plugin as its doing the filtering instead of doing the filtering and THEN iterating through the filtered results.
- Overall, the removal of the LINQ calls and the excess iteration saved as much as 35ms on the emulator, and some memory usage (assuming project had a lot of plugins).

nickrandolph:
GetBootstrapOwningAssemblies - yeh no issues with this change
LoadPlugins - my understanding of the linq is that it returns an IEnumerable and only iterates the collection when you're stepping through it (ie the for loop). The LINQ expression is much simpler to follow so would be preferable if there's no difference in behaviour. 
Are you able to provide the benchmark code that demonstrates the timing difference?

Strifex:
I didn't utilize a benchmark method or anything special but I did insert a stopwatch and started it before running the LoadPlugins method and then stopped it when it was done running and printed the time to the console.  I repeated this test 5 times before and after my changes and on average my changes saved about ~30ms on my emulator and ~35ms on my test device.

I understand your point about the deferred execution of the query only occurring as the foreach loop iterates, so I likely am not saving a second iteration of the final list in my changes.  However, what that tells us is that the overhead in doing .SelectMany() and .Where statements vs the foreach loops I have in my changes is somewhat significant.  If you look at the source code for the LINQ statements there is a lot more going on then the two loops I've written.  LINQ is a powerful tool, but its also one of the easiest ways to cause performance issues in hot paths in your application.  In my opinion, the most important parts of this framework for performance are startup, navigation, and bindings.  This change affects the startup - so if we can save even 10ms (which we look to be saving more than that) then that might be worth it considering how much startup overhead the framework has on platforms that don't utilize AOT.

For reference here is the approx source code of those two LINQ statements:

https://github.com/dotnet/runtime/blob/master/src/libraries/System.Linq/src/System/Linq/SelectMany.cs

https://github.com/dotnet/runtime/blob/master/src/libraries/System.Linq/src/System/Linq/Where.cs

Cheesebaron:
I think this is OK to merge. Things we can do to improve start-up time are nice. I would also argue that switching to plain foreach loops in this simple case doesn't destroy readability or maintainability.

PR #3726, opened 2020-02-28
update viewmodel-lifecycle docs

FinHorsley:

Docs update

Makes it more clear about the role of `Initialize()`. I'll try and add some more examples of binding the view to `InitializeTask` soon

I have removed some docs that have already been covered in the Navigation section of the docs, as it was making it harder to follow the viewmodel lifecycle aspects of this doc

No, just a docs update 🙂 

- [x] All projects build

### DIFF
--- a/docs/_documentation/fundamentals/viewmodel-lifecycle.md
+++ b/docs/_documentation/fundamentals/viewmodel-lifecycle.md
@@ -5,33 +5,22 @@ category: Fundamentals
 order: 3
 ---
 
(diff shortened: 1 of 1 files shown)


Double-Dude:
Hi there, I think the documentation is wrong.

It says "Initialize is called after the native navigation has take place.".

The Actual Behaviour is:
Initialize() is called before the Navigation takes place. The method is called immediately after Prepare() method is called. But MvxNavigationService does await the Task after Navigation takes place. But again, it means that the task is actually started immediately after ViewModel is created, Prepare() is called. 

This creates a big issue. Because if you put "Show Loading Spinner" inside the Initialize() method, the Spinner gets showed before the new ViewController is presented, then when ViewController is presented, it actually covered the Loading Spinner.

I've spent a long time digging into the source code and found out the reason.

PR #3745, opened 2020-03-31
Make build deterministic

martijn00:

- [ ] All projects build

### DIFF
--- a/Directory.build.props
+++ b/Directory.build.props
@@ -44,7 +44,7 @@
     <GeneratePackageOnBuild Condition=" '$(IsTestProject)' != 'true'">true</GeneratePackageOnBuild>
   </PropertyGroup>
 
-	<PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
+	<PropertyGroup Condition="'$(IsTestProject)' != 'true' and'$(TF_BUILD)' == 'true'">
 	  <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
 	  <Deterministic>true</Deterministic>
 	</PropertyGroup>
--- a/Directory.build.props
+++ b/Directory.build.props
@@ -42,9 +42,12 @@
 	
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <GeneratePackageOnBuild Condition=" '$(IsTestProject)' != 'true'">true</GeneratePackageOnBuild>
-    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
-    <Deterministic>false</Deterministic>
   </PropertyGroup>
+
+	<PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
+	  <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
+	  <Deterministic>true</Deterministic>
+	</PropertyGroup>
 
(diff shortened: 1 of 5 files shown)


Directory.build.props
```diff
@@ -30,17 +31,19 @@
...
     <DesignTimeBuild>false</DesignTimeBuild>
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <GeneratePackageOnBuild Condition=" '$(IsTestProject)' != 'true'">true</GeneratePackageOnBuild>
     <DebugType>portable</DebugType>
     <DebugSymbols>true</DebugSymbols>
+	
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <GeneratePackageOnBuild Condition=" '$(IsTestProject)' != 'true'">true</GeneratePackageOnBuild>
+    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
+    <Deterministic>false</Deterministic>
```
martijn00: This should be true on CI. Is there a way we can detect it in MSBuild?

Cheesebaron: What is this property and why is it needed?

Directory.build.props
```diff
@@ -4,24 +4,25 @@
     <PackageLicenseExpression>MS-PL</PackageLicenseExpression>
     <PackageProjectUrl>https://github.com/MvvmCross/MvvmCross</PackageProjectUrl>
     <PackageIconUrl>http://i.imgur.com/Baucn8c.png</PackageIconUrl>
```
Cheesebaron: We should probably change this to PackageIcon too. As PackageIconUrl is deprecated

Directory.build.props
Cheesebaron: I guess we can set it to true in the Cake build script if needed.

Cheesebaron: I guess this would also work:

```csharp
<PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
  <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
  <Deterministic>true</Deterministic>
</PropertyGroup>
```

martijn00: See https://github.com/clairernovotny/DeterministicBuilds

Directory.build.props
martijn00: PackageIcon is not working on older nuget clients. I've added the code for it, but disabled for now. Maybe we wait a bit longer to enable it?

Directory.build.props
```diff
@@ -30,19 +31,24 @@
...
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <GeneratePackageOnBuild Condition=" '$(IsTestProject)' != 'true'">true</GeneratePackageOnBuild>
     <DebugType>portable</DebugType>
     <DebugSymbols>true</DebugSymbols>
+	
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <GeneratePackageOnBuild Condition=" '$(IsTestProject)' != 'true'">true</GeneratePackageOnBuild>
   </PropertyGroup>
 
+	<PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
```
Cheesebaron, line 47: We need to exclude this for UnitTests

Cheesebaron: That is why the build fails

PR #3802, opened 2020-06-09
Fix issues after migration to AndroidX

alexshikov:
### :sparkles: What kind of change does this PR introduce? 
Bug fix

- RecyclerView does not load items by TemplateSelector
- Fragment<T> is missing one constructor

### :new: What is the new behavior ?
- Working as expected

No

Check RecyclerView example

No links

- [ ] All projects build
- [x ] Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))
- [x ] Rebased onto current develop

### DIFF
--- a/MvvmCross.DroidX/RecyclerView/MvxRecyclerAdapter.cs
+++ b/MvvmCross.DroidX/RecyclerView/MvxRecyclerAdapter.cs
@@ -142,7 +142,8 @@ namespace MvvmCross.DroidX.RecyclerView
 
         protected virtual View InflateViewForHolder(ViewGroup parent, int viewType, IMvxAndroidBindingContext bindingContext)
         {
-            return bindingContext.BindingInflate(viewType, parent, false);
+            var layoutId = ItemTemplateSelector.GetItemLayoutId (viewType);
+            return bindingContext.BindingInflate(layoutId, parent, false);
         }
 
         public override void OnBindViewHolder(AndroidX.RecyclerView.Widget.RecyclerView.ViewHolder holder, int position)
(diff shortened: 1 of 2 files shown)


Cheesebaron:
Thanks!

MvvmCross.DroidX/RecyclerView/MvxRecyclerAdapter.cs
```diff
@@ -142,7 +142,8 @@ public override AndroidX.RecyclerView.Widget.RecyclerView.ViewHolder OnCreateVie
 
         protected virtual View InflateViewForHolder(ViewGroup parent, int viewType, IMvxAndroidBindingContext bindingContext)
         {
-            return bindingContext.BindingInflate(viewType, parent, false);
+            var layoutId = ItemTemplateSelector.GetItemLayoutId (viewType);
```
pkhandelwal-cdl, line 145: @alexshikov @Cheesebaron 

I might be missing something but isn't the viewType parameter here the actual layout resource id? Why would the GetItemLayoutId method need to be called again? 

On line 127, the exact same method is called except this time passing in the index of the object in the backing data source, which makes sense - the template selector can return the layout id based on the index of the item in the list.

alexshikov: Because `viewType` is not the same as `layoutId`. Even though line 127 return `layoutId` you can override `GetItemViewType` and return your own data type (please see any Android examples with header/footer). And that means `InflateViewForHolder` should convert `viewType` to `layoutId` to pass it to `BindingInflate`.

Issue #3814, opened 2020-06-24
Use ValueTask in IMvxMainThreadAsyncDispatcher instead of Task


softlion:
## 🏗 Enhancement Proposal

"ValueTask" was introduced last year in netstandard to replace "Task" for performance reasons.
ValueTask is a struct, Task is a class. A struct is allocated and transmitted on the stack. A class is allocated on the heap. That makes a ValueTask very efficient by lowering the allocation time and the pressure on the garbage collector.

### Pitch

IMvxMainThreadAsyncDispatcher is used quite a lot. Replacing its return value from Task to ValueTask will not break any calling code. Only the implementations of IMvxMainThreadAsyncDispatcher should be slightly upgraded.

### Platforms affected (mark all that apply)
All

Strifex:
We need to be careful here or else we could hurt performance - its not so simple as "replace Task with ValueTask" for better performance.  I believe there is worse performance / overhead in the case of using a ValueTask for an operation that usually returns asynchronously.  The core reason to use ValueTask/ValueTask<T> over Task/Task<T> is for methods that "commonly" may return a synchronous result and in the case of the non-generic Task - do NOT already return Task.CompletedTask (which is already a cached Task under the hood).

I know one of the two methods in that class already returns Task.CompletedTask which would make the use of ValueTask irrelevant (at least according to this blog: https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/)  However, I think the second method may just be returning normally....something that needs to be looked at carefully and measured.

Cheesebaron:
If anyone looks into grabbing this task. I would like some benchmarks comparing the two, but also tests proving that it actually doesn't break anything.

Issue #3213, opened 2018-11-24
Blazor


andres-gimenez:
## 🚀 Feature Requests

Using MVVMCross with blazor.

### Contextualize the feature

Experiment with MVVMCross in Blazor. Start with IoC. And study the possibility of defining binding, converterts, commands and navigation.

### Describe the feature

Create an experimental branch to study all of this.

### Platforms affected (mark all that apply)
- [ ] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey:  @Xamarin.Forms
- [x] 🌍 Blazor

LRP-sgravel:
I was wondering if this will be possible given that Blazor/Razor Components will target .Net Core 3.0 as far as I understand.

Subscribing...

andrechi1:
MvvmCross is netstandard. Would Work without problem. What happens that Blazor is experimental. And you don't know their final result.

LRP-sgravel:
It is being integrated directly in .Net Core 3 under the Razor Components name.  https://docs.microsoft.com/en-us/aspnet/core/razor-components/?view=aspnetcore-3.0

nickrandolph:
We'd be happy to see mvvmcross working across Blazor (and wasm for Uno platform). At this stage we'd need someone from the community to drive this initiative.

mgochmuradov:
Here is a good post to start with:
[Separating code and presentation of Blazor pages](https://gunnarpeipman.com/blazor/blazor-code-behind/)

mgochmuradov:
Also, the Blazor page should fire `StateHasChanged()` on view models' `INotifyPropertyChanged` event.
See [SO answer](https://stackoverflow.com/a/49231672/1323341) for more details

mgochmuradov:
Blazor WebAssembly is going to be released soon (May 2020).
Are there any updates on MvvmCross for Blazor?

LRP-sgravel:
So I just tried to get Blazor WASM to work with MvvmCross and I encountered a problem during initialization.  I copied the stack trace below, but basically MVX relies on blocking monitors, with Blazor WASM does not support.

```
﻿System.Threading.SynchronizationLockException: Cannot wait on monitors on this runtime.
  at (wrapper managed-to-native) System.Threading.Monitor.Monitor_wait(object,int)
[...]

PR #3820, opened 2020-07-04
All asynchronous calls

andrechi1:

It's getting more asynchronous. And MvvmCross, you have. GetAwaiter(). GetResult();, which cause problems on some platforms. 
This is a proposal for change, so that all calls are asynchronous.
It's unfinished. But I send it, to have your feedback.

https://github.com/MvvmCross/MvvmCross/issues/3814#issue-644379944
https://github.com/MvvmCross/MvvmCross/issues/3213#issuecomment-652492077

- [x] All projects build


nickrandolph:
@andrechi1 this work was started by @Cheesebaron a while ago - see https://github.com/MvvmCross/MvvmCross/pull/3222
I know that we all want async throughout the framework but the issue is that most platforms have strict startup policies which don't play nicely with async. The solution is to force the async processing similar to what @Cheesebaron has in his PR.
This PR is massive - I suspect we'll need to carve off parts that can be reviewed and accepted, if possible

andrechi1:
For platforms that do not support asynchronous. They use something similar to

```C#
return new ValueTask();
```
```C#
return Task.CompletedTask;
```
```C#
return Task<bool>.FromResult(true);
```

Do it little by little. It's hard. Asynchronouss have to cover everything flowing.

I think it's better to keep working on this line. And merge when stable.

About #3222. Locks are controlled with ConfigureAwait(false); and ConfigureAwait(true); We have to be very careful. I have to keep working to make everything work out.

Cheesebaron:
This PR needs thorough testing on _all_ platforms to check that startup works correctly.

andrechi1:
Si, es un cambio muy grande. Yo estoy mirando MvxSingleton, y me está causando problemas. Quizás, deberíamos utilizar alguna librería de IoC.

nickrandolph:
@Cheesebaron how close was your PR from being ready? The only way I see this working is with us forcing the async pump, which is what was discussed around your PR a while ago. 
@andrechi1 simply forcing FromResult isn't an option as this will just end up deadlocking developer code that is designed to run asynchronously. The solution in the short term around the MvxSingleton is to just **not** do async stuff during starting.... which ironically is how all the platform startups seem to be designed. I know that this isn't **the** solution but attempting to force async into the startup logic is something we've struggled with for a long time.

andrechi1:
FromResult  should not produce an aa because it is a completed task. FromResul and new ValueTask, is the simple way to emulate an asynchronous interface in synchronous environment. In the future it can be improved.

The starter's working. I'm having mvxSetupSingleton the problem. It  no longer need to use TaskCompletionSource, and I want to simplify it.

An asynchronous start is important. Because it allows you to perform tasks in parallel. And reduce the start-up time.

Cheesebaron:
The async startup in MvvmCross doesn't help _that_ much, most of the time gained is swallowed by the overhead of the async methods anyways. What _really_ makes startup performance is profile AOT.

andrechi1:
Asynchronous methods are highly optimized. And devices have more and more cores. Already a ticket started 20% faster on Android.
What I want is for the app to start while it keeps doing things in the background.

andrechi1:
I close the PR, to send the changes step by step.

PR #3867, opened 2020-08-14
add wpf core support for certain plugins

mashbrno:
Network, ResourceLocalization and Visibility plugins can support WPF on .NET Core platform. The only change was to include WPF source codes in .netcore3.1 build. 

**Caveat**: I'm not sure, how to correctly reference WPF core libraries like `PresentationFramework` or `WindowsBase`

### DIFF
--- a/MvvmCross.Plugins/ResourceLoader/MvvmCross.Plugin.ResourceLoader.csproj
+++ b/MvvmCross.Plugins/ResourceLoader/MvvmCross.Plugin.ResourceLoader.csproj
@@ -4,7 +4,7 @@
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">uap10.0.16299;$(TargetFrameworks)</TargetFrameworks>
   </PropertyGroup>
 
-  <PropertyGroup Condition="$(TargetFramework.StartsWith('netcoreapp'))">
+  <PropertyGroup Condition="$(TargetFramework.StartsWith('netcoreapp')) and '$(OS)' == 'Windows_NT' ">
     <UseWPF>true</UseWPF>
   </PropertyGroup>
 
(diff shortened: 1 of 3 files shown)


Cheesebaron:
Yeah, I don't think that is the right way to reference the WPF netcore libraries.

I've seen CefSharp do something like in their netcore WPF projects. Not sure if it is possible to put conditions on these to only load when target is netcore:

```
<Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk.WindowsDesktop" />
<Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk.WindowsDesktop" />
```

martijn00:
See https://github.com/Baseflow/XamarinMediaManager/blob/develop/MediaManager.Forms/MediaManager.Forms.csproj#L13

mashbrno:
Thank you. That easy line works on my computer. I am eager to see wheter it will work on Azure Pipeline, but there is an issue with Sonar credentials.

MvvmCross.Plugins/ResourceLoader/MvvmCross.Plugin.ResourceLoader.csproj
```diff
@@ -4,6 +4,10 @@
     <TargetFrameworks Condition=" '$(OS)' == 'Windows_NT' ">uap10.0.16299;$(TargetFrameworks)</TargetFrameworks>
   </PropertyGroup>
 
+  <PropertyGroup Condition="$(TargetFramework.StartsWith('netcoreapp'))">
+    <UseWPF>true</UseWPF>
```
martijn00, line 8: `<UseWPF Condition=" '$(OS)' == 'Windows_NT' ">true</UseWPF>`

martijn00 (changes requested):

Issue #3825, opened 2020-07-07
Unable to build and run Playground.Mac sample


IntegerOverlord:
## 🐛 Bug Report

I am unable to build and run the Playground.Mac sample

It builds fine, but I get this error:

```
System.IO.FileNotFoundException: Could not load the file 'Plugin.Permissions'.
  at System.AppDomain.Load (System.Reflection.AssemblyName assemblyRef, System.Security.Policy.Evidence assemblySecurity) [0x00073] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/mcs/class/corlib/System/AppDomain.cs:758
  at System.AppDomain.Load (System.Reflection.AssemblyName assemblyRef) [0x00000] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/mcs/class/corlib/System/AppDomain.cs:724
  at at (wrapper remoting-invoke-with-check) System.AppDomain.Load(System.Reflection.AssemblyName)
  at System.Reflection.Assembly.Load (System.Reflection.AssemblyName assemblyRef) [0x00000] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/mcs/class/corlib/System.Reflection/Assembly.cs:440
  at ObjCRuntime.Runtime.CollectReferencedAssemblies (System.Collections.Generic.List`1[T] assemblies, System.Reflection.Assembly assembly) [0x00085] in /Library/Frameworks/Xamarin.Mac.framework/Versions/6.18.2.1/src/Xamarin.Mac/ObjCRuntime/Runtime.cs:544
  at ObjCRuntime.Runtime.CollectReferencedAssemblies (System.Collections.Generic.List`1[T] assemblies, System.Reflection.Assembly assembly) [0x00085] in /Library/Frameworks/Xamarin.Mac.framework/Versions/6.18.2.1/src/Xamarin.Mac/ObjCRuntime/Runtime.cs:544
  at ObjCRuntime.Runtime.RegisterEntryAssembly (System.Reflection.Assembly entry_assembly) [0x00032] in /Library/Frameworks/Xamarin.Mac.framework/Versions/6.18.2.1/src/Xamarin.Mac/ObjCRuntime/Runtime.cs:505
[...]

Issue #3579, opened 2019-09-28
Mac playground compiles but doesn't show any windows


ivmirx:
## 🔙 Regression

Hello! I'm starting with MvvmCross for Xamarin.Mac and discovered that its Playground doesn't work correctly.

### Current behavior

The app appears in the dock and the program hits a breakpoint in `ViewDidLoad()` in `RootView.cs`. But no window is displayed.

There are some errors: "Language file could not be loaded for <viewmodel_name>" and "MvxIoCResolveException: Failed to resolve type MvvmCross.Base.IMvxJsonConverter" but I don't think that they are relevant to the issue: [application_output.txt](https://github.com/MvvmCross/MvvmCross/files/3665863/application_output.txt)

### Reproduction steps

Run `Playground.Mac`

### Configuration

**Version:** 6.4.0 [latest develop], VS4Mac 8.3, macOS Mojave 10.14.6

**Platform:** 
- [ ] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [x] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

Cheesebaron:
A quick fix for this is in `RootView` remove the `IMvxOverridePresentationAttribute` interface.

ivmirx:
Thank you! Got it running and found a few other issues:

**Show Child**
Does not navigate to a child view but if pressed for the second time, logs the following:
`[INFO]  (MvvmCross.Logging.MvxLog) MvxAsyncCommand : execute ignored, already running.`

**Show Modal**
If pressed right after the app's start, then produces `NullReferenceException` (see the log below). But if "Show Child" was pressed first, then the modal shows but it seems that it's too large.

![image](https://user-images.githubusercontent.com/10554114/65894758-aaa88280-e380-11e9-86f7-61ebb11a38fe.png)

	System.NullReferenceException: Object reference not set to an instance of an object
	  at MvvmCross.Platforms.Mac.Presenters.MvxMacViewPresenter.ShowModalViewController (AppKit.NSViewController viewController, MvvmCross.Platforms.Mac.Presenters.Attributes.MvxModalPresentationAttribute attribute, MvvmCross.ViewModels.MvxViewModelRequest request) [0x00035] in /Users/Q/Sources/MvvmCross/MvvmCross/Platforms/Mac/Presenters/MvxMacViewPresenter.cs:229
[...]

PR #3959, opened 2020-10-20
Fix Mac playground and MvxMacViewPresenter

### LINKED ISSUE #2192: Occasionally in Playground.Mac the tab contents do not appear
image

Every so often the tab items end up being blank.

ivmirx:

Bugfixes in the Mac's presenter, along with one public API change. All was done in the process of repairing the Playground:

- "Show Child" does not crash anymore and returns back to the root
- "Show Modal" and "Show Sheet" show their views correctly on the second and following navigations
- "Show Window" works and opens a new window on each click
- "Show Tabs" works and its tabs can be switched

Mac navigation does not work both in Playground and new projects, most presentation modes are broken. [See here](https://github.com/MvvmCross/MvvmCross/issues/3579#issuecomment-536629734) for details. 

#### 1. `NSWindow MainWindow` was removed from `MvxApplicationDelegate`

Initially it created an excessive "ghost" window and produced the bug [described here](https://github.com/MvvmCross/MvvmCross/issues/3579#issuecomment-540247742). But I'd like to suggest its removal.

While `NSApplication.SharedApplication` offers a property `MainWindow`, this window is not "main" in the sense of ownership but rather the one focused right now. It might return `null` when the app is inactive or hidden ([source](https://developer.apple.com/documentation/appkit/nsapplication/1428723-mainwindow)). This is further complicated by another property `KeyWindow`, which may be equal to `MainWindow` but also can be its child window that currently receives key events. Or it can be `null` when no windows are receiving keyboard events. And, at the same time, neither of these two properties are guaranteed to be the frontmost window in the Z-order.

This volatility of the "main" window in macOS can lead to confusion when combined with a separate reference to it in `MvxApplicationDelegate`. It is always possible to obtain this window from `NSApplication.SharedApplication`.

So, since almost no code in MvvmCross uses this property **and** it isn't updated along the "real" `MainWindow` of the application **and** would be overall confusing in the multi-window environment of macOS, I think it's better to get rid of it. `readme.txt` and content files were updated accordingly.

#### 2. `MvxMacViewPresenter` checks if `MainWindow` is available before going for the last opened window

As described above, this window usually exists and usually is in the focus, so the last opened window should be the last resort.

#### 3. `MvxMacViewPresenter` doesn't match windows and attributes with `null` identifiers anymore

For example, "Show Sheet" was breaking because it was creating a new window with a `null` ID on the first display and then obtaining it instead of the main window when displayed for the second time (through `w.Identifier == attribute.WindowIdentifier`). With the new conditions, the four one-liners for window search were getting too long and repetitive, so I extracted them into the `FindPresentingWindow()` method.

#### 4. The condition for closing of modal and sheet windows was fixed

The condition in `Close()` for them was dependent on the parent window's `PresentedViewControllers` but because the loop is reversed, it was getting to the modal/sheet's windows first and then quitting on the `controller.ViewModel == viewModel` condition.

Initially, I thought that this bug was producing a memory leak because the `Windows` list wasn't reducing in size after closing modals/sheets, but there's a [much larger problem](https://github.com/xamarin/xamarin-macios/issues/5350) related to `NSApplication` calls, I will create a separate issue to discuss how to fix it.

#### 5. A note in the documentation what to do in the multi-window scenario

I've spent some time trying to improve the presenter and make it work well for apps with multiple windows that use the same root view model (one of the things I tried was setting a unique window `Identifier` by default). But that didn't solve the problems completely on the Presenter's level and at the same time complicated the life for people who just need to port a single-window mobile app to Mac. So, just placing a hint into the docs instead.

Yes

To get Playground.Mac running, the following is necessary:

- install Plugin.Permissions package into the project
- comment out the registration of `IMvxTextProvider` in [Playground.Core's App.cs](https://github.com/MvvmCross/MvvmCross/blob/develop/Projects/Playground/Playground.Core/App.cs#L28)

Fixes #2192, #3579, #3825, #3830 

- [x] All projects build

### DIFF
--- a/ContentFiles/macOS/AppDelegate.cs.pp
+++ b/ContentFiles/macOS/AppDelegate.cs.pp
@@ -8,7 +8,7 @@ namespace $rootnamespace$
     {
 		public override void DidFinishLaunching(NSNotification notification)
 		{
-			MvxMacSetupSingleton.EnsureSingletonAvailable(this, MainWindow).EnsureInitialized();
+			MvxMacSetupSingleton.EnsureSingletonAvailable(this).EnsureInitialized();
 			RunAppStart();
 		}
 	}
--- a/MvvmCross/Platforms/Mac/Core/IMvxMacSetup.cs
+++ b/MvvmCross/Platforms/Mac/Core/IMvxMacSetup.cs
@@ -10,8 +10,7 @@ namespace MvvmCross.Platforms.Mac.Core
 {
     public interface IMvxMacSetup : IMvxSetup
     {
-        void PlatformInitialize(IMvxApplicationDelegate applicationDelegate);        
-        void PlatformInitialize(IMvxApplicationDelegate applicationDelegate, NSWindow window);
+        void PlatformInitialize(IMvxApplicationDelegate applicationDelegate);
         void PlatformInitialize(IMvxApplicationDelegate applicationDelegate, IMvxMacViewPresenter presenter);
     }
 }
--- a/MvvmCross/Platforms/Mac/Core/MvxApplicationDelegate.cs
+++ b/MvvmCross/Platforms/Mac/Core/MvxApplicationDelegate.cs
@@ -11,21 +11,6 @@ namespace MvvmCross.Platforms.Mac.Core
 {
     public abstract class MvxApplicationDelegate : NSApplicationDelegate, IMvxApplicationDelegate
     {
-        private NSWindow window;
-        public virtual NSWindow MainWindow
-        {
-            get {
-                if (window == null)
-                {
-                    var style = NSWindowStyle.Closable | NSWindowStyle.Resizable | NSWindowStyle.Titled;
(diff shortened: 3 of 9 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/MvvmCross/MvvmCross?pullRequest=3959) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/MvvmCross/MvvmCross?pullRequest=3959) before we can accept your contribution.<br/><br/>:x: ivmirx [sign now](https://cla.dotnetfoundation.org/MvvmCross/MvvmCross?pullRequest=3959)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/MvvmCross/MvvmCross?pullRequest=3959) it.</sub>

Cheesebaron (approved):

Cheesebaron:
Thanks!

Issue #3114, opened 2018-09-23
The async loading in the Initialize method of the first viewmodel for UWP


yanxiaodi:
## 🔙 Regression

We already have some discussions regarding the async loading in the Initialize method of the viewmodel. According to the PR #2866 , the Android project should work, but UWP can't.

### Old (and correct) behavior

With the previous version, we can create a CustomMvxAppStart class to support an async loading in the Initialize method: https://nicksnettravels.builttoroam.com/post/2018/04/19/MvvmCross-Initialize-method-on-the-first-view-model.aspx

### Current behavior
Now the Android project doesn't need to create a new CustomMvxAppStart class in the Core project. But it should uniform the behaviour for all the platforms.

### Reproduction steps
Create a new solution including Xamarin.Android, Xamarin.UWP and Xamarin.Core. In the FirstViewModel, use some async codes like this:
```
public override async Task Initialize()
        {
            // Async initialization, YEY!
            await base.Initialize();
            await GetPostsAsync();
        }
```
In the App.cs in the Core project, register the FirstViewModel:
```
    public class App : MvxApplication
    {
        public override void Initialize()
        {
            CreatableTypes()
                .EndingWith("Service")
                .AsInterfaces()
                .RegisterAsLazySingleton();
            RegisterAppStart<PostListViewModel>();
        }
    }
```
The Android project works, but the UWP can't. I didn't test it for iOS and the other platforms. Thanks.

### Configuration

**Version:** 6.2

**Platform:** 
- [ ] :iphone: iOS
- [X ] :robot: Android
- [ ] :checkered_flag: WPF
- [X ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

nickrandolph:
@yanxiaodi I'm not sure I understand what you mean by this issue. There is no regression from what I can see - all platforms can take advantage of the async support for Initialize. However, Android is the only one that currently takes advantage of this. If you want to do async work, you can still specify a CustomMvxAppStart. Can you explain your issue more please

yanxiaodi:
I mean I don't need the CustomMvxAppStart for Android, but it is still necessary for UWP. There is only one Core project. Can I use a uniform way for all the platforms? Thanks.

nickrandolph:
My point is that this isn't a bug, it's just that we don't provide default support within UWP for async startup

yanxiaodi:
[...]

PR #3975, opened 2020-10-28
Bug fix - The async loading in the Initialize method of the first ViewModel.

pinkysek:
Bug fix maybe smaller refactoring. 

While using async loading in the Initialize method of the first ViewModel on UWP and iOS (maybe more except Android - Android platform is fine it using async StartUp) app remains stuck on start screen.

Main problem was in MvxAppStart in Start() method which is void and there was awaiter with result. https://github.com/MvvmCross/MvvmCross/blob/13eb483cd737a22b65e1e7d1b97e96b277b9a642/MvvmCross/ViewModels/MvxAppStart.cs#L28

Get result in void method is not necessary and this was probably main problem with stucked app on start. After removing GetResult() from this method Playground.UWP and Playground.iOS starts fine with async method in RootViewModel.https://github.com/MvvmCross/MvvmCross/blob/13eb483cd737a22b65e1e7d1b97e96b277b9a642/Projects/Playground/Playground.Core/ViewModels/RootViewModel.cs#L204

I don't think so.

None?

Fixes - https://github.com/MvvmCross/MvvmCross/issues/3114

- [x] All projects build


pinkysek:
Minor question how about make MvxAppStart with ConfigureAwait(false) or it's bad idea?

Cheesebaron:
This won't work on iOS. Will throw a NSInternalInconsistencyException, since startup didn't finish before Window.MakeKeyAndVisible() was called.

pinkysek:
> This won't work on iOS. Will throw a NSInternalInconsistencyException, since startup didn't finish before Window.MakeKeyAndVisible() was called.

Now you mean question about ConfigureAwait(false)?

Cheesebaron:
No. Not having `StartAsync` blocking

pinkysek:
> This won't work on iOS. Will throw a NSInternalInconsistencyException, since startup didn't finish before Window.MakeKeyAndVisible() was called.

@Cheesebaron  No this works. I did not change anything in RunAppStart method inside MvxApplicationDelegate. I made change in non-async Start() method where I removed GetResult() only. I tried this again and it works. 
 
 iOS throw NSInternalInconsistencyException only when you call StartAsync() method in RunAppStart method inside MvxApplicationDelegate. But my change is only inside MvxAppStart not in iOS platform.

 I don't know maybe I just described my commit incorrectly.

PR #4108, opened 2021-02-23
Ensure UICollectionView animations are on main thread

Hackmodford:
This was the cause of a lot of headache for me. It looks like the TableViewSource implements a check for main thread but the CollectionViewSource does not. Adding this ensures that collection view animations occur on the main thread.

Bug fix

If the ObservableCollection notifies changes on something other than the main thread, the collection view does not reload data.

When receiving a Collection changed notification, ensures that animations are run on the main thread.

No

I was binding to DynamicData's ReadOnlyObservableCollection instead of the MvxObservableCollection

- [ ] All projects build

### DIFF
--- a/MvvmCross/Platforms/Ios/Binding/Views/MvxCollectionViewSourceAnimated.cs
+++ b/MvvmCross/Platforms/Ios/Binding/Views/MvxCollectionViewSourceAnimated.cs
@@ -46,7 +46,7 @@ namespace MvvmCross.Platforms.Ios.Binding.Views
         {
             if (!NSThread.IsMain)
             {
-                InvokeOnMainThread(() => CollectionChangedOnCollectionChanged(sender, args));
+                BeginInvokeOnMainThread(() => CollectionChangedOnCollectionChanged(sender, args));
                 return;
             }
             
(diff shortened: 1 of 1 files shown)


Cheesebaron (approved):

MvvmCross/Platforms/Ios/Binding/Views/MvxCollectionViewSourceAnimated.cs
```diff
@@ -44,6 +44,12 @@ public MvxCollectionViewSourceAnimated(UICollectionView collectionView, NSString
 
         protected override void CollectionChangedOnCollectionChanged(object sender, NotifyCollectionChangedEventArgs args)
         {
+            if (!NSThread.IsMain)
+            {
+                InvokeOnMainThread(() => CollectionChangedOnCollectionChanged(sender, args));
```
Cheesebaron, line 49: Can we maybe change this to BeginInvokeOnMainThread instead?

Cheesebaron (changes requested):

MvvmCross/Platforms/Ios/Binding/Views/MvxCollectionViewSourceAnimated.cs
Hackmodford: No objection but curious what the difference is?

Cheesebaron: BeginInvokeOnMainThread, won't block, while InvokeOnMainThread will block until the action is done.

Hackmodford: Done. Good to know :)

Cheesebaron (approved):

PR #3665, opened 2019-12-19
Fix CurrentTopActivity returning null when App in background

Cheesebaron:
### :sparkles: What kind of change does this PR introduce? (Bug fix, feature, docs update...)
Bug fix

### :arrow_heading_down: What is the current behavior?
When you background the app. CurrentTopActivity will _always_ return `null`. This isn't really correct behavior.

### :new: What is the new behavior (if this is a feature change)?
Always return the last top Activity

### :boom: Does this PR introduce a breaking change?
No this fixes a lot of issues, if you accidentally try to navigate while the App is in background etc.

### :memo: Links to relevant issues/docs
Fixes #3455 

### :thinking: Checklist before submitting

- [x] All projects build
- [ ] Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))
- [ ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contributing/mvvmcross-docs-style-guide))
- [x] Rebased onto current develop

PR #4290, opened 2021-10-14
Fix Android Top Activity

### LINKED ISSUE #3455: MvxApplicationCallbacksCurrentTopActivity is lying about current Activity
## 🐛 Bug Report

MvxApplicationCallbacksCurrentTopActivity seems to not be telling the truth about which Activity is the current one.

It assumes that as soon as `OnStop` is called on Activity it is not the Top Activity anymore.

This is an issue if you for instance start a longer running task in a Command, and in the end of that command you try to navigate. However, in the mean time a user as backgrounded the App.
In such cases CurrentActivity will always be null in the Android presenter, breaking Fragment navigation when the App is in the background. Also other actions depending on getting a CurrentTopActivity are also broken.

### Expected behavior
CurrentActivity should be the last known Activity regardless of OnStop has been called.

### Reproduction steps

1. Start long running Command in Fragment A which at the end tries to navigate to Fragment B
2. Background the App before attempt to navigate to Activity B
3. When coming back to the App you are still on Fragment A

### Configuration

**Version:** 6.x

**Platform:** 
- [ ] :iphone: iOS
- [x] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

Cheesebaron:
Bug fix

I tried fixing the top activity handling in PR#3665. However, it didn't fix it entirely. The issue is in some entry points of the App, we use a different class to listen to lifecycle. So we end up not actually knowing which activity is the top one.

This changes the code to always depend on a `Application` where we register `IActivityLifecycleCallbacks`. So the structure is changed a bit of when an instance of `MvxCurrentTopActivity` is created.

This is now done as soon as possible, and there should only ever be the initial instance of it. 

It shouldn't.

Fixes #3455
Revisits #3665

- [x] All projects build

### DIFF
--- a/Projects/Playground/Playground.Droid/Resources/layout/SharedElementRootView.axml
+++ b/Projects/Playground/Playground.Droid/Resources/layout/SharedElementRootView.axml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<android.support.design.widget.CoordinatorLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.coordinatorlayout.widget.CoordinatorLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
     android:fitsSystemWindows="true">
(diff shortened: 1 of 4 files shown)


softlion:
I fixed that perfectly 5 years ago in one of my only PR for mvvmcross. It worked in all conditions - start with notification, background service, shortcut and correctly handles cases where the app is already running. 

And the PR was never accepted. 

That's one of the reason I left mvvmcross at that time. Good luck.

PR #4312, opened 2021-10-31
Remove broken TResult ViewModel and in navigation service

### LINKED ISSUE #4262: Remove broken Navigate methods that pass back data when calling Close
## 🐛 Bug Report

We have a way of awaiting data from a ViewModel you navigate to through the NavigationSerivce. This is inheritly broken, since there are many cases, such as tombstoning the App, and other platform specific navigation paradigms that are hard to shoehorn into this navigation pattern.

This interface should be removed because it is broken by design.

Example:

```csharp
var result = await NavigationService.Navigate<MyViewModel, SendingParams, ResultParams(mySendingParam);
```

in MyViewModel:

```csharp
await NavigationService.Close<ResultParams>(result);
```

The await for getting the result often breaks, especially when navigating more levels into the stack and when app is tombstoned.

### Configuration

**Version:** 6.x

**Platform:** 
- [x] :iphone: iOS
- [x] :robot: Android
- [x] :checkered_flag: WPF
- [x] :earth_americas: UWP
- [x] :apple: MacOS
- [x] :tv: tvOS
- [x] :monkey: Xamarin.Forms

Cheesebaron:
Bug fix

We have broken non-lifecycle aware Navigate methods which promise to return a result when Close is called in the ViewModel navigated to.

This is broken by design and only works in very simple cases, leading to many issues.

Removing them, we could add something back in the future, but it can't be done with promising a result in a task. It can only be done with a callback of some sort.

YES

Fixes #4262

- [x] All projects build


dbala2:
Is there an alternative ?

dbala2:
Actually I found your comment: https://github.com/MvvmCross/MvvmCross/issues/4553#issuecomment-1398435889

DimaOrdenov:
IMHO Best practice is to deprecate stuff, not deleting it like that

PR #4386, opened 2022-03-05
Switch base type of TargetType property of generic MvxAndroidTargetBinding

### LINKED ISSUE #4389: Target type of generic version of MvxTargetBinding<TTarget, TValue> is incorrect
### Discussed in https://github.com/MvvmCross/MvvmCross/discussions/4385

<div type='discussions-op-text'>

<sup>Originally posted by **ivmazurenko** March  3, 2022</sup>
https://github.com/MvvmCross/MvvmCross/blob/1c002b3ba5148e10822ddaaade1a0a5b9fb0adfa/MvvmCross/Binding/Bindings/Target/MvxTargetBinding.cs#L70

Please, take a look for this line. Maybe i am wrong, but this property is used only once, when default value will be created in `MvxFullBinding`:

https://github.com/MvvmCross/MvvmCross/blob/1c002b3ba5148e10822ddaaade1a0a5b9fb0adfa/MvvmCross/Binding/Bindings/MvxFullBinding.cs#L170

Thus, I think, this line should be:
``
public Type TargetType => typeof(TValue);
``

Because it is type of value that will be applyed for target.

Can anybody explain me, why it is done this way? </div>

ivmazurenko:

This pull request fixes base `TargetType` of base generic MvxTargetBinding class. I think, this is bug

MvxFullBinding tryes to set instance of `TTarget` type into target property of `TValue` type

MvxFullBinding sets instance of `TValue` into property of `TValue` type

Yes, derived types will be affected.

You can try to write new class (like following) and check, how it works. In my case it throws exception when fallback value is not presented and source returns "unset value".

```c#
    public sealed class BackgroundColorBinding : MvxAndroidTargetBinding<View, Color>
    {
        private readonly View _view;

        public BackgroundColorBinding(View view)
            : base(view)
        {
        }

        protected override void SetValueImpl(View target, Color value)
        {
            target.SetBackgroundColor(value.ToAndroidColor());
        }
    }
```

#4385

fixes #4389

- [x] All projects build

### DIFF
--- a/MvvmCross.Forms/Bindings/Target/MvxBindablePropertyTargetBinding.cs
+++ b/MvvmCross.Forms/Bindings/Target/MvxBindablePropertyTargetBinding.cs
@@ -36,7 +36,7 @@ namespace MvvmCross.Forms.Bindings.Target
             _propertyChangedSubscription = bindableObject.WeakSubscribe(OnBindableObjectPropertyChanged);
         }
 
-        public override Type TargetType => _actualPropertyType;
+        public override Type TargetValueType => _actualPropertyType;
 
         private MvxBindingMode _defaultBindingMode = MvxBindingMode.TwoWay;
         public override MvxBindingMode DefaultMode => _defaultBindingMode;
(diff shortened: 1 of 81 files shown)


entdark:
Maybe it's better to rename `TargetType` to `ValueType` or `TargetValueType` to avoid confusion?

Issue #3854, opened 2020-07-29
Refactoring the startup and asynchronously.


andres-gimenez:
## 🚀 Feature Requests

I've detected problems in the code and I'd like to fix them. Most related to asynchronous management. Asynchronouss appeared with net4.5 and has evolved a lot since then. And he has a better chance.

### Contextualize the feature

The goal is to improve the start. 

1. Make concurrency control more robust. 
2. Support for more platforms (e.g. webassambly with blazor).
3. Get as asynchronous an interface as possible.

### Describe the feature

ConfigureAwait(false) must be indicated in all asynchronous files.
https://medium.com/bynder-tech/c-why-you-should-use-configureawait-false-in-your-library-code-d7837dce3d7f
https://devblogs.microsoft.com/dotnet/configureawait-faq/

Refactorizar MvxSingleton, usando Lazy y AsynLazy. This allows for a more stable start. And to be able to create an asynchronous boot. For the processor cores to be used.

Minimize the use of Lock. And never about non-private objects.

Not use GetAwaiter().GetResult(); It usually produces hard locks At least in UAP.

In the common part to all platforms only use async, await, Task, and ValueTask. Each platform has its uniqueness with asynchronous programming. Using TaskCompletionSource is useful for converting a non-asynchronous interface to asynchronous. But about UAP, it's problematic.

Use ValueTask instead of Task. Improves performance.  #3814

### Platforms affected (mark all that apply)
- [x] :iphone: iOS
- [x] :robot: Android
- [x] :checkered_flag: WPF
- [x] :earth_americas: UWP
- [x] :apple: MacOS
- [x] :tv: tvOS
- [x] :monkey: Xamarin.Forms
- [x] :🌍: Blazor

Cheesebaron:
For `ConfigureAwait(false)` you write:

> ConfigureAwait(false) must be indicated in all asynchronous files

That is not correct. It is needed in cases where you are _not_ interested in returning to the capture context. This is not _always_ the case. So blindly adding it everywhere doesn't really solve anything.

As for `GetAwaiter().GetResult()`. It can be necesasary to use that if you are starting something from a method that is not async and where you cannot await and you actually need the result after. However, it can be dangerous especially if something else is trying to do stuff on the same SynchronizationContext we call `GetResult()` on.

MvxSingleton doesn't have any async code. It won't benefit from using anything lazy in here, it is just a singleton.

[...]

PR #4546, opened 2023-01-12
Move Setup to MvxAndroidApplication +semver:breaking

Cheesebaron:
Bug fix/feature

Today we try to ensure MvvmCross is initialized in every Activity/Service/BroadcastReceiver/Fragment. This is really unnecessary as we can simply rely on Android Application for our setup.

Android Application, will _always_ run before any Service or Activity is started. This is similar to iOS and other platforms.

`MvxSplashScreenActivity` has been renamed to `MvxStartActivity`.

MvvmCross is now only initialized in MvxAndroidApplications.

All consumers of MvvmCross on Android will _need_ to have a Application class looking something like this:

```csharp
[Application]
public class MainApplication : MvxAndroidApplication<Setup, App>
{
    public MainApplication(IntPtr javaReference, JniHandleOwnership transfer) : base(javaReference, transfer)
    {
    }
}
```

The major breaking change here is that if you opt to **not** have your `MainLauncher` Activity inherit from `MvxStartActivity` the registered start `ViewModel` in your `MvxApplication` class will be ignored!

If you have a `MvxStartActivity` the startup behavior will continue as it was previously and you can have special logic in your `MvxApplication` to conditionally show a specific `ViewModel` after the splash screen.

Yes!

Test various scenarios

- App with and without MvxStartActivity

- Set background limit or kill Activities when they are backgrounded using Developer Settings
  - Ensure App behaves correctly and MvvmCross is correctly initialized in when App is killed in background

Supersedes #3376 
#3854

- [x] All projects build


entdark:
Maybe it's better to get rid of `MvxSplashScreenActivity` completely? It delays the actual launch. And current proper ways for showing splash screen is either setting splash theme or using new Android `SplashScreen` API.

Cheesebaron:
> Maybe it's better to get rid of `MvxSplashScreenActivity` completely? It delays the actual launch. And current proper ways for showing splash screen is either setting splash theme or using new Android `SplashScreen` API.

There always needs to be a MainLauncher in an Android App. So with this PR the only thing having a Splash Screen is to navigate to IMvxAppStart's registered ViewModel.

Otherwise if you just have a plain MainLauncher Activity with a ViewModel, MvvmCross will load that VM for the Activity, then you can do the new Android 12 splash screen stuff without a Splash Screen Activity.

Issue #3584, opened 2019-09-30
Useless MvxViewModel behavior


iXamDev:
## 🏗 Enhancement Proposal

Hi!

I would like to talk about ViewModel Lifecycle. As we can see in https://www.mvvmcross.com/documentation/fundamentals/viewmodel-lifecycle
MvxViewModel has few methods releted to Android lifecycle: _SaveStateToBundle_, _ReloadFromBundle_.

In which usecase this methods should be execute?

I my opinion it's useless logic, that should be never used.
Although I have been using at Mvvmcross for about 3-4 years and have not once felt the need for them. But I use custom implementation of IMvxAndroidViewPresenter based on custom fragment backstack, which gives me opportunity to handle complex navigation scenario.
For a long time I doubted and thought that maybe I do not understand something.
And did not dare to ask you such a question.

But now with Android Q and Android jetpack we can see how Google implement own [ViewModel](https://developer.android.com/reference/android/arch/lifecycle/ViewModel.html) and [AndroidViewModel](https://developer.android.com/reference/android/arch/lifecycle/AndroidViewModel). And we may notice that there no such kludge behavior.

In my mind, the ViewModel remains in memory throughout the entire life cycle of the application (or until screen dismiss), the screen View elements (Activities/Fragments) can be recreated. When you recreate the them, it is enough to select the ViewModel that already exists for it from the cache. And you never need to save VM state and recreate VM itself. And google developer say the [same thing](https://www.youtube.com/watch?v=5qlIPTDE274).

### Pitch
Just remove all android specific logic from MvxViewModel.

### Platforms affected (mark all that apply)
- [ ] :iphone: iOS
- [X] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

softlion:
> I my opinion it's useless logic, that should be never used.

I assure you it happens.

> In my mind, the ViewModel remains in memory throughout the entire life cycle of the application

Yes because mvvmcross are caching them when a fragment is temporarily destroyed.
I assure you it happens. I've seen the behavior a lot of times in the past 6 month.

iXamDev:
> I assure you it happens.
[...]

PR #4552, opened 2023-01-17
Doc update: testing tombstoning and why it's important

kmiterror:
I'm shedding some light on tombstoning and why it is important. I'm also demonstrating how to test it on Android.

docs update

no details about how to test tombstoning

added description detailing why implementing tombstoning is important and how to test it

no

no

nothing

- [x] All projects build

### DIFF
--- a/docs/_documentation/fundamentals/viewmodel-lifecycle.md
+++ b/docs/_documentation/fundamentals/viewmodel-lifecycle.md
@@ -108,6 +108,14 @@ Each platform MvvmCross supports has a different way of handing low memory situa
 
 In order to cope with that, the framework provides you with a way to save your ViewModel's state and a way to restore it later, regardless what might occur with the Views.
 
+> Your app might crash if you won't implement this
+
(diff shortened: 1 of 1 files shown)


kmiterror:
> Sounds overengineered for testing. Have you tried enabling "Don't keep activities" in Developer options on Android?


Yes I have tried it and it doesn't work, meaning `ReloadFromBundle` is never called (only SaveStateToBundle is called)
With testing approach that I described both of them are called every time, and you can reliably test your app in the situation of state restoring.
On a side note, this might be a bug in MvvmCross because OnSaveInstanceState/OnCreate(bundle) on Activity level works properly with DKA 

Also there is another issue with Don't keep activities setting, you look at your app in a way almost no (if any) users sees it.
How to test it properly if you have a stack of activities and you want to test the whole flow?

Issue #3584 is an example of this functionality being not very well understood.
The point of this change is to spread awareness, and give reliable tools to test it.

entdark:
I see. I use single Activity with multiple Fragments so it worked well in my case and restored all the Fragments correctly.

PR #4577, opened 2023-03-24
Prevent null reference exception in MvxPickerViewModel when using a null entry in your list

Digifais:
This fix prevents a null reference exception when using a `null` entry in your list with `MvxPickerViewModel`.

Crash when having a `null` entry in your list with `MvxPickerViewModel`.

Doesn't crash and picker shows an empty entry.

No

/

/

- [x] All projects build

### DIFF
--- a/MvvmCross/Platforms/Ios/Binding/Views/MvxPickerViewModel.cs
+++ b/MvvmCross/Platforms/Ios/Binding/Views/MvxPickerViewModel.cs
@@ -105,7 +105,7 @@ namespace MvvmCross.Platforms.Ios.Binding.Views
 
         protected virtual string RowTitle(nint row, object item)
         {
-            return item.ToString();
+            return item?.ToString();
         }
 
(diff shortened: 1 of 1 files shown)


Cheesebaron:
What would the purpose of having a null item in the list be to begin with?

Digifais:
> What would the purpose of having a null item in the list be to begin with?

In our use case, we have a form that can be validated and have inputs that require to use a picker but are not always required to be filled in to go through to the next step, hence the null item so users can select an empty item if they selected one by accident to be able to clear their selection.

Cheesebaron:
Almost inclined to not merge this one. You can always inherit from MvxPickerViewModel and override the default behavior which does not allow nulls.

Digifais:
Which is indeed what I'm doing in our project right now. The OS supports this, so I just found it kind of odd that MvvmCross would prevent us from using nulls, although I also understand where you are coming from as well.

Issue #4371, opened 2022-02-18
Value Combiner registration


DavidMarquezF:
## 🚀 Feature Requests

There is no clear way to register custom ValueCombiners. I think that this should be added in the Setup class so that we can override it. Either this or improve the documentation by saying that the ValueCombiners are not registered with reflection by default like the ValueConverters and a way to do so.

### Platforms affected (mark all that apply)
- [x] :iphone: iOS
- [x] :robot: Android
- [x] :checkered_flag: WPF
- [x] :earth_americas: UWP
- [x] :apple: MacOS
- [x] :tv: tvOS
- [x] :monkey: Xamarin.Forms
- [x] :🌍: Blazor

Cheesebaron:
Duplicate of #327

DavidMarquezF:
Why is it a duplicate? There is no way to override a register function for value combiners. 

For now there is no clear way on how to register them.

Cheesebaron:
You can override `InitializeBindingBuilder` and provide your own binding builder where extra value combiners are added.

DavidMarquezF:
Yes, that is what I'm currently doing, but wouldn't it be more convenient if there was an overridable function such as `FillValueCombiners` directly like with value converters?

If this is something that hasn't been added for a reason, I think that adding in the documentation how to do it would help a lot.

Cheesebaron:
@snechaev what is the proposal exactly?

snechaev:
Proposal is to add the `FillValueCombiners(IMvxValueCombinerRegistry registry)` overload to the platform-specific `Setup` classes (`MvxIosSetup`, `MvxMacSetup`, etc.). This will give the user with exactly the same experience as the for converters (except the naming-convention-based autoregistration, which was considered problematic from a performance point of view). And with such an overload, the developer will be able to easily register the combiners, either manually one-by-one or with assembly-scan or whatever.

Cheesebaron:
@snechaev ok, sounds good to me.

PR #4615, opened 2023-05-29
Combiners registration via Setup.FillValueCombiners() override

snechaev:
Feature

There is no easy way to register `ValueCombiner`. The only way is to provide a custom implementation of `BindingBuilder`, which seems like overkill for such a task.

This PR adds the `FillValueCombiners(IMvxValueCombinerRegistry registry)` overload to the platform-specific Setup classes (`MvxIosSetup`, `MvxMacSetup`, etc.). This behaves the same way as the `void FillValueConverters(IMvxValueConverterRegistry registry)` for the converters. This gives the user with exactly the same experience as the for converters (except the naming-convention-based autoregistration, which was considered problematic from a performance point of view). With this overload, the developer able to easily register the combiners, either manually one-by-one or with assembly-scan or whatever.

No

Nothing special, but worth remembering that `name` in the registration process should be without suffixes and prefixes. For example, for the `MvxMySuperValueCombiner` the name `MySuper` should be used during registration.

https://github.com/MvvmCross/MvvmCross/issues/4371

- [x] All projects build

### DIFF
--- a/MvvmCross/Platforms/Android/Core/MvxAndroidSetup.cs
+++ b/MvvmCross/Platforms/Android/Core/MvxAndroidSetup.cs
@@ -9,6 +9,7 @@ using MvvmCross.Binding;
 using MvvmCross.Binding.Binders;
 using MvvmCross.Binding.BindingContext;
 using MvvmCross.Binding.Bindings.Target.Construction;
+using MvvmCross.Binding.Combiners;
 using MvvmCross.Converters;
 using MvvmCross.Core;
 using MvvmCross.Exceptions;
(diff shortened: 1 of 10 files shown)


snechaev:
The `dotnet format` check failed on files that have not changed in the current PR. I don't think that it is a good idea to fix the formatting in the files that are unrelated to the this PR and include such a changes in the PR.

Cheesebaron (approved):

Cheesebaron:
Build is failing:

```
D:\a\1\s\MvvmCross\Platforms\Wpf\Core\MvxWpfSetup.cs(139,51): error CS0246: The type or namespace name 'IMvxValueCombinerRegistry' could not be found (are you missing a using directive or an assembly reference?) [D:\a\1\s\MvvmCross.Wpf\MvvmCross.Platforms.Wpf.csproj
```

snechaev:
Sorry, I did not noticed that Windows projects are not built on Mac hosts at least with the simplified build process (`EnableWindowsTargeting=true`). The build is now fixed. No changes other than the missing `usings` have been introduced.

PR #4826, opened 2024-02-23
Enable nullable on target bindings on Android and iOS

Cheesebaron:
Feature

No nullable attributes on all target binding definitions

Adds nullable attributes
Also added DynamicallyAccessedMembers attribute where appropriate, this should help with code trimming in some cases.

A bunch of them accepted `object` while they actually need a concrete view type. So changed the ctor signature of some target bindings. If you were not misusing these, then this should have no change to you.

- [x] All projects build


Cheesebaron:
Maybe, but I am actually not 100% sure. I've already tried once before applying it to a bunch of the IoC code, but it ends up being almost all of it having to be marked with `RequiresUnreferencedCode` attribute. Haven't looked deeper into the Binding code, but I suspect, most of it would have either that or `DynamicallyAccessedMembers` since everything in there is heavily dependent on Reflection.

Perhaps a better way for a lot of this binding code would be code generation, but that is a huge task to plunge into.

PR #4652, opened 2023-07-20
Remove IMvxViewModelResult

Cheesebaron:
### :sparkles: What kind of change does this PR introduce? (Bug fix, feature, docs update...)
Enhancement

### :arrow_heading_down: What is the current behavior?
IMvxViewModelResult was left over even though most of the methods and classes were remove for its support

### :new: What is the new behavior (if this is a feature change)?
Removed the last remnants of it

### :boom: Does this PR introduce a breaking change?
Yes

### :memo: Links to relevant issues/docs
Fixes #4651

### :thinking: Checklist before submitting

- [x] All projects build
- [ ] Follows style guide lines ([code style guide](https://github.com/MvvmCross/MvvmCross#code-style-guidelines))
- [ ] Relevant documentation was updated ([docs style guide](https://www.mvvmcross.com/documentation/contributing/mvvmcross-docs-style-guide))
- [x] Rebased onto current develop

Issue #4651, opened 2023-07-20
Remove IMvxViewModel<TResult> and IMvxViewModel<in TParameter, TResult>


ErisApps:
## 🏗 Enhancement Proposal

Like the title states, I'd propose to remove `IMvxViewModel<TResult>` and `IMvxViewModel<in TParameter, TResult>`as they don't serve any value anymore.
https://github.com/MvvmCross/MvvmCross/blob/2bf57322efa3c9c5022f4cee65d8c3c89a1a7ba5/MvvmCross/ViewModels/IMvxViewModel.cs#L45-L54

### Pitch

Since MvvmCross 9, the MvxNavigationService got its ability removed to return objects back to the parent viewmodel.
However, it seems like these interfaces are still leftovers that don't really make sense anymore now that you can't return data in that way.

### Platforms affected (mark all that apply)

(Assuming all platforms as it is in the core layer of Mvx)
- [x] :iphone: iOS
- [x] :robot: Android
- [x] :checkered_flag: WPF
- [x] :earth_americas: UWP
- [x] :apple: MacOS
- [x] :tv: tvOS
- [x] :monkey: Xamarin.Forms

Issue #4553, opened 2023-01-18
Close<Result> has been removed from IMvxNavigationService


D-Diyare:
## 🔙 Regression

I've been using MvvmCross since version 7,  most I worked with was version 8, lately wanted to upgrade to version 9, what was strange for me was I noticed that I can't close the ViewModel with return of result like before, after I took a look at the source I saw only one (out of the two) versions of Close method remained and its : 

``` cs
Task<bool> Close(IMvxViewModel viewModel, CancellationToken cancellationToken = default);
```
however, the implementation still exists in MvxNavigationService

``` cs
public virtual async Task<bool> Close<TResult>(
            IMvxViewModelResult<TResult> viewModel, TResult result, CancellationToken cancellationToken = default)
```

### Old (and correct) behavior
Exists:
``` cs
public virtual async Task<bool> Close<TResult>(
            IMvxViewModelResult<TResult> viewModel, TResult result, CancellationToken cancellationToken = default)
```

### Current behavior
Not exists:
``` cs
public virtual async Task<bool> Close<TResult>(
            IMvxViewModelResult<TResult> viewModel, TResult result, CancellationToken cancellationToken = default)
```

### Reproduction steps
bring back the above method so we can't return the result from the closed ViewModel 

### Configuration

**Version:** 9.x

**Platform:** 
- [ ] :iphone: iOS
- [ ] :robot: Android
- [ ] :checkered_flag: WPF
- [ ] :earth_americas: UWP
- [ ] :apple: MacOS
- [ ] :tv: tvOS
- [ ] :monkey: Xamarin.Forms

Cheesebaron:
This is not a regression. This is done on purpose.

D-Diyare:
So what is the workaround for this currently? I cannot seem to find it in the documentation, the documentation doesn't cover upgrading to version 9, I understand you guys working so hard on this but when something changes at least it needs to be mentioned somewhere.
Best regards

Cheesebaron:
The workaround is to not use it. You have plenty of options to make something work. I.e. the Parameters for you ViewModel, you could have a callback for backwards navigation. Or you could have some kind of global state. Totally up to you how to tackle it.

dsmitchell:
> The workaround is to not use it. You have plenty of options to make something work. I.e. the Parameters for you ViewModel, you could have a callback for backwards navigation. Or you could have some kind of global state. Totally up to you how to tackle it.

[...]

PR #4848, opened 2024-04-06
Added new result awaiting and setting ViewModel navigation

entdark:
Adds new `IMvxResultAwaitingViewModel<TResult>` and `IMvxResultSettingViewModel<TResult>` interfaces and corresponding abstract classes to create a result navigation contract between the awaiting and setting the result ViewModels. Under the hood it uses a new `IMvxResultViewModelManager` that registers, unregisters and sets result to registered result awaiting ViewModels.

It's better to use predefined abstract classes `MvxResultAwaitingViewModel<TResult>`, `MvxMultiResultAwaitingViewModel<TResult1, TResult2>`, `MvxResultSettingViewModel<TResult>`, `MvxNavigationResultAwaitingViewModel<TResult>`, `MvxNavigationResultSettingViewModel<TResult>`, and their overloads with parameters. Because when implementing `IMvxResultAwaitingViewModel<TResult>` and `IMvxResultSettingViewModel<TResult>` manually it requires implementing ViewModels registering, saving, restoration, unregistering, and setting the result manually as well.

No navigation with awaiting the result.

Navigation with awaiting the result.

No.

0. Prepare an Android 13+ device or emulator since we need runtime Notification permissions.
1. Deploy the Playground.Droid app.
2. Start it **without** debugging since the app will be shut down later.
3. Click "SHOW CHILD" button - it will show a child `Fragment` on bottom
4. Minimize the app (not close) by clicking Home or Overview.
5. Go to the Playground.Droid app system settings.
6. Make sure Notifications permissions are granted.
7. Disable Notifications permissions.
8. Go back to the app.
9. The app will be restored with the child `Fragment` on bottom.
10. Click "CLOSE" in the child `Fragment`.
11. Check the logcat (via your preferred IDE): it should display
```
Info (26885) / RootViewModel: [Information] Result Playground.Core.Models.SampleModel from Playground.Core.ViewModels.ChildViewModel
```
12. We received the awaited result after the full app restoration (with shutting down).

You can test this with any dangerous permissions.
Developer option "Don't keep activities" works differently - it does not shut down the app. Though the feature also works in this case.

For other platforms it just works without restoration issues.

PRs: #4312 #4652 
Issues: #3342 #3980 #4230 #4262 #4553 #4651 

- [X] All projects build

### 🧠 Extra
I am open for discussing changes for the implementation.
Here are some ideas:
- move 3 new navigation extensions to the navigation service directly
- handle default implementations differently through default interface declaration
- rename those long interface and class names to shorter ones

### DIFF
--- a/MvvmCross/ViewModels/Result/IMvxResultAwaitingViewModel.cs
+++ b/MvvmCross/ViewModels/Result/IMvxResultAwaitingViewModel.cs
@@ -1,0 +1,32 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MS-PL license.
+// See the LICENSE file in the project root for more information.
+
+namespace MvvmCross.ViewModels.Result;
+
+/// <summary>
+/// Implementing this interface requires to manually register and unregister ViewModel to result:
+/// <code>
+/// protected override void ReloadFromBundle(IMvxBundle state)
+/// {
+///     base.ReloadFromBundle(state);
+///     this.ReloadAndRegisterToResult&lt;<typeparamref name="TResult"/>&gt;(state);
+/// }
+/// protected override void SaveStateToBundle(IMvxBundle bundle)
+/// {
+///     base.SaveStateToBundle(bundle);
+///     this.SaveRegisterToResult&lt;<typeparamref name="TResult"/>&gt;(bundle);
+/// }
+/// public override ViewDestroy(bool viewFinishing = true)
+/// {
+///     base.ViewDestroy();
+///     if (viewFinishing)
+///         this.UnregisterToResult&lt;<typeparamref name="TResult"/>&gt;();
+/// }
+/// </code>
+/// </summary>
+/// <typeparam name="TResult"></typeparam>
+public interface IMvxResultAwaitingViewModel<TResult> : IMvxViewModel
+{
+    bool ResultSet(IMvxResultSettingViewModel<TResult> viewModel, TResult result);
+}
--- a/MvvmCross/ViewModels/Result/MvxResultSettingViewModel.cs
+++ b/MvvmCross/ViewModels/Result/MvxResultSettingViewModel.cs
@@ -1,0 +1,18 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MS-PL license.
+// See the LICENSE file in the project root for more information.
+
+namespace MvvmCross.ViewModels.Result;
+
+public abstract class MvxResultSettingViewModel<TResult> : MvxViewModel, IMvxResultSettingViewModel<TResult>
+{
+    public virtual void SetResult(TResult result)
+    {
+        this.SetResult<TResult>(result);
+    }
+}
+
+public abstract class MvxResultSettingViewModel<TParameter, TResult> : MvxResultSettingViewModel<TResult>, IMvxViewModel<TParameter>
+{
+    public abstract void Prepare(TParameter parameter);
+}
--- a/MvvmCross/ViewModels/Result/MvxResultViewModelExtensions.cs
+++ b/MvvmCross/ViewModels/Result/MvxResultViewModelExtensions.cs
@@ -1,0 +1,54 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MS-PL license.
+// See the LICENSE file in the project root for more information.
+
+namespace MvvmCross.ViewModels.Result;
+
+public static class MvxResultViewModelExtensions
+{
+    public const string BundleRegisterKey = "__mvxResultVMRegisterKey";
+
+    public static void ReloadAndRegisterToResult<TResult>(this IMvxResultAwaitingViewModel<TResult> viewModel, IMvxBundle savedStateBundle)
+    {
+        if (Mvx.IoCProvider?.TryResolve<IMvxResultViewModelManager>(out var manager) == true &&
+            savedStateBundle?.Data?.TryGetValue(BundleRegisterKey, out string restoreRegisterStr) == true &&
+            bool.TryParse(restoreRegisterStr, out bool restoreRegister) && restoreRegister)
+        {
+            manager.RegisterToResult(viewModel);
+        }
+    }
+
+    public static void SaveRegisterToResult<TResult>(this IMvxResultAwaitingViewModel<TResult> viewModel, IMvxBundle savedStateBundle)
+    {
+        if (Mvx.IoCProvider?.TryResolve<IMvxResultViewModelManager>(out var manager) == true &&
+            manager.IsRegistered(viewModel) &&
+            savedStateBundle?.Data is { } data)
+        {
+            data[BundleRegisterKey] = true.ToString();
+        }
+    }
+
+    public static void RegisterToResult<TResult>(this IMvxResultAwaitingViewModel<TResult> viewModel)
+    {
+        if (Mvx.IoCProvider?.TryResolve<IMvxResultViewModelManager>(out var manager) == true)
+        {
+            manager.RegisterToResult(viewModel);
+        }
+    }
+
+    public static void UnregisterToResult<TResult>(this IMvxResultAwaitingViewModel<TResult> viewModel)
+    {
+        if (Mvx.IoCProvider?.TryResolve<IMvxResultViewModelManager>(out var manager) == true)
+        {
+            manager.UnregisterToResult(viewModel);
+        }
+    }
+
+    public static void SetResult<TResult>(this IMvxResultSettingViewModel<TResult> viewModel, TResult result)
+    {
+        if (Mvx.IoCProvider?.TryResolve<IMvxResultViewModelManager>(out var manager) == true)
+        {
+            manager.SetResult(viewModel, result);
+        }
+    }
+}
--- a/MvvmCross/ViewModels/Result/MvxResultViewModelManager.cs
+++ b/MvvmCross/ViewModels/Result/MvxResultViewModelManager.cs
@@ -1,0 +1,55 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MS-PL license.
+// See the LICENSE file in the project root for more information.
+
+namespace MvvmCross.ViewModels.Result;
+
+public class MvxResultViewModelManager : IMvxResultViewModelManager
+{
+    private readonly Dictionary<Type, HashSet<IMvxViewModel>> _registrations = new();
+
+    public void RegisterToResult<TResult>(IMvxResultAwaitingViewModel<TResult> viewModel)
+    {
+        if (!_registrations.TryGetValue(typeof(TResult), out var resultRegistrations))
+        {
+            resultRegistrations = new();
+            _registrations[typeof(TResult)] = resultRegistrations;
+        }
+
+        resultRegistrations.Add(viewModel);
+    }
+
+    public bool UnregisterToResult<TResult>(IMvxResultAwaitingViewModel<TResult> viewModel)
+    {
+        if (_registrations.TryGetValue(typeof(TResult), out var resultRegistrations) &&
+            resultRegistrations.Remove(viewModel))
+        {
+            return true;
+        }
+        return false;
+    }
+
+    public bool IsRegistered<TResult>(IMvxResultAwaitingViewModel<TResult> viewModel)
+    {
+        if (_registrations.TryGetValue(typeof(TResult), out var resultRegistrations) &&
+            resultRegistrations.Contains(viewModel))
+        {
+            return true;
+        }
+        return false;
+    }
+
+    public void SetResult<TResult>(IMvxResultSettingViewModel<TResult> viewModel, TResult result)
+    {
+        if (_registrations.TryGetValue(typeof(TResult), out var resultRegistrations))
+        {
+            foreach (var resultRegistration in resultRegistrations.Cast<IMvxResultAwaitingViewModel<TResult>>().ToArray())
+            {
+                if (resultRegistration.ResultSet(viewModel, result))
+                {
+                    resultRegistrations.Remove(resultRegistration);
+                }
+            }
+        }
+    }
+}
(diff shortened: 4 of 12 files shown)


Cheesebaron:
Thanks will have a look at reviewing this one soon.

picosmos:
@Cheesebaron Can you say when you think you'll be looking at this?

We are currently migrating a project and are considering whether to use the code from this PR or our own implementation from another project. (it's a similar strategy, but we would prefer to use framework features if possible. But unfortunately we can't wait that long anymore.)

entdark:
@picosmos
I think if you hurry then it's better to go with your own.
I am not pleased with the solution in a few places and also probably there will be review fixes.

Cheesebaron:
> @Cheesebaron Can you say when you think you'll be looking at this?
> 
> We are currently migrating a project and are considering whether to use the code from this PR or our own implementation from another project. (it's a similar strategy, but we would prefer to use framework features if possible. But unfortunately we can't wait that long anymore.) Thank you for your great work :)

Sorry, I won't rush a review just because someone needs it. This looks something that would be easy to add to your own presenter override or use an alternative. Even if it gets reviewed and merged in, there is no guarantee there is a release with this in soon either.

I hope you understand :)

picosmos:
> > @Cheesebaron Can you say when you think you'll be looking at this?
> > We are currently migrating a project and are considering whether to use the code from this PR or our own implementation from another project. (it's a similar strategy, but we would prefer to use framework features if possible. But unfortunately we can't wait that long anymore.) Thank you for your great work :)
> 
> Sorry, I won't rush a review just because someone needs it. This looks something that would be easy to add to your own presenter override or use an alternative. Even if it gets reviewed and merged in, there is no guarantee there is a release with this in soon either.
> 
> I hope you understand :)

Sure, that's the right way to do. I think we use our own implementation.

amirvenus:
> > > @Cheesebaron Can you say when you think you'll be looking at this?
> > > We are currently migrating a project and are considering whether to use the code from this PR or our own implementation from another project. (it's a similar strategy, but we would prefer to use framework features if possible. But unfortunately we can't wait that long anymore.) Thank you for your great work :)
> > 
> > 
> > Sorry, I won't rush a review just because someone needs it. This looks something that would be easy to add to your own presenter override or use an alternative. Even if it gets reviewed and merged in, there is no guarantee there is a release with this in soon either.
> > I hope you understand :)
> 
> Sure, that's the right way to do. I think we use our own implementation.

I think one approach would be to add a PR for a new Mvx demo app (similar to the Tip Calculator) using an implementation that wouldn't require changes to the Mvx code.

entdark:
@amirvenus
> I think one approach would be to add a PR for a new Mvx demo app (similar to the Tip Calculator) using an implementation that wouldn't require changes to the Mvx code.

Check #4553 - there are solutions in comments.

MvvmCross/ViewModels/Result/MvxResultViewModelManager.cs
```diff
@@ -0,0 +1,55 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MS-PL license.
+// See the LICENSE file in the project root for more information.
+
+namespace MvvmCross.ViewModels.Result;
+
+public class MvxResultViewModelManager : IMvxResultViewModelManager
+{
+    private readonly Dictionary<Type, HashSet<IMvxViewModel>> _registrations = new();
```
Cheesebaron, line 9: Are there any implications having strong references to ViewModels in this dict?

MvvmCross/ViewModels/Result/MvxResultViewModelManager.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    public void RegisterToResult<TResult>(IMvxResultAwaitingViewModel<TResult> viewModel)
+    {
+        if (!_registrations.TryGetValue(typeof(TResult), out var resultRegistrations))
```
Cheesebaron, line 13:
```suggestion
        if (!_registrations.TryGetValue(typeof(TResult), out HashSet<IMvxViewModel> resultRegistrations))
```

MvvmCross/ViewModels/Result/MvxResultViewModelManager.cs
```diff
@@ -0,0 +1,55 @@
...
+            resultRegistrations = new();
+            _registrations[typeof(TResult)] = resultRegistrations;
+        }
+
+        resultRegistrations.Add(viewModel);
+    }
+
+    public bool UnregisterToResult<TResult>(IMvxResultAwaitingViewModel<TResult> viewModel)
+    {
+        if (_registrations.TryGetValue(typeof(TResult), out var resultRegistrations) &&
```
Cheesebaron, line 24:
```suggestion
        if (_registrations.TryGetValue(typeof(TResult), out HashSet<IMvxViewModel> resultRegistrations) &&
```

MvvmCross/ViewModels/Result/MvxResultViewModelManager.cs
```diff
@@ -0,0 +1,55 @@
...
+            resultRegistrations.Remove(viewModel))
+        {
+            return true;
+        }
+        return false;
+    }
+
+    public bool IsRegistered<TResult>(IMvxResultAwaitingViewModel<TResult> viewModel)
+    {
+        if (_registrations.TryGetValue(typeof(TResult), out var resultRegistrations) &&
```
Cheesebaron, line 34:
```suggestion
        if (_registrations.TryGetValue(typeof(TResult), out HashSet<IMvxViewModel> resultRegistrations) &&
```

Cheesebaron (commented):
Hey! Sorry for the long wait reviewing the code. Here's a few thoughts about the solution. Let me know what you think.

entdark:
@Cheesebaron hi, I saw the review comments, thanks. I will reply soon.

MvvmCross/ViewModels/Result/MvxResultViewModelManager.cs
entdark: We need to find which VMs to notify by the result type in `MvxResultViewModelManager.SetResult`. Notice that `MvxResultViewModelManager.SetResult` takes `IMvxResultSettingViewModel` parameter while other methods take `IMvxResultAwaitingViewModel`. `IMvxResultSettingViewModel` doesn't know anything about the awaiter, and the only thing in common for them is `TResult` - that's why we need to store VMs associated to the `TResult` type.

MvvmCross/ViewModels/Result/MvxResultSettingViewModel.cs
```diff
@@ -0,0 +1,18 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MS-PL license.
+// See the LICENSE file in the project root for more information.
+
+namespace MvvmCross.ViewModels.Result;
```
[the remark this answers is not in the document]
entdark: Yeah, for convenience. And yes, the default implementation could be moved to the interface and leave `MvxResultSettingViewModel` just inherited from `IMvxResultSettingViewModel` but empty inside.

MvvmCross/ViewModels/Result/IMvxResultAwaitingViewModel.cs
```diff
@@ -0,0 +1,32 @@
...
+/// public override ViewDestroy(bool viewFinishing = true)
+/// {
+///     base.ViewDestroy();
+///     if (viewFinishing)
+///         this.UnregisterToResult&lt;<typeparamref name="TResult"/>&gt;();
+/// }
+/// </code>
+/// </summary>
+/// <typeparam name="TResult"></typeparam>
+public interface IMvxResultAwaitingViewModel<TResult> : IMvxViewModel
```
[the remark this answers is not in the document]
entdark: That is needed to register VMs in `MvxResultViewModelManager._registrations`. Probably not the best way. I could make non-generic `IMvxResultAwaitingViewModel` that `IMvxResultAwaitingViewModel<TResult>` inherits from and use the same in `MvxResultViewModelManager._registrations` instead of `IMvxViewModel`. But then it's kind of incorrect since `IMvxResultAwaitingViewModel` supposes to have a result to await. End users may be confused which to use.
On the other hand I could use `object` in `MvxResultViewModelManager._registrations` instead of `IMvxViewModel`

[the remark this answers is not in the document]
Cheesebaron: Maybe instead of using `IMvxViewModel` as a common base, then use a new interface. Then you have something for `MvxResultViewModelManager` to look for, while not depending on `IMvxViewModel`.

MvvmCross/ViewModels/Result/MvxResultViewModelExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    public static void SaveRegisterToResult<TResult>(this IMvxResultAwaitingViewModel<TResult> viewModel, IMvxBundle savedStateBundle)
+    {
+        if (Mvx.IoCProvider?.TryResolve<IMvxResultViewModelManager>(out var manager) == true &&
+            manager.IsRegistered(viewModel) &&
+            savedStateBundle?.Data is { } data)
+        {
+            data[BundleRegisterKey] = true.ToString();
```
Cheesebaron, line 27: Could this be a const instead?

Cheesebaron:
I think I am OK with this implementation. One last thing, could you add some docs for this with examples of usage. That would help anyone with adopting this.

MvvmCross/ViewModels/Result/MvxResultViewModelExtensions.cs
entdark: if you mean `BundleRegisterKey` then it's already `const` defined above.

Cheesebaron: No the `true.ToString()` it will always result in `"true"` :)

follesoe:
It would be awesome if @entdark could push it to the last mile. This would help migrate code to MvvmCross 9 🙌 (as far as I can see it's only some documentation missing). Highly appreciate the effort done so far.

Cheesebaron:
I've gone through the code and upon playing with it, I might change it some to address some of the issues around having some of the logic in extension methods which isn't possible to override the behavior off. So I think I will merge this and finish it myself. This is not to discount the work @entdark has done. I really appreciate the work done here, I just need to ensure everyone is successful using this feature for the future.

entdark:
I still plan to document, and remember about the issue, I just did not have free time lately. I was waiting for the PR review for months so it would be nice if you waited for my documentations as well. Unless you want to do it your way. @Cheesebaron

Cheesebaron:
So, I've gone ahead and removed the hidden dependencies in the Extensions.

Another thing I experience, is in the Android sample, the result never gets set from the child model, do you have any idea why this is happening @entdark?

entdark:
@Cheesebaron
And I don't quite understand your changes and necessity of explicitly passing `IMvxResultViewModelManager`, but it's up to you ofc.
Maybe you opened the wrong VM?
And have you tested w/o your changes?

Cheesebaron:
The hidden dependencies in extension methods are not preferred because it hides dependencies. But also does not allow for overwriting behavior like you can most places in MvvmCross.

As for opening wrong VM, I don't think so. Opened the only instance of ChildVM there is in RootVM. Nothing is returned. Made sure OnDestroy is called on the Fragment

entdark:
Are you saying that the result is not set after the restoration or within just first app life w/o closing the app at all?

Cheesebaron:
> Are you saying that the result is not set after the restoration or within just first app life w/o closing the app at all?

In this case, I am simply opening the child VM using the first button in the list. Then pressing close button. Nothing is returned in `ResultSet`.


https://github.com/user-attachments/assets/b334b770-ff5c-4ef5-9706-77f23be755ef

entdark:
@Cheesebaron I will summarize what I wrote on Discord about this.
I found the issue of failing setting the result:
When `RootViewModel` navigates to `ChildViewModel` it registers to result when navigation is completed: https://github.com/entdark/MvvmCross/blob/new_result_navigation/MvvmCross/Navigation/MvxNavigationExtensions.cs#L88-L91
But navigation gets completed when `Initialize` task get completed and it's delayed by 8,5 seconds: https://github.com/entdark/MvvmCross/blob/new_result_navigation/Projects/Playground/Playground.Core/ViewModels/Navigation/ChildViewModel.cs#L60
So, if you wait 8,5 seconds and then close the `ChildViewModel` then result will be set correctly.
Currently it fails to set the result when `ChildViewModel` gets closed fast.

After some thinking I got that it's kind of weird that navigation depends on `Initialize` completion, especially if it fails at some point but visually navigation could be done, but `navigated` will be `false` at that point.
I think navigation should be just navigation and `Initialize` should not block it and go just afterward failing or not
So, to summarize it's better to change the navigation logic and no need to change anything in the `NavigateRegisteringToResult`.

Cheesebaron:
Added a separate sample for the navigation with result to not pollute existing ones.

entdark:
That is still an issue with heavy loading in `Initialize`

Cheesebaron:
> That is still an issue with heavy loading in `Initialize` in real projects.

I think if it is well enough documented, most people will have a pretty good time with this feature. So I think that is the last thing missing.

entdark:
But why would a failed `Initialize` fail navigation task? If visually we already saw the page appeared, it's just the logic behind that does not mark it as successful navigation. I think it can lead into more new unnecessary issues/discussions in the future.

Cheesebaron:
> But why would a failed `Initialize` fail navigation task? If visually we already saw the page appeared, it's just the logic behind that does not mark it as successful navigation. I think it can lead into more new unnecessary issues/discussions regarding result navigation in the future.

I am not 100% certain why it was done like this originally. My guess is to not trigger other lifecycle methods before Initialize task is fully finished.

Changing it to fire and forget would also be a breaking change, so I am inclined not to touch it for now. I'd rather do it in another PR.

entdark:
> I am inclined not to touch it for now

👌
Then I am fine with solving problems after their appearance if any.

If code part is settled then all I need is to document, right? I got some free time now.

Cheesebaron:
> > I am inclined not to touch it for now
> 
> 👌 Then I am fine with solving problems after their appearance if any.
> 
> If code part is settled then all I need is to document, right? I got some free time now.

Yeah, I don't think I am going to make more changes to this. Please go ahead and add some documentation for this feature.

entdark:
There you go.

PR #4849, opened 2024-04-08
Fixed creating and loading an Activity VM and its lifecycle twice after restoring on Android

entdark:
Bug fix.

An Android `Activity` restores its VM from `Intent` first, loads the VM, runs its lifecycle but eventually returns `null` to the restoration process so it does the direct loading from the VM type and repeats the process. In the end it does create and load the same VM twice and the first one stays in memory while the second one gets actually loaded (always `false` on line 65): https://github.com/MvvmCross/MvvmCross/blob/93c88422263385e69d60d8c9407f9ea6ed86c55a/MvvmCross/Platforms/Android/Views/MvxAndroidViewsContainer.cs#L64-L69

An Android `Activity` restores its VM from `Intent` and if succeeded then doesn't do direct loading so eventually there is only one VM lifecycle happens

No.

Try the same steps from #4848.

-

- [x] All projects build


Cheesebaron:
Thanks @entdark

PR #4893, opened 2025-04-02
Added MvxSceneDelegate for iOS SceneDelegate support

Cheesebaron:
This feature was added when over 10km above sea level and going ~1000 km/h

Feature

Currently we can only launch an App using AppDelegate. In order to support newer features, we need SceneDelegate support.

Added a new MvxSceneAppDelegate and MvxSceneDelegate classes to support scenarios when wanting to launch the App using SceneDelegate

Yes, it removes the AppDelegate property from the iOS presenter and changes a few methods to take IMvxLifecycle instead of AppDelegate instead.

Launch the playground app on iOS it is now using SceneDelegate

- [x] All projects build

### DIFF
--- a/MvvmCross/Platforms/Ios/Core/MvxSceneDelegate.cs
+++ b/MvvmCross/Platforms/Ios/Core/MvxSceneDelegate.cs
@@ -1,0 +1,84 @@
+#nullable enable
+using MvvmCross.Core;
+using MvvmCross.ViewModels;
+
+namespace MvvmCross.Platforms.Ios.Core;
+
+public abstract class MvxSceneDelegate : UIResponder, IUIWindowSceneDelegate, IMvxLifetime
+{
+    public event EventHandler<MvxLifetimeEventArgs>? LifetimeChanged;
+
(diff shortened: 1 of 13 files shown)


entdark:
Finally ;D

MvvmCross/Platforms/Ios/Core/MvxSceneDelegate.cs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    [Export("sceneDidBecomeActive:")]
+    public virtual void DidBecomeActive(UIScene scene)
+    {
+        FireLifetimeChanged(MvxLifetimeEvent.ActivatedFromMemory);
+    }
+
+    [Export("sceneWillResignActive:")]
+    public void WillResignActive(UIScene scene)
```
entdark, line 47: Make virtual as well?

Cheesebaron: Yeah, will have a look at that. Need to add some docs too on how to use, so will fix before release. Thanks for pointing this out :)

PR #4898, opened 2025-05-03
.NET 9 locking

MarkCiliaVincenti:
Improved performance for .NET 9.0+


Cheesebaron:
I am inclined to not accept this. Main issue being that it depends on a unknown package for older targets like .NET Standard. If it were a more popular and known author such as SimonCropp/Polyfill (not saying it polyfills all the Lock stuff), then I would have more confidence.

Also, this claims "improved performance", while the locking in MvvmCross does not happen in places where the performance gains of the newer Lock object even matters that much.

Issue #4957, opened 2025-09-30
[Bug]: Regression: typeof(TViewType) in MvxViewExtensions breaks ViewModel resolution


met94:
### Description

It looks like a recent change introduced a regression in `MvxViewExtensions`:  
[Reference to code](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Views/MvxViewExtensions.cs#L47C13-L47C76)

Previously, the code relied on `view.GetType()`, but it now uses `typeof(TViewType)`. These two are not equivalent in this context.  

When `FindAssociatedViewModelTypeOrNull` is called, `typeof(TViewType)` often resolves to `IMvxFragmentView` or `IMvxAndroidView` rather than the actual view type. As a result, the framework fails to correctly identify the associated ViewModel type.  

This change breaks existing functionality and prevents proper ViewModel resolution.

✅ Works fine on **9.4.0**  
❌ Broken on **develop**

### Steps to Reproduce
1. Show a fragment that has some bindings.  
2. Rotate the screen.  

### Expected Behavior
The ViewModel is resolved and all bindings are created successfully.  

### Actual Behavior
The framework fails to resolve a ViewModel for the fragment, the bindings do not initialize, and you see a warning in the output similar to:
```
[MvvmCross.Platforms.Android.Views.Fragments.MvxBindingFragmentAdapter] [Warning]
ViewModel type for Activity MyXamarinApp.Droid.Views.Main.MainContainerActivity not found
when trying to show fragment: MyXamarinApp.Droid.Views.Main.MainFragment
```

### Version with bug

develop

### Is this a regression from previous behavior?

Yes, this used to work in a previous version

### Last version that worked well

9.4.0

### Affected platforms

Android

snydesc:
I'm also running into this issue, but for me it was during the initial creation of the main Activity.  I use `RegisterCustomAppStart ` in my App.cs instead of doing `RegisterAppStart<RootViewModel>()`.

This will end up calling `LoadViewModel `found in `MvxActivityViewExtensions `which also calls `FindAssociatedViewModelTypeOrNull `
This produces the following error: `No ViewModel class specified for MainView in LoadViewModel`

Although this is not the best workaround, if you add the following to your MainActivity `OnCreate` before calling `base.OnCreate(savedInstanceState)` it seems to resolve the problem for this scenario.  

```
[...]

PR #4958, opened 2025-09-30
Update Fragment ViewModel caching to prevent stale reuse and memory retention

### LINKED ISSUE #4913: [Bug]: Stale ViewModel is sometimes returned from MvxMultipleViewModelCache in Android
### Description

On Android devices (We only develop on Android devices)
we have been noticing that when rotating the screen or bringing the app from the background, sometimes an old stale version of the view model is displayed. 

For example I may have a view/viewmodel displayed with a text field that says "ABC".  I then close this screen.  I then open the same screen again and enter "DEF" in the same text field.  I rotate the screen, The text field will now say "ABC".  

### Steps to Reproduce

This does not happen all the time, and seems to be tied to me rotating the screen during certain lifecycle events.

To verify that an old instance of the VM is being used. I have a GUID as a member variable that I print out when the ViewModel is displayed.  We will see something of the following: 

SCREEN LOAD ---->  TestVM 6cd569db-ac8e-4365-abe8-7ac32836032a
** Gallery App is displayed and user picks photo, after photo is selected, our app id displayed, and TestVM with ID 6cd569db-ac8e-4365-abe8-7ac32836032a is shown**
User closes TestVM with ID 6cd569db-ac8e-4365-abe8-7ac32836032a
User open TestVM again 
SCREEN LOAD ---->  TestVM de6d0a10-8728-4dc3-903a-4a0bcd3674f3
User closes TestVM with ID de6d0a10-8728-4dc3-903a-4a0bcd3674f3
User opens TestVM again
SCREEN LOAD ---->  TestVM 6cd569db-ac8e-4365-abe8-7ac32836032a

I created my own implementation of MvxMultipleViewModelCache and noticed that when "de6d0a10-8728-4dc3-903a-4a0bcd3674f3" is added to the cache, it fails because 6cd569db-ac8e-4365-abe8-7ac32836032a is already in there.  

What should happen is "de6d0a10-8728-4dc3-903a-4a0bcd3674f3" should never have been created.  It should have used "6cd569db-ac8e-4365-abe8-7ac32836032a" which was in the cache.  

### Link to public reproduction project repository

_No response_

### Version with bug

9.2.0

### Is this a regression from previous behavior?

Not sure, did not test other versions

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

Android 14

### Did you find any workaround?

I implemented my own IMvxMultipleViewModelCache, in void Cache if the view model is already in the cache, I remove it and add the viewmodel that's passed into the Cache function.

met94:
Bug fix
Fixes #4913

When a ViewModel of a specific type is already present in the cache, MvvmCross prevents caching a second instance of the same type until the first one is explicitly removed. This results in scenarios where the previously cached instance is returned, even though a new instance was intended. Consequently, the new ViewModel is not cached and may not be retrieved as expected. Additionally, the retained ViewModel remains in memory longer than necessary, which can interfere with garbage collection and lead to unintended memory retention.

When a ViewModel of a specific type is already cached and a new instance of the same type is requested to be cached, the new instance will now replace the existing one in the cache. This ensures that the most recent ViewModel is always retrievable and avoids unintended reuse of stale instances.

Additionally, the cached ViewModel is now explicitly removed during the OnResume lifecycle event. Since MvvmCross only attempts to retrieve cached ViewModels during OnCreate, retaining them beyond that point serves no purpose and may lead to unnecessary memory retention. This change helps reduce memory pressure and improves lifecycle alignment.

No

1. **Verify ViewModel replacement in cache:**
   - Create two instances of the same `ViewModel` type.
   - Ensure that when the second instance is cached, it replaces the first.
   - Confirm that retrieving the cached `ViewModel` returns the most recent instance.

2. **Validate cache clearing on OnResume:**
   - Navigate away from the view and trigger `OnResume`.
   - Ensure the `ViewModel` is removed from the cache.
   - Confirm that subsequent retrieval does not return the old instance. 

3. **Note on bug #4913 and #4957:**
   - This PR addresses bug https://github.com/MvvmCross/MvvmCross/issues/4913, which includes a minimal repro project and testing instructions.
   - However, due to a separate issue introduced in https://github.com/MvvmCross/MvvmCross/issues/4957, the repro steps from #4913 may not currently work as expected.
   - Reviewers should be aware that the fix is valid, but full verification may be blocked until #4957 is resolved.

- **Bug #4913**
  This PR addresses the caching issue described here. The issue includes a minimal repro project and testing instructions.

- **Bug #4957** 
  This newer issue interferes with the ability to fully verify the fix for #4913. Reviewers should be aware that the repro steps from #4913 may not work as expected until #4957 is resolved.

- [x] All projects build

### DIFF
--- a/MvvmCross/Platforms/Android/Views/Fragments/MvxBindingFragmentAdapter.cs
+++ b/MvvmCross/Platforms/Android/Views/Fragments/MvxBindingFragmentAdapter.cs
@@ -111,6 +111,17 @@ namespace MvvmCross.Platforms.Android.Views.Fragments
             object? sender, MvxValueEventArgs<MvxCreateViewParameters> e) =>
             FragmentView?.EnsureBindingContextIsSet(e.Value.Inflater);
 
+        protected override void HandleResumeCalled(object? sender, EventArgs e)
+        {
+            IMvxMultipleViewModelCache? cache = null;
+            if (Mvx.IoCProvider?.TryResolve(out cache) == true && FragmentView?.ViewModel != null)
+            {
+                // clear cache if still there
+                cache!.GetAndClear(FragmentView.ViewModel.GetType(), FragmentView.UniqueImmutableCacheTag);
+                return;
+            }
+        }
+
         protected override void HandleSaveInstanceStateCalled(object? sender, MvxValueEventArgs<Bundle> e)
         {
             // it is guaranteed that SaveInstanceState call will be executed before OnStop (thus before Fragment detach)
--- a/MvvmCross/Platforms/Android/Views/Fragments/MvxBindingFragmentAdapter.cs
+++ b/MvvmCross/Platforms/Android/Views/Fragments/MvxBindingFragmentAdapter.cs
@@ -113,12 +113,10 @@ namespace MvvmCross.Platforms.Android.Views.Fragments
 
         protected override void HandleResumeCalled(object? sender, EventArgs e)
         {
-            IMvxMultipleViewModelCache? cache = null;
-            if (Mvx.IoCProvider?.TryResolve(out cache) == true && FragmentView?.ViewModel != null)
+            if (Mvx.IoCProvider?.TryResolve(out IMvxMultipleViewModelCache? cache) == true && cache != null && FragmentView?.ViewModel != null)
             {
                 // clear cache if still there
-                cache!.GetAndClear(FragmentView.ViewModel.GetType(), FragmentView.UniqueImmutableCacheTag);
-                return;
+                cache.GetAndClear(FragmentView.ViewModel.GetType(), FragmentView.UniqueImmutableCacheTag);
             }
         }
 
--- a/MvvmCross/Platforms/Android/Views/MvxMultipleViewModelCache.cs
+++ b/MvvmCross/Platforms/Android/Views/MvxMultipleViewModelCache.cs
@@ -29,8 +29,7 @@ namespace MvvmCross.Platforms.Android.Views
             var type = toCache.GetType();
 
             var cachedViewModelType = new CachedViewModelType(type, viewModelTag);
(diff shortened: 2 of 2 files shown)


MvvmCross/Platforms/Android/Views/Fragments/MvxBindingFragmentAdapter.cs
```diff
@@ -111,6 +111,17 @@ protected override void HandleCreateViewCalled(
...
+        {
+            IMvxMultipleViewModelCache? cache = null;
+            if (Mvx.IoCProvider?.TryResolve(out cache) == true && FragmentView?.ViewModel != null)
+            {
+                // clear cache if still there
+                cache!.GetAndClear(FragmentView.ViewModel.GetType(), FragmentView.UniqueImmutableCacheTag);
+                return;
+            }
+        }
+
```
Cheesebaron, line 122: The `IMvxMultipleViewModelCache` field can be inlined

```suggestion
        protected override void HandleResumeCalled(object? sender, EventArgs e)
        {
            if (Mvx.IoCProvider?.TryResolve(out IMvxMultipleViewModelCache? cache) == true && cache != null && FragmentView?.ViewModel != null)
            {
                // clear cache if still there
                cache.GetAndClear(FragmentView.ViewModel.GetType(), FragmentView.UniqueImmutableCacheTag);
                return;
            }
        }

```

Cheesebaron (changes requested):

MvvmCross/Platforms/Android/Views/Fragments/MvxBindingFragmentAdapter.cs
met94: Thanks for the feedback. I have applied the change.

Cheesebaron (approved):

PR #5004, opened 2025-10-28
Allow to set window controller type as Type for the MvxWindowPresentationAttribute

### LINKED ISSUE #5003: [Mac] Allow to specify window controller type in addition to name for the MvxWindowPresentationAttribute
### Description

I propose to extend the [`MvxWindowPresentationAttribute`](https://github.com/MvvmCross/MvvmCross/blob/5370490eb5e6961bbc7ad70c8b005447c636c4cc/MvvmCross/Platforms/Mac/Presenters/Attributes/MvxWindowPresentationAttribute.cs) to allow to specify the associated window controller not as a class name (as a string), but also as a type (`System.Type`). This will provide additional convenience and compile-time checks (include AOT-related checks).

### Intended Use-Case

For now if we want to use `MvxWindowPresentationAttribute` for Mac with a specific `NSViewController`, the only way to specify the controller is to set the type *name* (not the `System.Type`, but the class name as a string). Moreover, we need to specify the full name, including assembly.

```csharp
    [MvxWindowPresentation(WindowControllerName = "Playground.Mac.TestWindowController, Playground.Mac")]
    public partial class ModalView : MvxViewController<ModalViewModel>
    {
    }
```

The main advantage of this approach is that it will work for any dependency tree. Even the controller class is not available in the compile-time, this will work. Also, it is used in case when we instantiate the controller by name from the storyboard. 

The disadvantages:
* it is too verbose (full name needed) and may be counterintuitive.
* it is not refactoring- and navigation-friendly. As the class name represented as a string, most of the tools will ignore it during the refactoring, code navigation and completion/suggestion.
* non AOT-friendly. AOT checks will not handle the name and developer will need to perform diagnostics and additional marking of the code manually.

# New approach
With the proposed approach we will be able to do the following instead
```csharp
    [MvxWindowPresentation(WindowControllerType = typeof(TestWindowController))]
    public partial class ModalView : MvxViewController<ModalViewModel>
    {
    }
```
This approach solves all the mentioned disadvantages and also includes the controller type into the AOT checks pipeline using the `DynamicallyAccessedMembersAttribute`.

**Note**: we still need to keep the original ability to set controller name as a string to handle the non-typical dependency graph as well as a "controller from the storyborad" case.

snechaev:
Feature and documentation update

Currently, only the name of the window controller class (as a string) can be specified in the `MvxWindowPresentationAttribute`.

The `MvxWindowPresentationAttribute` now allows specifying the type (`System.Type`) of the window controller in addition to its name.

No.

* Get a project with a `MvxWindowPresentationAttribute` presentation and without storyboard specified.
* Implement custom window controller inherited from the `MvxWindowController`
* specify the controller type in the `MvxWindowPresentationAttribute`
* make sure that navigation to this view works as expected.

Docs: https://www.mvvmcross.com/documentation/platform/macos/mac-view-presenter#presentation-attributes (also updated in this PR)
Fixes #5003 

- [x] All projects build

### DIFF
--- a/MvvmCross/Platforms/Mac/Presenters/MvxMacViewPresenter.cs
+++ b/MvvmCross/Platforms/Mac/Presenters/MvxMacViewPresenter.cs
@@ -220,12 +220,16 @@ namespace MvvmCross.Platforms.Mac.Presenters
             }
             else
             {
+                var controllerType = attribute.WindowControllerType ?? Type.GetType(attribute.WindowControllerName);
+                if (controllerType is null)
+                {
+                    throw new MvxException(
+                        $"Could not determine window controller type for the {attribute.ViewModelType?.Name ?? "<unknown vm>"} view model. " +
+                        $"Please specify either the {nameof(MvxWindowPresentationAttribute.WindowControllerType)} or " +
+                        $"{nameof(MvxWindowPresentationAttribute.WindowControllerName)} property of the {nameof(MvxWindowPresentationAttribute)} " +
+                        $"for the corresponding view model.");
+                }
                 // Instantiate using Reflection - failure is possible if blank constructor is missing
-                var controllerType = attribute.WindowControllerType ?? Type.GetType(attribute.WindowControllerName)
-                    ?? throw new MvxException($"Could not determine window controller type for the {attribute.ViewModelType?.Name ?? "<unknown vm>"} view model. " +
-                                              $"Please specify either in the {nameof(MvxWindowPresentationAttribute.WindowControllerType)} or " +
-                                              $"{nameof(MvxWindowPresentationAttribute.WindowControllerName)} property of the {nameof(MvxWindowPresentationAttribute)} " +
-                                              $"for the corresponding view model.");
                 windowController = (MvxWindowController)Activator.CreateInstance(controllerType);
             }
             windowController.ShouldCascadeWindows = attribute.ShouldCascadeWindows;
--- a/MvvmCross/Platforms/Mac/Presenters/MvxMacViewPresenter.cs
+++ b/MvvmCross/Platforms/Mac/Presenters/MvxMacViewPresenter.cs
@@ -221,7 +221,12 @@ namespace MvvmCross.Platforms.Mac.Presenters
             else
             {
                 // Instantiate using Reflection - failure is possible if blank constructor is missing
-                windowController = (MvxWindowController)Activator.CreateInstance(Type.GetType(attribute.WindowControllerName));
+                var controllerType = attribute.WindowControllerType ?? Type.GetType(attribute.WindowControllerName)
+                    ?? throw new MvxException($"Could not determine window controller type for the {attribute.ViewModelType?.Name ?? "<unknown vm>"} view model. " +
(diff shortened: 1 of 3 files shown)


docs/_documentation/platform/macos/mac-view-presenter.md
```diff
@@ -49,6 +49,10 @@ Properties introduced inside the attribute will override the storyboard properti
 ```
 will set the CustomWindow Width to 800.
 
+*Note*: when storyboard name is set, the WindowControllerType is ignored.
+
+To use a custom window controller without storyboard, you can specify either the _WindowControllerType_ or the _WindowControllerName_ (specify the full type name, including the assembly name) property while *not* specifying the _StoryboardName_. The _WindowControllerType_ property takes precedence over the _WindowControllerName_ property.
+
```
Cheesebaron:
```suggestion
> Note: when storyboard name is set, the WindowControllerType is ignored.

To use a custom window controller without storyboard, you can specify either the _WindowControllerType_ or the _WindowControllerName_ (specify the full type name, including the assembly name) property while *not* specifying the _StoryboardName_. The _WindowControllerType_ property takes precedence over the _WindowControllerName_ property.

```

MvvmCross/Platforms/Mac/Presenters/MvxMacViewPresenter.cs
```diff
@@ -221,7 +221,12 @@ protected virtual MvxWindowController CreateWindowController(MvxWindowPresentati
             else
             {
                 // Instantiate using Reflection - failure is possible if blank constructor is missing
-                windowController = (MvxWindowController)Activator.CreateInstance(Type.GetType(attribute.WindowControllerName));
+                var controllerType = attribute.WindowControllerType ?? Type.GetType(attribute.WindowControllerName)
+                    ?? throw new MvxException($"Could not determine window controller type for the {attribute.ViewModelType?.Name ?? "<unknown vm>"} view model. " +
+                                              $"Please specify either in the {nameof(MvxWindowPresentationAttribute.WindowControllerType)} or " +
+                                              $"{nameof(MvxWindowPresentationAttribute.WindowControllerName)} property of the {nameof(MvxWindowPresentationAttribute)} " +
+                                              $"for the corresponding view model.");
```
Cheesebaron: Please don't nest multiple levels of ternaries

Cheesebaron (changes requested):

MvvmCross/Platforms/Mac/Presenters/MvxMacViewPresenter.cs
snechaev: Replaced with a ternary operator + separated null check with throw. Please let me know if you prefer any other style here.

Cheesebaron (approved):

snechaev:
It looks like the auto-merge has got stuck, but I can't see why. Am I missing something and some additional changes or actions needed?

Cheesebaron:
> It looks like the auto-merge has got stuck, but I can't see why. Am I missing something and some additional changes or actions needed?

Yeah don't worry. Code quality stuff doesn't run for code coming from forks due to security. I've merged the PR, thank you!

PR #5028, opened 2025-11-25
Android multiple back stacks

entdark:
1. Added 2 new parameters to `MvxFragmentPresentationAttribute` that are required and add a possibility to use [multiple back stacks](https://developer.android.com/guide/fragments/fragmentmanager#multiple-back-stacks) (in bottom tabs-like navigation):
- `AllowReordering`: "optimization" that is required by multiple back stacks
- `SetAsPrimaryFragment`: helper that handles back button in nested Fragments
2. Added a sample with `BottomNavigationView` (tabs) that uses multiple back stacks in each tab.

There is no special parameters that help to use multiple back stack with presenter.

A possibility to pop nested `Fragment`s. A possibility to use multiple back stacks without modifying the Android presenter.

No.

1. Launch the app from Playground.Droid project.
2. Click on "Multiple back stacks" button.
3. Click `Go deeper` multiple times.
4. Switch between tabs on bottom.
5. Confirm the back stacks get restored when switching between tabs.
6. Click system Back button in Tab 1 to see how it closes the inner `Fragment`s in Tab 1.

### *️⃣ Extra
1. Unfortunately while we can add some logic of multiple back stacks requirements to the presenter we have to manually call `SaveBackStack` and `RestoreBackStack` (see the tabs selection in the sample). That is so because the `FragmentManager` API does not let us to find out if any back stack is already saved, and to "pre-save" any back stack ahead. So we manually do those. So we have to decide on our own to restore the back stack or to navigate to it first if never navigated before.
2. To switch between back stacks we pass the root `Fragment` name of our back stack to `SaveBackStack` that can be either `Tag` or `typeof(Fragment).FragmentJavaName()`. But eventually for `Tag` it works incorrectly because of it cannot find proper host `Fragment` because uses only `FragmentJavaName()`: https://github.com/MvvmCross/MvvmCross/blob/v10.0.0/MvvmCross/Platforms/Android/Presenters/MvxAndroidViewPresenter.cs#L1121
3. The back stack restoration happens with `Fragment`s restoring their VMs (finding cached ones) and that requires to put explicit `ViewModelType` to every `MvxFragmentPresentationAttribute` of nested `Fragment`s.
4. There is a bug/feature of `SetAsPrimaryFragment` that also pops to the very first `Fragment` in the tab since it has `AddToBackStack` that is required by multiple back stacks. Although `AddToBackStack` is logically not needed for the root tabs `Fragment`s, but is required for the save/restore API. The workaround would be either to manually intercept back button with `OnBackPressedCallback` that checks the current `Fragment` `ChildFragmentManager` back stack count and does not let the root one to be popped, or to play around with more nesting, or to not use `SetAsPrimaryFragment`.

[Multiple back stacks](https://developer.android.com/guide/fragments/fragmentmanager#multiple-back-stacks)
[Primary navigation fragment](https://developer.android.com/reference/androidx/fragment/app/FragmentTransaction#setPrimaryNavigationFragment(androidx.fragment.app.Fragment))
[Allow reordering](https://developer.android.com/reference/androidx/fragment/app/FragmentTransaction#setReorderingAllowed(boolean))
[Save back stack](https://developer.android.com/reference/androidx/fragment/app/FragmentManager#saveBackStack(java.lang.String))
[Restore back stack](https://developer.android.com/reference/androidx/fragment/app/FragmentManager#restoreBackStack(java.lang.String))

- [x] All projects build

### DIFF
--- a/MvvmCross/Platforms/Android/Presenters/Attributes/MvxFragmentPresentationAttribute.cs
+++ b/MvvmCross/Platforms/Android/Presenters/Attributes/MvxFragmentPresentationAttribute.cs
@@ -31,7 +31,9 @@ public class MvxFragmentPresentationAttribute : MvxBasePresentationAttribute
         string? tag = null,
         string popBackStackImmediateName = "",
         MvxPopBackStack popBackStackImmediateFlag = MvxPopBackStack.Inclusive,
-        bool addFragment = false
+        bool addFragment = false,
+        bool allowReordering = false,
+        bool setAdPrimaryFragment = false
     )
     {
         ActivityHostViewModelType = activityHostViewModelType;
@@ -48,6 +50,8 @@ public class MvxFragmentPresentationAttribute : MvxBasePresentationAttribute
         PopBackStackImmediateName = popBackStackImmediateName;
         PopBackStackImmediateFlag = popBackStackImmediateFlag;
         AddFragment = addFragment;
+        AllowReordering = allowReordering;
+        SetAsPrimaryFragment = setAdPrimaryFragment;
     }
 
     public MvxFragmentPresentationAttribute(
@@ -65,7 +69,9 @@ public class MvxFragmentPresentationAttribute : MvxBasePresentationAttribute
         string? tag = null,
         string popBackStackImmediateName = "",
         MvxPopBackStack popBackStackImmediateFlag = MvxPopBackStack.Inclusive,
-        bool addFragment = false
+        bool addFragment = false,
+        bool allowReordering = false,
+        bool setAdPrimaryFragment = false
     )
     {
         if (Mvx.IoCProvider?.TryResolve(out IMvxAndroidGlobals globals) == true &&
@@ -106,6 +112,8 @@ public class MvxFragmentPresentationAttribute : MvxBasePresentationAttribute
         PopBackStackImmediateName = popBackStackImmediateName;
         PopBackStackImmediateFlag = popBackStackImmediateFlag;
         AddFragment = addFragment;
+        AllowReordering = allowReordering;
+        SetAsPrimaryFragment = setAdPrimaryFragment;
     }
 
     /// <summary>
@@ -186,4 +194,14 @@ public class MvxFragmentPresentationAttribute : MvxBasePresentationAttribute
     /// Setting this to true, will use Add instead of Replace on the Fragment transaction
     /// </summary>
     public bool AddFragment { get; set; }
+
+    /// <summary>
+    /// Setting this to true, will use SetReorderingAllowed on the Fragment transaction
+    /// </summary>
+    public bool AllowReordering { get; set; }
+
+    /// <summary>
+    /// Setting this to true, will use SetPrimaryNavigationFragment on the Fragment transaction
+    /// </summary>
+    public bool SetAsPrimaryFragment { get; set; }
 }
--- a/MvvmCross/Platforms/Android/Presenters/Attributes/MvxFragmentPresentationAttribute.cs
+++ b/MvvmCross/Platforms/Android/Presenters/Attributes/MvxFragmentPresentationAttribute.cs
@@ -33,7 +33,7 @@ public class MvxFragmentPresentationAttribute : MvxBasePresentationAttribute
         MvxPopBackStack popBackStackImmediateFlag = MvxPopBackStack.Inclusive,
         bool addFragment = false,
         bool allowReordering = false,
-        bool setAdPrimaryFragment = false
+        bool setAsPrimaryFragment = false
     )
     {
         ActivityHostViewModelType = activityHostViewModelType;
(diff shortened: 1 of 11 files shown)


MvvmCross/Platforms/Android/Presenters/Attributes/MvxFragmentPresentationAttribute.cs
```diff
@@ -31,7 +31,9 @@ public MvxFragmentPresentationAttribute(
         string? tag = null,
         string popBackStackImmediateName = "",
         MvxPopBackStack popBackStackImmediateFlag = MvxPopBackStack.Inclusive,
-        bool addFragment = false
+        bool addFragment = false,
+        bool allowReordering = false,
+        bool setAdPrimaryFragment = false
```
Cheesebaron: Spelling mistake?

```suggestion
        bool setAsPrimaryFragment = false
```

MvvmCross/Platforms/Android/Presenters/MvxAndroidViewPresenter.cs
```diff
@@ -590,6 +590,9 @@ protected virtual void PerformShowFragmentTransaction(
 
             OnBeforeFragmentChanging(ft, fragment, attribute, request);
 
+            if (attribute.AllowReordering)
+	            ft.SetReorderingAllowed(true);
```
Cheesebaron: What is the default on a Fragment Transaction? Is it `false`? or perhaps some other state?

If it defaults to `false` perhaps this could just be written as:

```csharp
ft.SetReorderingAllowed(attribute.AllowReordering);
```

Projects/Playground/Playground.Core/ViewModels/Navigation/MultiBackStackViewModel.cs
```diff
@@ -0,0 +1,90 @@
...
+public class MultiBackStackViewModel(ILoggerFactory logFactory, IMvxNavigationService navigationService) : MvxNavigationViewModel(logFactory, navigationService)
+{
+	private bool _initialNavigationDone = false;
+
+	private async Task ShowInitialViewModelsExecute()
+	{
+		if (_initialNavigationDone)
+			return;
+		_initialNavigationDone = true;
+		var viewModelTypes = new[] { typeof(MultiBackStackTab1ViewModel), typeof(MultiBackStackTab2ViewModel) };
```
Cheesebaron: Unused?

Projects/Playground/Playground.Droid/Fragments/MultiBackStackView.cs
```diff
@@ -0,0 +1,142 @@
...
+    FragmentHostViewType = typeof(MultiBackStackView),
+    AllowReordering = true,
+    ViewModelType = typeof(MultiBackStackTab1ViewModel))]
+public class MultiBackStackTab1View : MvxFragment<MultiBackStackTab1ViewModel>
+{
+    public override View OnCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState)
+    {
+	    base.OnCreateView(inflater, container, savedInstanceState);
+
+	    var view = this.BindingInflate(Resource.Layout.MultiBackStackTab1View, null);
```
Cheesebaron: You should prefer using


```suggestion
	    var view = this.BindingInflate(Resource.Layout.MultiBackStackTab1View, container, false);
```

Projects/Playground/Playground.Droid/Fragments/MultiBackStackView.cs
```diff
@@ -0,0 +1,142 @@
...
+    {
+	    base.OnCreateView(inflater, container, savedInstanceState);
+
+	    var view = this.BindingInflate(Resource.Layout.MultiBackStackTab2View, null);
+
+	    return view;
+    }
+}
+
+[MvxFragmentPresentation(typeof(RootViewModel), Resource.Id.content_frame, true,
```
Cheesebaron: No need to provide this if you are using `IMvxOverridePresentationAttribute`

Cheesebaron (commented):
Just a few comments to the code here which I'd like you to address :)

MvvmCross/Platforms/Android/Presenters/Attributes/MvxFragmentPresentationAttribute.cs
entdark: Yes 😅

MvvmCross/Platforms/Android/Presenters/MvxAndroidViewPresenter.cs
entdark: Default is `false`. While the suggestion is fine, I think calling it with `false` and not calling at all potentially can have different behaviour.

Projects/Playground/Playground.Core/ViewModels/Navigation/MultiBackStackViewModel.cs
entdark: Unused!

Cheesebaron (approved):

PR #5088, opened 2026-03-25
Migrate to Microsoft.Extensions.DependencyInjection

Cheesebaron:
Feature/Refactor

Using custom IMvxIoCProvider

Switching to Microsoft.Extensions.DependencyInjection for IoC and new MvxHost for hosting model in favor of MvxSetup and related classes

Yes huge!

Run Playground.Droid or Playground.iOS to see it in action

- [x] All projects build

This was co-authored with help from @copilot using the Claude Sonnet 4.6 model


entdark:
Shouldn't those `AddMvvmCross*` method be like `MvxAdd*`, since it's never ever used the word `MvvmCross` explicitly in namings before, `Mvx` only afaik.

Does it solve any real problem? Does it speed up DI resolving?
Could `Mvx.IoCProvider` be just deprecated and fallback to the new behaviour to make the breaking changes less breaking?

Cheesebaron:
> Shouldn't those `AddMvvmCross*` method be like `MvxAdd*`, since it's never ever used the word `MvvmCross` explicitly in namings before, `Mvx` only afaik.
> 
> Does it solve any real problem? Does it speed up DI resolving? Could `Mvx.IoCProvider` be just deprecated and fallback to the new behaviour to make the breaking changes less breaking?

Thank you for the feedback. Exactly why I put it in draft :)

> Does it solve any real problem?

YES! Ever tried to add HttpClientFactory, resilience, StrawberryShake GraphQL client or any other library that expects MEDI to be present?
Then you have to jump hoops to add these types manually to the MvvmCross IoC container.

Most libraries simply assume that MEDI is available and is the standard most places.

Take for instance logging. When you have MEDI available, you simply write:

```
serviceProvider.AddLogging();
```

With MvvmCross we have to do:

```
iocProvider.RegisterSingleton(logProvider);

iocProvider.RegisterSingleton(loggerFactory);
iocProvider.RegisterType(typeof(ILogger<>), typeof(Logger<>));
```

Not only that, then the consumer of MvvmCross cannot use the callback in `AddLogging` to use the nicely provided extension methods in logging libraries to just call `AddSerilog` or whatever to add and configure their logging providers.

Also if you want to use Microsoft.Extensions.Configuration, good luck :)

Then there is the whole story of annotations to support trimming and AOT :D

So most definitely it solves a very big pain point.

> Does it speed up DI resolving

I don't know, I haven't benchmarked it.

> Could Mvx.IoCProvider be just deprecated and fallback to the new behaviour to make the breaking changes less breaking?

Not sure how this would look. Some kind of shim to register into MEDI when calling Mvx.IoCProvider? That could be possible. On the other hand, how many places do you use `Mvx.IoCProvider` outside of registering types at startup? I hope it is not many.

snechaev:
> Not sure how this would look. Some kind of shim to register into MEDI when calling Mvx.IoCProvider? That could be possible. On the other hand, how many places do you use Mvx.IoCProvider outside of registering types at startup? I hope it is not many.

I'm not sure if this will be possible at all. When I tried using the Autofac DI instead of the built-in `MvxIoCProvider`, I found that the internal Mvx logic did not follow the configure-build-resolve pattern, meaning that `Register` can be called *after* `Resolve`. Modern DI frameworks do not allow such out-of-order operations and strictly requires to register anything before first `Resolve` call. So, even with the shim (call translation between) old-style IoC provider and MEDI, the full backward compatibility will not be possible.

Cheesebaron:
> > Not sure how this would look. Some kind of shim to register into MEDI when calling Mvx.IoCProvider? That could be possible. On the other hand, how many places do you use Mvx.IoCProvider outside of registering types at startup? I hope it is not many.
> 
> I'm not sure if this will be possible at all. When I tried using the Autofac DI instead of the built-in `MvxIoCProvider`, I found that the internal Mvx logic did not follow the configure-build-resolve pattern, meaning that `Register` can be called _after_ `Resolve`. Modern DI frameworks do not allow such out-of-order operations and strictly requires to register anything before first `Resolve` call. So, even with the shim (call translation between) old-style IoC provider and MEDI, the full backward compatibility will not be possible.

Yeah, this won't be possible anymore with the approach taken in this PR. Everything has to register at startup. So this will be a breaking change.

entdark:
Why is `Setup.cs` not needed anymore (I am seeing you removed those in Playground on each platform)? What if I want to adjust things?

Cheesebaron:
> Why is `Setup.cs` not needed anymore (I am seeing you removed those in Playground on each platform)? What if I want to adjust things?

You will have the opportunity to overwrite anything in this block:

```csharp
    MvxIosHostBuilder.CreateBuilder(Window)
        .StartWith<RootViewModel>()
        .ConfigureServices(services =>
        {
            services.AddMvxBindings(config => config
                .FillTargetFactories(registry =>
                    registry.RegisterCustomBindingFactory<BinaryEdit>(
                        "MyCount",
                        view => new BinaryEditTargetBinding(view))));
            services.AddLogging(l => l.AddSerilog());
            services.AddMvxMessenger();
            services.AddMvxVisibility();
            services.AddMvxColor();
            services.AddMvxJson();
            services.AddMvxViewModels(typeof(RootViewModel).Assembly);
            services.AddMvxIosViews(typeof(SceneDelegate).Assembly);
        })
        .Build()
```

So the `AddMvxBindings` has callbacks that allows you to register factories for various things. You can remove dependencies from the container and add your own before calling `Build()`.
So anything you do in `ConfigureServices` will happen _after_ MvvmCross does its own registration, so you can do the overwrites there.

entdark:
Roger that.
Where do I register (return) my own presenter (`IMvxAndroidViewPresenter`)? Where do I return my own binding builder (`MvxBindingBuilder`)?
Will it have all the corresponding methods that Setup had overrides for?

Cheesebaron:
> Roger that. Where do I register (return) my own presenter (`IMvxAndroidViewPresenter`)? Where do I return my own binding builder (`MvxBindingBuilder`)? Will it have all the corresponding methods that Setup had overrides for?

Yep good point. I will make it easier to override these or specify them up front as these are often customized.

I think I will also have a look at the default presenters and make it easier to register new attributes without having to subclass it. That will be another PR. Have a look at #5100

Cheesebaron:
So now you should be able to override presenter and navigation service like:

```csharp
 MvxAndroidHostBuilder.CreateBuilder(application)
       .UsePresenter<MyAndroidPresenter>()
       .UseNavigationService<MyNavService>()
       .StartWith<MainViewModel>()
       .Build();
```

Question is if we need the ease of use to override Views Container and Dispatcher. I've never had the need to override these.

snechaev:
> Question is if we need the ease of use to override Views Container and Dispatcher. I've never had the need to override these.

I wrote about a case where I needed to use a custom View Container in addition to a custom Presenter - see https://github.com/MvvmCross/MvvmCross/issues/5100#issuecomment-4299591324. I can't say for sure whether this is a common case or not. So, we definitely need a way to override. Does it worth making a significant effort to make this override easy to do - I do not know.

entdark:
> So now you should be able to override presenter and navigation service like:
> 
> ```cs
>  MvxAndroidHostBuilder.CreateBuilder(application)
>        .UsePresenter<MyAndroidPresenter>()
>        .UseNavigationService<MyNavService>()
>        .StartWith<MainViewModel>()
>        .Build();
> ```
> 
> Question is if we need the ease of use to override Views Container and Dispatcher. I've never had the need to override these.

The beauty of MvvmCross is the ability to override almost everything and alter some behaviors to your own needs. If it's taken away, then it's a big step back. So, if one does not need something now, they might need it in the future.

I, for instance, need a custom `MvxBindingBuilder`: https://github.com/entdark/MvvmCross-TabsNavigation/blob/master/TabsNavigation.Android/Setup.cs#L57 to support my own window insets handling in XML layouts.

So, everything from the current `Mvx[Platform]Setup` should be adjustable.

snechaev:
> So, everything from the current Mvx[Platform]Setup should be adjustable.

@entdark, per my understanding, all of this is still fully adjustable. The only difference is that before, we had dedicated overloads to customize the setup, whereas now we have access to the `ContainerBuilder` and can register or replace anything we want. including the `MvxBindingBuilder`, for instance. For some elements we have shortcuts like `UsePresenter<T>()`, but they generally do not do anything special; they just remove the old registration and add the new one. The same thing can be done manually for (almost) any Mvx service or component. 

Could you please clarify your point a little bit more? Do you see any problems with the existing way of overriding and replacing components? Or are you primarily concerned about the absence of shortcut methods for literally each component or the lack of a centralized way (like a setup class with overrides) to handle these replacements?

entdark:
My point was the ability to use custom stuff as before. If it's still achievable then I am ok with it.

Cheesebaron:
As @snechaev pointed out, most methods in the old MvxSetup that you could override were just a way to provide an instance to the IoC container.

The difference using the new IoC is that you can call remove on the container then add your own. For your convenience I've added a few methods that makes it easy on the builder, to add presenter, view container and navigation service. If you want to override something else, you always have the possibility to register your own instance in the IoC container before calling build.

PR #5119, opened 2026-05-18
Rework Playground.Mac to work with new Microsoft DI

snechaev:
Bug fix

Playground.Mac won't start

Playground.Mac can start

no

Related/Caused by: https://github.com/MvvmCross/MvvmCross/pull/5088

- [x] All projects build

### DIFF
--- a/MvvmCross/Platforms/Mac/Hosting/MvxMacHostBuilder.cs
+++ b/MvvmCross/Platforms/Mac/Hosting/MvxMacHostBuilder.cs
@@ -3,12 +3,15 @@
 // See the LICENSE file in the project root for more information.
 #nullable enable
 
-using AppKit;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using MvvmCross.Base;
 using MvvmCross.Hosting;
 using MvvmCross.Platforms.Mac.Presenters;
+using MvvmCross.Platforms.Mac.Views;
 using MvvmCross.Presenters;
+using MvvmCross.ViewModels;
+using MvvmCross.Views;
 
 namespace MvvmCross.Platforms.Mac.Hosting;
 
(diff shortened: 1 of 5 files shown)


MvvmCross/Platforms/Mac/Hosting/MvxMacHostBuilder.cs
```diff
@@ -18,17 +21,33 @@ namespace MvvmCross.Platforms.Mac.Hosting;
...
-        ArgumentNullException.ThrowIfNull(window);
-        return new MvxMacHostBuilder(window);
+        return new MvxMacHostBuilder();
     }
 
-    private MvxMacHostBuilder(NSWindow window)
+    private MvxMacHostBuilder()
     {
-        Services.TryAddSingleton(window);
+        Services.TryAddSingleton(NSApplication.SharedApplication.Delegate);
```
snechaev, line 33: In the iOS version we receive the app main window as the parameter and then pass it to the view presenter. For macOS we need to pass application delegate instead of the main window. But anyway, I decided to remove explicit parameter of the `INSApplicationDelegate` here. We still have an implicit references to the `NSApplication.Shared.MainWindow` over the codebase, so I think there is nothing bad if we will also has an implicit reference to the app delegate. At the same time this make initialization code less wordy.

MvvmCross/Platforms/Mac/Views/MvxMacViewServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+using MvvmCross.Base;
+using MvvmCross.ViewModels;
+using MvvmCross.Views;
+
+namespace MvvmCross.Platforms.Mac.Views;
+
+/// <summary>
+/// Extension methods for registering macOS view-to-viewmodel mappings on an <see cref="IServiceCollection"/>.
+/// </summary>
+public static class MvxMacViewServiceCollectionExtensions
```
snechaev, line 14: All this class andn ist dependencies are almost a copy&paste of the corresponding classes from the iOS. And as I can see we have the same copy&paste for Android. Is it ok or we can somehow generalize/deduplicate this code?

Projects/Playground/Playground.Mac/AppDelegate.cs
```diff
@@ -21,19 +21,29 @@ public AppDelegate()
...
-            MvxMacHostBuilder.CreateBuilder(mainWindow)
+            MvxMacHostBuilder.CreateBuilder()
                 .StartWith<RootViewModel>()
                 .ConfigureServices(services =>
                 {
+                    services.AddLogging(l => l.AddSerilog());
                     services.AddMvxBindings();
                     services.AddMvxJson();
+                    services.AddMvxViewModels(typeof(RootViewModel).Assembly);
+                    services.AddMvxMacViews(typeof(AppDelegate).Assembly);
```
snechaev, line 39: Is it expected that we should add Views and ViewModels manually? May be it should be done in the `HostBuilder` by default with a way to override/reonfigure it if needed.

Cheesebaron (approved):

Cheesebaron:
Thank you!

PR #5122, opened 2026-05-20
Enable nullable attributes globally

Cheesebaron:
Bugfix/refactor

Nullable attributes only applied and enabled on some projects

Globally enabled nullable attributes and had Copilot fix the warnings. Used a mix of Claude Sonnet 4.6 and GPT-5.5 to work on this

Potentially

- [x] All projects build


Cheesebaron (commented):
## Review of #5122 – Enable nullable attributes globally

The overall direction is correct: centralising `<Nullable>enable</Nullable>` in `Directory.Build.props` and removing per-project/per-file pragmas is the right approach. However, the automated fix used the null-forgiving operator (`!`) in several places where the proper fix is to make the return/property type nullable (`?`). Those are not equivalent: `null!` silences the compiler while the value is still `null` at runtime, defeating the purpose of nullable reference types.

---

### 🔴 Critical — incorrect nullability contracts

**1. `IMvxViewTypeResolver.Resolve` must return `Type?`**

Multiple implementations can return `null`, so the interface contract is incorrect:

```csharp
// IMvxViewTypeResolver.cs
Type Resolve(string tagName);  // should be Type?
```

Implementations that use `return null!;` to work around the non-nullable return type:
- `MvxCompositeViewTypeResolver.Resolve` — iterates resolvers, returns `null!` as fallback
- `MvxNamespaceListViewTypeResolver.Resolve` — was correctly `Type?` before this PR; the PR *downgraded* it to `Type` with `return null!;`
- `MvxJustNameViewTypeResolver.Resolve` — returns `null!` on two paths
- `MvxLongLowerCaseViewTypeResolver.ResolveLowerCaseTypeName` — `Registry.TryResolve` may fail, making `type` null; `return type!` just masks this

`MvxCachedViewTypeResolver` is also affected: it stores the resolver's result (which may be null) into a `Dictionary<string, Type>` and returns it from a `Type`-returning method—both without the `!` suppressors but still functionally broken.

**Fix:** Change the interface to `Type? Resolve(string tagName)`, remove all `null!` from these implementations, update callers to handle nullable.

---

**2. `MvxTvosViewsContainer.CurrentRequest` and `IMvxCurrentRequest` must use `?`**

```csharp
// MvxTvosViewsContainer.cs
public MvxViewModelRequest CurrentRequest { get; private set; } = null!;
// ...
finally { CurrentRequest = null!; }  // intentionally set to null after each use
```

The `finally` block has always reset this to `null` — that's by design. Suppressing the warning here is wrong; the property should be `MvxViewModelRequest?` and `IMvxCurrentRequest` (iOS, tvOS, Mac versions) updated accordingly.

---

**3. `MvxExpandableListAdapter` — virtual methods returning `null!`**

```csharp
public virtual Object GetGroup(int groupPosition) => null!;
public virtual Object GetChild(int groupPosition, int childPosition) => null!;
public virtual Tuple<int, int> GetPositions(object childItem) { ... return null!; }
```

These methods genuinely return `null`; the return types must be nullable: `Object?`, `Object?`, and `Tuple<int,int>?`.

---

**4. `MvxCachingFragmentStatePagerAdapter.SaveState` — unsafe `!`**

```csharp
var bundle = base.SaveState() as Bundle;  // bundle is Bundle? here
SaveFragmentsInfoState(bundle!);          // ⚠️ bundle may actually be null
```

`base.SaveState()` now correctly returns `IParcelable?`; the `as Bundle` cast can yield `null`. A null guard is needed:

```csharp
if (bundle != null)
    SaveFragmentsInfoState(bundle);
```

---

### 🟡 Significant — nullable semantics degraded on public API

**5. `BindingContext` changed from nullable to non-nullable in many views**

Several Activity, Fragment, and View classes changed from:
```csharp
public IMvxBindingContext? BindingContext { get; set; }
```
to:
```csharp
public IMvxBindingContext BindingContext { get; set; } = null!;
```

If the previous `?` annotation was intentional (the binding context is legitimately null before initialization), consumers that null-check `BindingContext` will now get a false-positive warning from the analyzer, and consumers that skip the null-check will get no warning where they should.

At minimum, keep `= null!` but verify that any interface that declares this property (`IMvxBindingContext BindingContext`) is also updated to `IMvxBindingContext?` where appropriate, so the contract is consistent.

---

**6. `MvxBaseCollectionViewSource.UnknownCellIdentifier` (tvOS)**

```csharp
public static readonly NSString UnknownCellIdentifier = null!;
```

This field is a sentinel whose value is intentionally `null`. Its type must be `NSString?`.

---

### 🔵 Minor — redundant `null!`

In `MvxCachingFragmentPagerAdapter`, the lists were widened to accept null:

```csharp
private List<string?> _savedFragmentTags = new List<string?>();
private readonly List<Fragment.SavedState?> _savedState = new List<Fragment.SavedState?>();
```

The subsequent `Add(null!)` calls are then redundant—just use `Add(null)` since the type already permits it.

---

### ✅ What's done well

- Moving `<Nullable>enable</Nullable>` to `Directory.Build.props` — correct consolidation
- Removing per-file `#nullable enable`/`restore` pragmas — correct cleanup
- Properly marking fields/parameters nullable: `_itemsSource?`, `_subscription?`, `IEnumerable?`, event handlers as `EventHandler?`, etc.
- `Activity?.GetType() ?? typeof(Activity)` in `MvxCachingFragmentStatePagerAdapter` — proper null-safe expression
- `SaveState()` return type changed to `IParcelable?` — correct
- `RestoreState(IParcelable? state, ClassLoader? loader)` — correct parameter annotations

